PR #30, opened 2020-01-07
Adding Release Pipeline

sfoslund:
Had to open a new PR due to git issues after making the repo public. See the old PR [here](https://github.com/dotnet/vscode-dotnetcore/pull/26). 

Adding release pipeline that includes:
- Trigger via adding tag of the form "v*"
- Publishing to VSCode marketplace
- Creating a GitHub release

Updating versioning:
- Release version format: { Major }.{ Minor }.{ Patch }
- Daily build version format: { Major }.{ Minor }.{ Patch }-alpha-{ BuildId }


sfoslund:
@wli3 I've removed the GitHub release step since it looks like getting a service connection into the dnceng project will be complicated. We can always add it back later but since it's not a vital step (it essentially only saves us from uploading the .vsix from azure pipelines to GitHub) I think it's fine to start with what we have. Let me know if you disagree. 

I will need to update the publisher once we have an account to publish on the VSCode marketplace with but that won't change the code so this is ready for review.

wli3:
@sfoslund how often do we release? Download and reupload by hand is an error prone process. In general we should avoid doing that. But if it is less frequent that is ok. I don't know where to put it yet, but we need to document the manual step.

Although do we have to use the github release page? We can upload it to a predictable location and just put the link in readme like https://github.com/dotnet/core-sdk

sfoslund:
@wli3 No, we'll always be uploading to the VSCode Marketplace so if that's good enough then I don't see a reason to use GitHub releases too. 
I don't know how often we will be releasing, Livar is suggesting ~3 previews before the first release but I imagine it will be much less frequent after the first official release.

wli3:
@sfoslund I see, so we need a manual step anyway

sfoslund:
@wli3 I don't think so- uploading to the VSCode Marketplace is in this release pipeline so it's automated. If we don't feel the need for the GitHub releases page then we don't have any manual steps (unless we want to run manual smoke tests on the artifact, as we discussed earlier)

Waiting to merge this until I hear back about the publishing account so I can update the publisher in the package.json.

PR #37, opened 2020-01-24
Rename extension, update dependencies

sfoslund:
- Renaming extension to match new repo name
- Updating dependencies per [security alert](https://github.com/dotnet/vscode-dotnet-runtime/network/alert/dotnetcore-acquisition-library/package-lock.json/diff/open)
- Refactoring file structure in acquisition-library


azure-pipelines.yml
```diff
@@ -14,7 +14,7 @@ pr:
 
 variables:
   is-release: $[startsWith(variables['Build.SourceBranch'], 'refs/tags/v')]
-  package-name: 'dotnetcore-acquisition'
+  package-name: 'vscode-dotnet-runtime'
```
NTaylorMullen: 👏

dotnetcore-acquisition-extension/package.json
```diff
@@ -1,16 +1,16 @@
 {
-	"name": "dotnetcore-acquisition",
+	"name": "vscode-dotnet-runtime",
```
NTaylorMullen: Any reason you're not also changing the folder name `dotnetcore-acquisition-extension` -> `vscode-dotnet-runtime-extension`

sfoslund: Not in particular, I figure that it's not exposed externally so might as well keep it the same. I can update it if you think that's more clear.

NTaylorMullen: Would probably make it more clear to have the repo consistent but not a big deal either 😄

sfoslund: Latest commit has the folder names updated 👍

LLotme (commented):
gh pr checkout 37

PR #38, opened 2020-01-29
Update for publishing and marketplace display

### LINKED ISSUE #25: Do a future-proofing design pass on public API
Prior to the first release a pass on all of the public API should be done to ensure we can add/remove parameters/return values etc. in future releases. It'll be very difficult to onboard early adopters if we break the primary dotnet acquisition experience.

An example of future proofing this design involves returning an object from `dotnet.acquire` with a property for the dotnet.exe path. This way any future parameters we want to add as part of the return result wont be considered a breaking change. Same future proof design can be applied to parameters of common public API (take an object). All that being said, not every public API needs to be future proofed. For instance `uninstallAll` can continue to take 0 arguments and return nothing, each public API should be individually looked at to do "the right thing".

/cc @livarcocc

sfoslund:
- Updating documentation for display on VSCode marketplace
- Bundling extension (required adding a root package.json to avoid duplication of files required for marketplace display, such as README.md)
- Updating API for flexibility by taking and returning interfaces (Fixes #25)


README.md
```diff
@@ -1,58 +1,55 @@
...
+Extension authors do not know if the .NET core runtime is installed on the target machine. Existing solutions have a number of challenges:
 
-Today, extensions either ship or do dynamic acquisition of .NET Core on first activation. While it works, there are a number of challenges that have come up:
+1. **Duplication of .NET Core runtimes and slow updates**: Currently, each extension is acquiring its own copy of .NET core at ~30mb each.
+2. **Clean up**: When extensions install .NET Core in a non-VSCode folder location it is likely to be left behind.
+3. **Servicing and floating versions**: It is difficult to ensure that extensions will use the latest releases, particuarly without re-shipping.
+4. **Corrupted installations**: Corrupted installations can arise when VS Code is shut down mid-download or unzip.
+5. **Network security policies**: Alternative installation methods may result in errors due to blocking from network security policies.
+6. **Locked down envs**: Downloading a VSIX from the marketplace allows developers who are unable to freely install software to install extensions manually.
```
KathleenDollard: This item feels like it is written in the positive. The rest negative. I think negative works here.

README.md
```diff
@@ -1,58 +1,55 @@
...
-1. **Barrier to entry due to network security policies**: There have been a number of companies that "whitelist" sites instead of blacklisting them for security. As a result, developers have to request access to URIs to enable access. The VS marketplace is typically already unblocked if they are VS Code users while others may not be. 
-1. **Locked down envs**: There have been situations where developers are not allowed to freely download software. The ability to download a VSIX from the marketplace allows these customers to centrally acquire extensions and install them manually. Any dynamic acquisition experience does not easily work in these environments.
-1. **.NET Core missing dependencies**: VSCode runs on more platforms than .NET Core and therefore users run into situations where .NET Core cannot run as-is (generic containers, Alpine etc.) once dynamically acquired. This leads extension authors to detect these situations and attempt to prompt users to install missing pieces.
+This extension attempts to solve the above issues.
 
-This repo attempts to solve the above issues.
+## Contributing to Repository
```
KathleenDollard: Does this appear in the gallery or just in the repo?

sfoslund: It appears in both places. It seems to be the norm for gallery listings to have this kind of information ([example](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azurefunctions)).

PR #40, opened 2020-02-10
Adding easy issue reporting, updating branding

### LINKED ISSUE #24: Avoid unnecessary install script requests
Avoid making requests for install scripts with every acquisition (See this [PR comment](https://github.com/dotnet/vscode-dotnetcore/pull/23/files#r357738215))

sfoslund:
Changes are separated by commits:
- Fixes #24 by memoizing web requests
- Updating branding (display name, description, icon) and adding acquisition strategy documentation
- Adding easy issue reporting on error with redirection to GitHub:
Screenshot (12)


vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -21,9 +22,13 @@ export class WebRequestWorker {
         if (isNullOrUndefined(this.cachedData)) {
             // Have to acquire data before continuing
             this.cachedData = await this.makeWebRequest();
-        } else {
+        } else if (isNullOrUndefined(this.currentRequest)) {
```
wli3: In which case currentRequest is null?

wli3: Actually what is "currentRequest"? Why `this.makeWebRequest().then((result) => this.cachedData = result);` is blocking? it is using promise, no?

wli3:
The rest looks good. The error reporting mechanism is pretty fancy.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
sfoslund: It won't be null, this could just as easily be `=== undefined`. Is it generally better style to do that as opposed to `isNullOrUndefined`?

Since we don't await `this.currentRequest.then...`, it's not blocking, `then` just registers the callback and continues execution. This means that `currentRequest` will be defined until that callback is called, eliminating redundant requests.

Documentation/dev-spec.md
```diff
@@ -0,0 +1,39 @@
...
+- Users (extension authors) ensure that they have a runtime installed by calling the `acquire` command, providing a Major.Minor version to install.
+- Users are expected to call us on extension startup to ensure that the runtime exists.
+- Downloaded runtimes are stored in [extension global storage](https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage).
+
+## Scenarios
+
+### First Acquisition
+
+- Resolve version by fetching release.json
+  - If we are offline, we will fail here.
```
NTaylorMullen:
```suggestion
  - If we are offline, we will fail here because the user also will not be able to download a runtime offline.

```

Documentation/dev-spec.md
```diff
@@ -0,0 +1,39 @@
...
+Note: This scenario can be conducted offline successfully.
+
+### Subsequent Acquisitions (Runtime Patche Released)
+
+- Resolve version with the cached release.json
+  - We will update release.json in the background.
+  - **Note**: For efficiency, we do not block on release.json acquisition. This means that we may not install the most updated runtime until the following acquisition (once the release.json with the update has been acquired and cached).
+- Check if version has been installed previously or is currently being installed
+- Install runtime via cached install script
+  - We will update the install script in the background. If we are offline this will fail silently.
```
NTaylorMullen: Hmm, just food for thought there might be scenarios where we're trying to execute the install script but the update in the background happened so quickly the file gets corrupted on run.

vscode-dotnet-runtime-library/src/Utils/IssueReporter.ts
```diff
@@ -0,0 +1,21 @@
...
+    context.logger.dispose(); // Ensure log file is up to date
+    const issueBody = `<!-- IMPORTANT: Please be sure to remove any private information before submitting and attach the log file located at ${ context.logger.getFileLocation() }. -->
+
+**Repro steps:**
+
+1.
+
+**Error:** ${ error.stack }`;
+
+    const url = `${issuesUrl}/new?body=${encodeURIComponent(issueBody)}`;
```
NTaylorMullen: For long call stacks this will fail to work because of URI length restrictions. We (Razor + O#) have taken the approach of automatically copying the issue body to the users clipboard and notifying them to paste the body.

vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const errorMessage = 'An error occurred while installing .NET';
+const reportOption = 'Report an issue';
+
+export async function callWithErrorHandling<T>(callback: () => T, context: IIssueContext): Promise<T | undefined> {
+    try {
+        return await callback();
+    } catch (error) {
+        if (error.constructor.name !== 'UserCancelledError') {
+            window.showErrorMessage(`${errorMessage}: ${ error.message }`, reportOption).then((response: string | undefined) => {
```
NTaylorMullen: If there's an actual bug (even a non-user impacting bug) the extension and any dependent will be unusable with this error message repeatedly popping up in the users IDE. You should have an option on this dialog to "not show again" and in conjunction add a command that users can manually invoke to report an issue.

vscode-dotnet-runtime-library/src/Utils/IssueReporter.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function formatIssueUrl(error: Error, context: IIssueContext): string {
+    context.logger.dispose(); // Ensure log file is up to date
+    const issueBody = `<!-- IMPORTANT: Please be sure to remove any private information before submitting and attach the log file located at ${ context.logger.getFileLocation() }. -->
+
+**Repro steps:**
+
+1.
+
+**Error:** ${ error.stack }`;
```
NTaylorMullen: You also want the error message don't you?

vscode-dotnet-runtime-library/src/Utils/IssueReporter.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+import { IIssueContext } from './IIssueContext';
+
+const issuesUrl = `https://github.com/dotnet/vscode-dotnet-runtime/issues`;
+
+export function formatIssueUrl(error: Error, context: IIssueContext): string {
+    context.logger.dispose(); // Ensure log file is up to date
+    const issueBody = `<!-- IMPORTANT: Please be sure to remove any private information before submitting and attach the log file located at ${ context.logger.getFileLocation() }. -->
```
NTaylorMullen: Can you provide an example of what this would look like for a common error? I worry that there wont be enough identifying information to understand where errors are coming from for all of the components that create new processes etc.

vscode-dotnet-runtime-library/src/Utils/IssueReporter.ts
NTaylorMullen: I guess with the error message you'll have to do some privacy work to strip out user identifiable information. We do that [here](https://github.com/dotnet/aspnetcore-tooling/blob/201a91955f88000037f3fc72a58eb5aee941ec18/src/Razor/src/Microsoft.AspNetCore.Razor.VSCode/src/Diagnostics/ReportIssueCreator.ts#L144) in Razor's tooling

Documentation/dev-spec.md
sfoslund: Fair point, I've created #42 to track this.

vscode-dotnet-runtime-library/src/Utils/IssueReporter.ts
sfoslund: In my experimentation it looked like the error message always shows up in the stack trace, so it would be redundant. Is that not always the case?

NTaylorMullen: Not entirely sure. I think if an error is manually constructed (imagine VSCode internals throwing for unsupported scenarios) it might be different?

vscode-dotnet-runtime-library/src/Utils/IssueReporter.ts
sfoslund: Here's a screenshot of the issue message:

Screenshot (21)

And when you paste the error message (in this case, we failed to resolve the version `foo`):

Screenshot (20)

I've included a note about attaching the log file in the original message, which includes more debugging details. Do you think that's sufficient?

NTaylorMullen: Looks freaking awesome!

NTaylorMullen: Let me know when you want me to do another pass at the PR. I typically notify reviewers that I'm ready for another pass by commenting 🆙 📅 on my PR.

sfoslund: It's ready for another pass now- thanks!

PR #63, opened 2020-04-24
More information popup on timeout

### LINKED ISSUE #41: Fix timeout on linux functional test CI
We are seeing occasional CI failures due to functional tests timing out ([example, see attempt 1](https://dev.azure.com/dnceng/public/_build/results?buildId=526442&view=logs&j=d867d6e6-1e00-5f99-d13a-5f3fa11b81a0&t=06f6a5ae-854b-5480-ad18-711e339fe30a)). This is likely due to variability in host connectivity and speed when running checks. Need to increase timeout to give wiggle room for slow machines.

sfoslund:
First commit adds a popup on installation timeout to direct the user to troubleshooting information.
TimeoutPopup

Second commit fixes #41


NTaylorMullen:
I'm not convinced automatically showing this message is the right thing to do. Wouldn't extensions calling into dotnet-acquire want the option to show their own message or suppress ours?

Having a default message that can be shown when certain parameters are passed to acquire would be a better user experience in my opinion.

sfoslund:
@NTaylorMullen this was a request from the ARM extension team, but it's definitely true that not all extensions will want this behavior. What do you think about adding an optional argument to dotnet-acquire to disable showing this message?

NTaylorMullen:
> What do you think about adding an optional argument to dotnet-acquire to disable showing this message?

That'd be a great idea. I would think carefully about the type of argument though. It'd probably make sense to have a general error handling configuration parameter where this would be an option on that. Handling errors in the acquisition extension will be a very common-place procedure and each extension will probably want to do it differently.

sfoslund:
@NTaylorMullen It would make sense to me to use an enum:
```
enum ErrorConfiguration {
    DisplayAllErrorPopups,
    DisableErrorPopups
}
```
For now I don't see the need for more than these two options but this makes it easy to add to when other extensions want more specialized options.

sfoslund:
@NTaylorMullen @wli3 I added a `ErrorConfiguration` input for suppressing or enabling error popups in the last commit (0320170), let me know what you think.

sfoslund:
> Do you expect this configuration to expand to have multiple allowed configs?

Yes, probably. It would be good to have flexibility in case this is required later if nothing else. 

> Dangerous to share error configs with install/uninstall

Good point, I'll separate those out.

vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
```diff
@@ -14,22 +14,24 @@ export async function callWithErrorHandling<T>(callback: () => T, context: IIssu
...
-            window.showErrorMessage(`${errorConstants.errorMessage}: ${ error.message }`, errorConstants.reportOption, errorConstants.hideOption).then(async (response: string | undefined) => {
-                if (response === errorConstants.hideOption) {
-                    showMessage = false;
-                } else if (response === errorConstants.reportOption) {
-                    const [url, issueBody] = formatIssueUrl(error, context);
-                    await env.clipboard.writeText(issueBody);
-                    open(url);
-                }
-            });
+        if (context.errorConfiguration === ErrorConfiguration.DisplayAllErrorPopups) {
```
wli3, line 17: anyway to test these changes?

vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
```diff
@@ -4,28 +4,34 @@
...
-                    showMessage = false;
-                } else if (response === reportOption) {
-                    const [url, issueBody] = formatIssueUrl(error, context);
-                    await env.clipboard.writeText(issueBody);
-                    open(url);
-                }
-            });
+        if (context.errorConfiguration === ErrorConfiguration.DisplayAllErrorPopups) {
+            if ((error.message as string).includes(timeoutConstants.timeoutMessage)) {
+                window.showErrorMessage(`${errorConstants.errorMessage}: ${ error.message }`, timeoutConstants.moreInfoOption).then(async (response: string | undefined) => {
```
wli3, line 19: you could add an layer of abstraction for showing windows

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -16,6 +16,7 @@ import {
     DotnetAcquisitionUnexpectedError,
     DotnetOfflineFailure,
 } from '../EventStream/EventStreamEvents';
+import { timeoutConstants } from '../Utils/Constants';
```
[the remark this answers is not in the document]
sfoslund: I was patterning off another extension I was looking at. So you suggest defining constants in one of the files where they're used, even if they're used in many files? Or separate constant files for constants that are used together (reducing coupling between larger parts of the system)?

vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
sfoslund: Just pushed an update with tests added, I added a layer of abstraction for displaying popups, as you suggested.

sfoslund:
@wli3 @NTaylorMullen thanks for your feedback, I think I'm all caught up. Look okay now?

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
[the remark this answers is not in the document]
wli3: > defining constants in one of the files where they're used, even if they're used in many files?

Define the constants where it is used. If it is shared, it should be considered just like referencing a method. Does it have to be shared? Why the other component need to reference it? And what the dependency layering should be.

What I don't like here especially is the constants are put in a same file just for the sake that they are constants. And everything depends on one thing that they don't need to.

Although if it is the existing style, it would be hard to move away from (half and half is also bad). Maybe move const to the file only when it is shared.

PR #73, opened 2020-05-01
Fixing WebRequestWorker bugs

### LINKED ISSUE #68: Download fails on Windows 7 (keeps displaying the download message)
From ARM extension

Fail to download .NET Core tooling 3.1.3 on Windows 7 machine.
 
More info:
1.	This issue only reproduces on Win 7, not reproduce on Win 8.1 and Win 10.
2.	Changing the timeout value “dotnetAcquisitionExtension.installTimeoutValue” can’t fix this issue.
3.	It will keeps “Still downloading the .NET Core tooling” state after timeout.

### LINKED ISSUE #72: Web Request errors
3 common errors associated with the way we construct the web request for the install script:

> Request to https://dot.net/v1/dotnet-install.ps1 Failed: Error: unable to get local issuer certificate

> Error: Client network socket disconnected before secure TLS connection was established

> Error: socket hang up

sfoslund:
- 7e7c774 fixes #68 (note there isn't a good way to fix this issue without elevation so we're adding documentation)
- b5208fa fixes #72

### DIFF
--- a/vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
@@ -14,14 +14,14 @@ export class WebRequestWorker {
 
     constructor(private readonly extensionState: Memento,
                 private readonly eventStream: IEventStream,
-                private readonly uri: string,
+                private readonly url: string,
                 private readonly extensionStateKey: string) {}
 
     public async getCachedData(): Promise<string> {
         this.cachedData = this.extensionState.get<string>(this.extensionStateKey);
         if (isNullOrUndefined(this.cachedData)) {
             // Have to acquire data before continuing
-            this.cachedData = await this.makeWebRequest(true);
+            this.cachedData = await this.makeWebRequestWithRetries(true, 2);
         } else if (isNullOrUndefined(this.currentRequest)) {
             // Update without blocking, continue with cached information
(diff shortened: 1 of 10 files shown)


vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -38,7 +38,9 @@ export class WebRequestWorker {
     // Protected for ease of testing
     protected async makeWebRequest(throwOnError: boolean): Promise<any> {
         const options = {
-            uri: this.uri,
+            url: this.url,
+            strictSSL: false,
```
wli3, line 42: talk to Levi first?

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -58,4 +65,16 @@ export class WebRequestWorker {
     protected async cacheResults(response: string) {
         await this.extensionState.update(this.extensionStateKey, response);
     }
+
+    private async makeWebRequestWithRetries(throwOnError: boolean, retriesCount: number): Promise<any> {
+        for (let i = 0; i < retriesCount; i++) {
+            try {
+                return await this.makeWebRequest(throwOnError);
```
wli3, line 72: nit: we only retire 2 twice this maybe fine. But ideally you want to do exponential back off. It can reduce service load. Maybe there is an library to do that.

PR #78, opened 2020-05-04
Remove any type where possible. Prefer TS type system.

IEvangelist:


vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -9,15 +9,15 @@ import { IEventStream } from '../EventStream/EventStream';
...
+    private cachedData: string | undefined;
+    private currentRequest: Promise<string | undefined> | undefined;
 
     constructor(private readonly extensionState: Memento,
                 private readonly eventStream: IEventStream,
                 private readonly uri: string,
                 private readonly extensionStateKey: string) {}
 
-    public async getCachedData(): Promise<string> {
+    public async getCachedData(): Promise<string | undefined> {
```
sfoslund, line 20: Why is adding the undefined here necessary?

IEvangelist: The `this.extensionState.get<string>` returns the union `string | undefined`, making `this.cachedData` a union of `string | undefined`, which then makes the entire method that too.

vscode-dotnet-runtime-library/src/Acquisition/VersionResolver.ts
```diff
@@ -24,7 +24,7 @@ export class VersionResolver implements IVersionResolver {
     public async getFullVersion(version: string): Promise<string> {
         try {
             const response = await this.webWorker.getCachedData();
-            const releasesVersions = new ReleasesResult(response);
+            const releasesVersions = new ReleasesResult(response!);
```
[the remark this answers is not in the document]
IEvangelist: I'm good with that. I didn't really know what y'all expected here.

wli3:
Although is there any lint rule we could set?

sfoslund:
@wli3 Sure, I'll add that change to #74 (since I don't seem to have permission to push to this branch 😄)

PR #80, opened 2020-05-05
Adding configuration option to use existing .NET installation

sfoslund:
This is currently supported directly by the ARM extension but it belongs in this extension as a work around for failed installations or an option for users who don't want another .NET installation on their machines. See the ARM docs: https://github.com/microsoft/vscode-azurearmtools/wiki/Troubleshooting-DotNet-Acquisition#work-around-installing-dotnet-core-manually


vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
```diff
@@ -54,6 +55,11 @@ export class OutputChannelObserver implements IEventStreamObserver {
                     this.stopDownloadIndicator();
                 }
                 break;
+            case EventType.DotnetAcquisitionSuccessEvent:
+                if (event instanceof DotnetExistingPathResolutionCompleted) {
+                    this.outputChannel.append(`Using configured .NET path: ${ (event as DotnetExistingPathResolutionCompleted).resolvedPath }\n`);
```
NTaylorMullen, line 60: This is definitely a hammer. I might make this log appear more like a warning stating that there's a variable overriding dynamic acquisition. The reason why that's important is because users applications may fail because we're pointing our dotnet acquisition to some random .NET and not verifying it's the correct kind. Also, by overriding the acquired dotnet you opt-out of servicing which is also awful.

All that being said. This configuration option might need some more thought. The .NET acquisition extension is built to handle multiple extensions using it simultaneously. Imagine you have 3 different extensions using this extension and there's an override for dotnet? You're then assuming the one dotnet on the box can handle very single VSCode extension.

NTaylorMullen: Oh and to be clear I like how you allow multiple versions but you're then making the assumption that users know which version of dotnet to force. Also, having an "all" is dangerous for reasons above 😄

sfoslund: > I might make this log appear more like a warning stating that there's a variable overriding dynamic acquisition

Good idea, will do. 

> This configuration option might need some more thought.

Yeah, I was struggling to come up with a clear format for the configuration that balances usability and configuration for multiple extensions/versions. 
One other option is to specify the extension ID (ex. `msazurermtools.azurerm-vscode-tools`) instead of the version. Then the `all` option doesn't make sense anyways, so we can remove that and users don't have to know what version of .NET the extensions use. The downside here is that they won't know which version to manually install if they don't already have it on the machine.

NTaylorMullen: > The downside here is that they won't know which version to manually install if they don't already have it on the machine.

ya that's a big concern I have as well. To me instead of having this configuration option it makes me feel like there should be "installable" workloads that install directly into the VSCode extension directory for the .NET core acquisition extension. That way if users have trouble dynamically acquiring pieces you can point them directly to an offline install. With that approach you'd maintain the ability to service the "offline installed" dotnet's. Just a thought though.

sfoslund: I don't think anyone on our team would be thrilled about a new method of installation, it's hard to maintain.

If the main concern is that the user doesn't know which version of .NET to manually install, it shouldn't be that hard to expose that information to the user. It's already present in the output channel and we could add it to the error popup as well, something along the lines of `.NET Installation Failed (3.1): ...`?

NTaylorMullen: > I don't think anyone on our team would be thrilled about a new method of installation, it's hard to maintain.

Totally understand.



> If the main concern is that the user doesn't know which version of .NET to manually install, it shouldn't be that hard to expose that information to the user. It's already present in the output channel and we could add it to the error popup as well, something along the lines of `.NET Installation Failed (3.1): ...`?

You could but you also don't want users mistaking network blips as "I need to go and install something else offline".

sfoslund: > You could but you also don't want users mistaking network blips as "I need to go and install something else offline".

I don't necessarily see this as being any more encouraging to the user to try a work around than the existing behavior. This would just be adding the version number to the existing message, I wouldn't expect the average user to see that and jump to trying to manually install, unless the problem is recurrent. 

@wli3 do you have any thought here?

NTaylorMullen: > I wouldn't expect the average user to see that and jump to trying to manually install, unless the problem is recurrent.

That's fair. I only bring it up because O# who does similar acquisition logic has yet to make that leap to include the info in a pop-up and instead has README documentation in the GitHub repo on how to offline install their bits.

sfoslund: Yeah, I was thinking about how we could keep it in docs instead of the popup, but the problem there is that we would have to maintain a list of extensions who depend on us and which version they require instead of getting it dynamically based on what they request. It seems like it would be problematic to maintain such a list.

sfoslund: Also including the version in the error popup doesn't directly tell the user to manually install. In fact, it could make sense to do this in general- consider the case when two extensions are acquiring different versions and one succeeds and the other fails.

wli3: I am thinking about 2 ways.

1. instead of keeping track of which extension want which local build. We could override it by runtime version. And use the local runtime when the runtime version the extension asked is compatible.

2. I am now sure this using local runtime feature need to be generalized. We could add/keep this functionality in ARM extension to avoid regression. Due to the complexity mentioned above, each extension could decide if to use local runtime first and then run the acquisition extension.

sfoslund: @StephenWeatherford per the discussion above we are considering cutting this feature (the built in configuration option to point us to an existing .NET install). What is your motivation for wanting it here instead of in the ARM extension?

StephenWeatherford: The motivation is that the dotnet extension is supposed to be solving the installation problem for everyone.  But there are multiple known scenarios where your installation doesn't work (no PowerShell, group policies, old .NET, etc.).  Ideally the installation scripts should handle this correctly, or there should be options for using something other than PowerShell).  I can't imagine we're the only ones facing that problem.

So all I care about is that the dotnet installation should either work in all configurations or the should guide users on how to fix it, manually install, or work around it without each individual extension having to code for the scenario. Otherwise it defeats the whole purpose.  And I'm fine with offline docs as long as you point to them where the user can easily find them.

sfoslund: @StephenWeatherford Yes, I see where you're coming from and we are definitely aware of the issues with the install scripts, there is an active effort to improve them now. I agree that if there's a clean way to do this we should, the concern is that anything we can do on our end will be more confusing to the user than helpful. 

Just to lay things out more clearly for my own understanding, I see the main decision as which format we want to user to specify the configuration in: 
1. [runtime version, path] or 
2. [extension, path]. 

The complexity is that the user may not know which runtime a given extension needs.

@wli3 to clarify, your suggestion would be to have the user specify the version and location of the runtime (option 1), with the note that we would use a specified runtime as long as it is compatible (ie the version the user specifies does not have to exactly match the version the extension is requesting as long as they are compatible)? Do you have any thoughts on including the runtime version the extension is requesting in the error popup to give the user more guidance on which version to install?

wli3: > The motivation is that the dotnet extension is supposed to be solving the installation problem for everyone. But there are multiple known scenarios where your installation doesn't work (no PowerShell, group policies, old .NET, etc.)

I do like this long term vision. However, this is a very complex problem space. A large portion of the team's effort is spent on acquisition (via VS, docker, install script). However all of them have restrictions on different situations. It will be an on going effort to resolve these problems. And we do need to rank them with other bugs we have. cc @marcpopMSFT 

> Do you have any thoughts on including the runtime version the extension is requesting in the error popup to give the user more guidance on which version to install?

That should also be fine. If we cannot download them. It is still a path lead to solution of this problem

sfoslund: @wli3 I definitely agree that we need to rank new feature requests among the rest of our team's workload. In this case, it looks like we're coming to the conclusion that the right thing to do here is to allow the user to specify the version and the .NET path, which is what I've already implemented in this PR. It would take a few minutes to add the runtime version number to error popups, which mitigates the risk of users not knowing which runtime to install. Unless you see any other complexities here or someone disagrees that this is the best path forward, I think it's worthwhile for me to make that quick addition and then merge these changes, since I've already implemented this.

package.json
```diff
@@ -55,6 +55,17 @@
...
 					"description": "Timeout for installing .NET in seconds."
+				},
+				"dotnetAcquisitionExtension.existingDotnetPath": {
+					"type": {
+						"version": "string",
+						"path": "string"
+					},
+					"default": {
+						"version": "all",
+						"path": "C:\\Program Files\\dotnet"
```
StephenWeatherford, line 66: ??

Isn't this going to mean you'll try to look here by default?

package.json
```diff
@@ -55,6 +55,17 @@
...
+					},
```
StephenWeatherford, line 67: Also, what are your testing plans when you release?  I would hope it includes a quick sanity test of our extension.

sfoslund: We have set up manual testing with our vendors, they will be testing with our [sample extension](https://github.com/dotnet/vscode-dotnet-runtime/tree/master/sample), which essentially exposes our commands to the user. I will sanity test locally with the ARM extension as well before releasing.

package.json
sfoslund: I'll remove this, it's left over from my testing.

vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
marcpopMSFT: I see this extension as trying to help people get .net installed in most cases and Iike the fallback option this PR adds of them installing separately and specifying a version.  As Will notes, we have limited time and can't solve every install case in every tool that install .net core so just relying on the fallback and allowing for some failures in this extension make sense to me.

sfoslund:
@NTaylorMullen @wli3 I've added the version to the error popup so the user knows which version is required. Let me know if you have any other feedback, thanks!

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -48,6 +49,15 @@ export class DotnetCoreAcquisitionWorker {
         this.context.eventStream.post(new DotnetUninstallAllCompleted());
     }
 
+    public resolveExistingPath(existingPaths: IExistingPath[] | undefined, version: string): IDotnetAcquireResult | undefined {
+        if (existingPaths) {
+            const versionPath = existingPaths.filter((pair) => pair.version === version || pair.version.toLowerCase() === 'all');
```
wli3, line 54: I don't think we are closed on the previous discussion. Version matching is always an complex topic.

This implementation will not provide a good experience. This version comparison is precise. So the runtime need to be something like 3.1.**2**. And runtime bumps their patch version every month. The user or the consumer extension need to keep editing the config file to keep up. 

Maybe we should have a wild cast like 3.1.* however that is still not enough. We need to make sure the runtime on disk it actually higher than the runtime the consumer extension asked -- cannot ask 3.1.2 give 3.1.1. That why I think it is easier just to give a paths to existing runtime folders. Since the runtime compatibility is implicit already.

Just set the expectation. I don't think this is an issue we can close in a week. Do we need this in a short time? If not, let's talk more on that.

@sfoslund

sfoslund: > That why I think it is easier just to give a paths to existing runtime folders. Since the runtime compatibility is implicit already.

Can you clarify what you mean by this? Are you suggesting the user supplies a path without specifying the version and we attempt to use it without checking the version, just assuming it will fail if it's not compatible?

sfoslund: Summarizing a discussion with @wli3:

We don't want to duplicate the logic that already exists for determining if a runtime is compatible with an application in the extension for many reasons. If we ask the user to give us the runtime path and the runtime version then we don't think there's a way to avoid duplicating runtime selection without putting too much burden on the users. 

The other option is to ask the user for the runtime path and the extension they want to use it for. This way, the runtime selection will only happen once, the drawback being that this extension has no control over what that error experience will look like if the runtime is not compatible (since we will not do any compatibility checks on the runtime before handing it over to the extension that leverages us).

It seems like the second option is essentially equivalent to what the ARM extension is doing at the moment (taking a runtime path and assuming it will work), so I think this is an acceptable trade off, especially since this is not a major scenario and primarily seen as a work around if the user encounters other issues. 

@wli3 do you agree? If we decide to ask the user to specify the runtime path and extension and pass the path along without any compatibility checks then implementing it would be a quick change to this PR.

wli3: That sounds good. @NTaylorMullen is that good enough for your scenario?

BTW the error message from runtime is pretty useful. And it should be able to guide the user to download the right runtime. We could potentially read the runtime.configure.json and tell the user to find something higher(but do not cross major version) than the version specified.

```
It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '6.0.0' was not found.
  - The following frameworks were found:
      2.1.16 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
      2.1.18 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
      3.1.1 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
      3.1.2 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
      3.1.4 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
      5.0.0-preview.4.20251.6 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]

You can resolve the problem by installing the specified framework and/or SDK.
```

NTaylorMullen: > @NTaylorMullen is that good enough for your scenario?
To clarify the question being asked is "is it ok to prompt the user for an extension -> runtime mapping"?

wli3: My understanding is, the original reason for this ask: in case there is no internet or other error blocking downloading the runtime, we need a way to let the user to place the runtime on disk somehow and run the extension on it.

Does "providing a map of extension name, runtime path on disk, and then use the custom runtime location when running the extension" satisfy the ask above?

NTaylorMullen: I'd think it would but @StephenWeatherford would have to be the one to give the final "OK" since they're using this in production currently

Documentation/troubleshooting.md
```diff
@@ -16,6 +16,17 @@ Learn more about configuring Visual Studio Code settings [here](https://code.vis
 
 The .NET Install Tool for Extension Authors requires TLS 1.2 to be enabled in order to install .NET. For more information on TLS1.2, see [the documentation](https://docs.microsoft.com/mem/configmgr/core/plan-design/security/enable-tls-1-2-client).
 
+## Manually Installing .NET
+
+If .NET installation is failing or you want to reuse an existing installation of .NET, you can use the `dotnetAcquisitionExtension.existingDotnetPath` setting. .NET can be manually installed from [the .NET website](https://aka.ms/dotnet-core-download). To direct this extension to that installation, update your settings with the version and the path as illustrated below. Note that if the version is marked as `all` then this location will be used for any version of .NET requested.
+
+```json
+    "dotnetAcquisitionExtension.existingDotnetPath": [
+        {"version": "3.1", "path": "C:\\Program Files\\dotnet"},
```
StephenWeatherford, line 25: It's not clear what the user is pointing to - the folder?  the dotnet.exe file?  Having real-world examples for Mac and PC would be very helpful.

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
StephenWeatherford: Our current setting will override even calling to your extension so we won't be broken by this decision and am not too concerned.  My random thoughts:

- From the user's standpoint, it seems like specifying a mapping based on runtime version (not extension) is easiest (by major version + minor version only). That maps directly with what the extension can currently control, right?  You could always add the ability to override by extension later if necessary.  
- Nice to have: Allow the user to browse to the location on disk and locate the appropriate dotnet.exe file, then update the setting for them.
- It would be good if you could discover the full version pointed at and warn the user if a more recent point release is available (e.g. 3.1.2 is pointed at, but 3.1.3 is available).
- Agree with Taylor that having an "all" option is a bad idea.

sfoslund: Thanks for the feedback everyone. We've decided to take the mapping based on runtime version off the table, since this presents a lot of issues with ensuring runtime compatibility. It sounds like no one has strong feelings against the extension based mapping, so I can work on implementing that one (without the "all" option).

sfoslund: I've pushed these changes, the mapping is now based on the extension instead of the runtime version.

sample/src/extension.ts
```diff
@@ -37,7 +38,7 @@ export function activate(context: vscode.ExtensionContext) {
             await vscode.commands.executeCommand('dotnet.showAcquisitionLog');
 
             // Console app requires .NET Core 2.2.0
-            const commandRes = await vscode.commands.executeCommand<IDotnetAcquireResult>('dotnet.acquire', { version: '2.2' });
+            const commandRes = await vscode.commands.executeCommand<IDotnetAcquireResult>('dotnet.acquire', { version: '2.2', requestingExtensionId });
```
wli3, line 41: is this a public API change? We need an additional requestingExtensionId ? What if the user does not provide it?

vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -124,4 +132,21 @@ suite('DotnetCoreAcquisitionExtension End to End', function() {
...
+    assert.include(result!.dotnetPath, context.version);
+    assert.include(mockDisplayWorker.warningMessage, 'Ignoring existing .NET paths');
+  }).timeout(40000);
+
+  test('Install Command With Path Config Defined', async () => {
+    const context: IDotnetAcquireContext = { version: '0.1', requestingExtensionId: 'alternative.extension' };
+    const result = await vscode.commands.executeCommand<IDotnetAcquireResult>('dotnet.acquire', context);
+    assert.exists(result);
+    assert.exists(result!.dotnetPath);
+    assert.equal(result!.dotnetPath, 'foo');
```
wli3, line 150: You manual tested the real thing at least once? A mock in this level does not provide 100% confidence.

wli3 (approved):
Only 2 questions

sample/src/extension.ts
sfoslund: Yes, this is an API change for the extension calling us. If the extension calling acquire does not provide its own ID everything still works except they get this warning: https://github.com/dotnet/vscode-dotnet-runtime/pull/80/files#diff-4ac984a01117704ea53847e086aca20cR57 (cc @StephenWeatherford, as you'll need to accommodate this change once this version ships)

vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
sfoslund: Yes, I have tested manually

sample/src/extension.ts
wli3: Make sure you get an acknowledgement

StephenWeatherford: ETA on the release?  Thx.

StephenWeatherford: Actually, we're shipping in about a week.  I'm going to start passing it in then.

sfoslund: Not sure when the next release will be, we don't have a definite schedule. I would guess in a few weeks.

PR #127, opened 2020-10-01
Bump extension version to 0.2.0

sfoslund:
Updating the feature band as there was an API change for the dotnet.acquire command

Keeping this in draft PR for now until we get sign off from the manual testers


sfoslund:
We've gotten sign off from the manual testers, this should be good to go

wli3 (approved):

Issue #138, opened 2020-11-06
scriptAcquisitionUrl should use an aka.ms redirect link instead of pointing directly to the install script download location


StephenWeatherford:
This way it could be easily rolled back to a previous version if necessary.

It would also be nice to have a redirect link available for the troubleshooting page as well (currently ARM extension has a direct link, but the location could change in the future).

sfoslund:
@bozturkMSFT you already have tiered releases, right? And I seem to remember hearing that your team was going to make less investment in the scripts in the future, so we're unlikely to see any instability, is that correct?

StephenWeatherford:
Either way it's extremely easy to add a new aka.ms link, this is standard engineering practice.

bozturkMSFT:
@sfoslund Sorry, I have just seen this. No, we don't have tiered releases. We are putting the scripts in maintenance mode at the end of this year, but until then we have a bulk of changes coming. So aka.ms links sound like a good thing to have in the coming months.

Issue #137, opened 2020-11-04
.NET Acquisition Failed: undefined (0.2.2)


StephenWeatherford:
[DotNetAcquisition1604528413523.txt](https://github.com/dotnet/vscode-dotnet-runtime/files/5490784/DotNetAcquisition1604528413523.txt)

Privacy Alert! The contents copied to your clipboard may contain personal data. Prior to posting to GitHub, please remove any personal data which should not be publicly viewable. https://privacy.microsoft.com/en-US/privacystatement

**Repro steps:**

1.

**Extension Version:** 0.2.2
**Error Message:** .NET Acquisition Failed: undefined
**Error Stack:** Error: .NET Acquisition Failed: undefined
	at /Users/anonymous/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-0.2.2/dist/extension.js:211:111566
	at runMicrotasks (<anonymous>)
	at processTicksAndRejections (internal/process/task_queues.js:94:5)

sfoslund:
Post discussion notes:
- The wget failure is an issue with the install scripts, which is being tracked here: https://github.com/dotnet/install-scripts/issues/95
- Within the extension we should ensure that if the install scripts write to stderr we see those results in the error message

PR #142, opened 2020-11-12
Error reporting bug fixes

sfoslund:
3 commits fix 3 separate, small bugs:

- 72481d3: Include an event in the event stream for the ID of the extension that's calling us
- cc0fba1: Fixes https://github.com/dotnet/vscode-dotnet-runtime/issues/138
- d54ff3c: Fixes https://github.com/dotnet/vscode-dotnet-runtime/issues/137 (See the comment, the bug itself is fixed but we want to ensure the proper error message is surfaced in the popup).

### DIFF
--- a/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
@@ -289,6 +289,20 @@ export class DotnetInstallationValidated extends DotnetAcquisitionMessage {
     }
 }
 
+export class DotnetAcquisitionRequested extends DotnetAcquisitionMessage {
+    public readonly eventName = 'DotnetAcquisitionRequested';
+
+    constructor(public readonly version: string,
+                public readonly requestingId = '') {
+        super();
+    }
+
+    public getProperties() {
+        return {AcquisitionStartVersion : this.version,
+                RequestingExtensionId: this.requestingId};
+    }
+}
+
 export class TestAcquireCalled extends IEvent {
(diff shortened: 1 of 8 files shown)


vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
```diff
@@ -289,6 +289,20 @@ export class DotnetInstallationValidated extends DotnetAcquisitionMessage {
...
+export class DotnetAcquisitionRequested extends DotnetAcquisitionMessage {
+    public readonly eventName = 'DotnetAcquisitionRequested';
+
+    constructor(public readonly version: string,
+                public readonly requestingId = '') {
+        super();
+    }
+
+    public getProperties() {
+        return {AcquisitionStartVersion : this.version,
```
NTaylorMullen, line 301: Just to clarify, these correspond to telemetry events yes?

NTaylorMullen (approved):
Looks good.

In the future I'd recommend breaking each commit into its own PR. Even if the commit is small having separate PRs enables better "back tracking" for historical purposes; but more importantly it makes it easier to group corresponding code review comments. Meaning, if you react to code review feedback in bulk PRs like this it's not always easy to understand what piece of the PR that feedback corresponds to.

wli3 (approved):

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
sfoslund: Yup!

PR #148, opened 2020-11-18
Add command documentation

sfoslund:


Documentation/commands.md
```diff
@@ -0,0 +1,23 @@
+# Commands
+
+This article outlines the commands exposed by the .NET Install Tool for Extension Authors. To see these commands used in an extension, check out the [sample extension](https://github.com/dotnet/vscode-dotnet-runtime/tree/master/sample) in this repository. Note that the majority of these commands are not user facing, meaning they can only be called programatically using the [VSCode API](https://code.visualstudio.com/api/extension-guides/command#programmatically-executing-a-command) by other extensions.
+
+## dotnet.acquire
+
+This command will install a .NET runtime. It accepts a [IDotnetAcquireContext](https://github.com/dotnet/vscode-dotnet-runtime/blob/master/vscode-dotnet-runtime-library/src/IDotnetAcquireContext.ts) object and returns the path to the .NET runtime executable. Note that the version value in IDotnetAcquireContext must be a valid major.minor runtime version, for example 3.1. The extension will automatically identify and install the latest patch of the provided version. It is generally recommended that extension authors call this command immedietly on every extension start up to ensure that the .NET runtime has been installed and is ready to use.
```
NTaylorMullen, line 7: You briefly go over the format in which this command accepts versions but if you're documenting it there's probably value in documenting what's returned and why / the variations of the input. That being said, I don't know how deep you want to go on the documentation so feel free to ignore this 😄

wli3 (approved):

Documentation/commands.md
```diff
@@ -0,0 +1,23 @@
...
+
+## dotnet.uninstallAll
```
NTaylorMullen, line 9: Hmm, we probably shouldn't document this. Technically if an extension author has to do this then we've done something wrong. Moreover, given this extension is a singular extension that manages the .NET runtime we don't want to advertise ways other extensions can break experiences. Meaning, if extension A calls "uninstallAll" then extension B is indirectly affected.

Documentation/commands.md
```diff
@@ -0,0 +1,23 @@
...
+
+## dotnet.showAcquisitionLog
+
+This command surfaces an output channel to the user which provides status messages during extension commands. It does not accept parameters and has a void return type.
+
+## dotnet.ensureDotnetDependencies
+
+This command is only applicable to linux machines. It attempts to ensure that .NET dependencies are present and, if they are not, installs them or prompts the user to do so. It accepts a [IDotnetEnsureDependenciesContext](https://github.com/dotnet/vscode-dotnet-runtime/blob/master/vscode-dotnet-runtime-library/src/IDotnetEnsureDependenciesContext.ts) object and has a void return type.
+
+## dotnet.reportIssue
```
NTaylorMullen, line 21: Why would an extension author call this?

sfoslund: Probably wouldn't. I was just documenting it for completeness, but I'll remove it along with UninstallAll.

NTaylorMullen (approved):
👍

PR #162, opened 2021-02-04
Refactoring extension publishing

sfoslund:
Refactoring extension publishing to allow adding a second extension in this repo in the future. I've tested this change manually both locally and running the pipeline.


vscode-dotnet-runtime-extension/package.json
```diff
@@ -4,26 +4,70 @@
...
-		"onCommand:dotnet.ensureDotnetDependencies"
+		"onCommand:dotnet.ensureDotnetDependencies",
+		"onCommand:dotnet.reportIssue"
 	],
-	"main": "./out/extension.js",
+	"main": "./dist/extension.js",
+	"contributes": {
+		"commands": [
+			{
+				"command": "dotnet.reportIssue",
```
NTaylorMullen, line 40: Do you think we'd end up having a separate "report issue" for the SDK extension? Only asking because it's something we may need to start thinking about if it's the path we take so commands don't collide

vscode-dotnet-runtime-extension/package.json
```diff
@@ -4,26 +4,70 @@
...
+		],
+		"configuration": {
+			"title": ".NET Install Tool",
+			"properties": {
+				"dotnetAcquisitionExtension.enableTelemetry": {
+					"type": "boolean",
+					"default": true,
+					"description": "Enable Telemetry for the .NET install tool."
+				},
+				"dotnetAcquisitionExtension.installTimeoutValue": {
```
NTaylorMullen, line 53: Same Q

vscode-dotnet-runtime-extension/.vscodeignore
```diff
@@ -0,0 +1,8 @@
+.vscode/**
```
NTaylorMullen, line 1: All these wouldn't be awful to keep at top level would they? Otherwise a new extension would just replicate them all anyhow

vscode-dotnet-runtime-extension/.vscode/launch.json
```diff
@@ -5,7 +5,7 @@
 {
 	"version": "0.2.0",
 	"configurations": [{
-			"name": "Run Acquisition Extension",
+			"name": "Run .NET Runtime Acquisition Extension",
```
NTaylorMullen, line 8: Hmm, do you think your typical workflow in this repo will consist of you opening each individual folder in code or opening the workspace directory in code? Asking because I imagine there will be a lot of shared code among the two extensions in regards to acquisition.

NTaylorMullen (approved):

wli3 (approved):
Looks good to me(while I don't know too much in detail). Please merge once Taylor approves

vscode-dotnet-runtime-extension/package.json
sfoslund: I imagine we'll have to choose a different command prefix. It's unfortunate that we used `dotnet` and not something more specific to this extension, but changing that now would be breaking.

vscode-dotnet-runtime-extension/.vscode/launch.json
sfoslund: Yes, I imagine we'll be opening the workspace directly, which will include both extensions. This renaming ensures that, when there's another extension with its own launch tasks, we don't get confused between SDK and runtime tests.

vscode-dotnet-runtime-extension/package.json
NTaylorMullen: > but changing that now would be breaking.

It'd only be breaking if other extensions invoked it though right? I'd be curious if this was actually invoked by other extensions in the current world. At the very least Google doesn't seem to think any other OSS extension uses it: https://www.google.com/search?q=%22dotnet.reportIssue%22&rlz=1C1CHBF_enUS892US892&oq=%22dotnet.reportIssue%22&aqs=chrome..69i57j33i299.1498j0j7&sourceid=chrome&ie=UTF-8

vscode-dotnet-runtime-extension/.vscode/launch.json
NTaylorMullen: Can you have multiple .vscode folders/launch.json's in sub-folders? Haven't ever tried 😄

sfoslund: Yup, that's how it works right now with the sample extension vs the runtime extension.

vscode-dotnet-runtime-extension/package.json
sfoslund: Oh I guess that's true, I don't think any extensions would call it. Although just from pure aesthetics I don't love the idea of command prefixes not being consistent across the extension.

vscode-dotnet-runtime-extension/.vscode/launch.json
NTaylorMullen: Out of curiosity then, does it have a "smart" `{workspaceFolder}` understanding based on where the .vscode folder is?

sfoslund: Yes, for example: 
Screenshot (2)

PR #163, opened 2021-02-09
[Prototype] Adding SDK extension

sfoslund:
TODO:
- [x] Add/ expand SDK extension tests
- [x] Add installed SDK to `PATH`
- [x] Update SDK storage location to guard against long paths
  - Install to `%AppData%/.dotnet`
  - This means we will have to hook into the extension's uninstall to ensure we clean up after ourselves
- [x] Update CI to package SDK extension
- [x] Update publishing to marketplace pipeline for SDK extension
 
Prototype VSIX from demo: [SDKExtensionPrototype.zip](https://github.com/dotnet/vscode-dotnet-runtime/files/5954363/SDKExtensionPrototype.zip)

Demo (Sped up in the middle for time):
SDK Extension


sfoslund:
@wli3 @NTaylorMullen do you have any feedback on my general approach here before I expand on this PR? The first commit is mostly adding boilerplate for the new extension so the code is in 23b4f9f and a8c62d6

vscode-dotnet-sdk-extension/package.json
```diff
@@ -0,0 +1,107 @@
+{
+	"name": "vscode-dotnet-sdk",
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/dotnet/vscode-dotnet-runtime.git"
```
NTaylorMullen, line 5: lol we may need another repo rename 😆

vscode-dotnet-sdk-extension/package.json
```diff
@@ -0,0 +1,107 @@
...
+			"title": ".NET SDK Install Tool",
+			"properties": {
+				"dotnetSDKAcquisitionExtension.enableTelemetry": {
+					"type": "boolean",
+					"default": true,
+					"description": "Enable Telemetry for the .NET SDK install tool."
+				},
+				"dotnetSDKAcquisitionExtension.installTimeoutValue": {
+					"type": "number",
+					"default": 120,
```
NTaylorMullen, line 64: 2mins may not be enough for an SDK installation given they're pretty massive. I don't have a good number in mind but maybe Kathleen would?

NTaylorMullen (changes requested):
Hmmm, so if I read it all correctly you expanded the runtime library extension to serve as a psuedo extension that handles a general set of UI commands and interactions which are shared by both extensions but made "unique" by specific callbacks that can be registered by specific graced command providers.

I have some concerns with the approach because it couples both the runtime and SDK extensions pretty innately. Meaning any change to one of the core concepts needs to be thoroughly evaluated as it may impact the other. Also, given that a lot of the code that's runtime/sdk specific exists in the same place it's easy to conflate the two problems. Now all that said I can 100% understand why you went that path with the pre-existing runtime extension because it acted as the source of truth for the majority of the logic of the runtime extension.

So my suggestions are pretty drastic and feel free to schedule a follow up for us to talk more because it may be out-of-scope but I'd suggest:
1. The runtime-library extension should become just a library. Minimal VSCode interactions (no more vscode.registerCommand, issue reporting or output channel publishing). It provides classes that power all of those vscode interactions but doesn't directly interact with any of them.
2. Each sdk/runtime library copy & paste's the register command pieces in each so they're each individually maintained. This sounds like it'd be "more" work but in practice it'll be significantly less and will ultimately enable you to make product decisions on a per-product basis without the potential of influencing the other unintentionally. Those register command bits new-up classes and provide various information to them (extensionState, install dir etc.).
3. Each SDK/Runtime hookup to the EventSTreamObserver independently. If you want the runtime lib can have a default set of listeners that each extension can manually new-up to share the code but each one should be intentional in how it's listening to each event.

sfoslund:
@NTaylorMullen thanks for the feedback, that makes sense. I've gone ahead and factored out the command/ setup logic into the respective extensions, let me know if you have thoughts on these changes. I'm also getting some weird errors when I'm updating the sample extension, do you recognize these errors or see an issue with how I'm referencing the extension?

vscode-dotnet-runtime-library/src/Acquisition/IDotnetCoreAcquisitionWorker.ts
```diff
@@ -0,0 +1,18 @@
...
+ * ------------------------------------------------------------------------------------------ */
+
+import { IWindowDisplayWorker } from '../EventStream/IWindowDisplayWorker';
+import { IDotnetAcquireResult } from '../IDotnetAcquireResult';
+import { IExistingPath } from '../IExtensionContext';
+
+export interface IDotnetCoreAcquisitionWorker {
+    uninstallAll(): void;
+
+    resolveExistingPath(existingPaths: IExistingPath[] | undefined, extensionId: string | undefined, windowDisplayWorker: IWindowDisplayWorker): IDotnetAcquireResult | undefined;
```
NTaylorMullen, line 13: Hmm, shouldn't existing path's be compared against the currently requested version? Aka, if an extension that depends on the runtime extension "updates" and now requires a .NET 6 SDK where before they only required a 3.1 SDK how does this "resolve existing path" cross-check that the existing path version is correct? Very possible I'm misunderstanding this API but wanted to double check

vscode-dotnet-runtime-library/src/Acquisition/VersionResolver.ts
```diff
@@ -38,15 +44,15 @@ export class VersionResolver implements IVersionResolver {
...
 
-    private resolveVersion(version: string, releases: ReleasesResult): string {
+    private resolveVersion(version: string, releases: ReleasesResult, runtimeVersion: boolean): string {
         this.validateVersionInput(version);
 
         const channel = releases.releasesIndex.filter((channelVal) => channelVal.channelVersion === version);
         if (isNullOrUndefined(channel) || channel.length !== 1) {
             throw new Error(`Unable to resolve version: ${version}`);
         }
-        const runtimeVersion = channel[0].latestRuntime;
```
NTaylorMullen, line 48: I forget, why again is it `channel[0]`? Aka what do the other indexes represent in the channels array?

vscode-dotnet-runtime-library/src/EventStream/EventStreamRegistration.ts
```diff
@@ -0,0 +1,55 @@
...
+    telemetryReporter: ITelemetryReporter | undefined;
+}
+
+export function registerEventStream(context: IEventStreamContext): [EventStream, vscode.OutputChannel, LoggingObserver, IEventStreamObserver[]] {
+    const outputChannel = vscode.window.createOutputChannel(context.displayChannelName);
+    if (!fs.existsSync(context.logPath)) {
+        fs.mkdirSync(context.logPath);
+    }
+    const logFile = path.join(context.logPath, `DotNetAcquisition-${context.extensionId}-${ new Date().getTime() }.txt`);
+    const loggingObserver = new LoggingObserver(logFile);
```
NTaylorMullen, line 31: nit side note, looked into the logging observer and you currently build out a massive string to represent the log file that will eventually be written. Would be more performant to just store each log entry in an array and then call "WriteLine" effetcively on each of those once at "reportIssue" time.

vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
```diff
@@ -38,7 +38,7 @@ export class OutputChannelObserver implements IEventStreamObserver {
                 }
 
                 const startVersionString = this.inProgressDownloads.join(', ');
-                this.outputChannel.append(`Downloading .NET runtime version(s) ${startVersionString} ...`);
```
NTaylorMullen, line 41: I think indicating whether we're acquiring the runtime or SDK is still important. Might be worth publishing a separate event (so you don't need to update each of the subevents here) indicating that you're about to acquire an SDK or a runtime etc.

vscode-dotnet-sdk-extension/package.json
```diff
@@ -0,0 +1,110 @@
...
+		"is-online": "^8.2.1",
+		"mocha": "^8.1.3",
+		"open": "^7.0.2",
+		"rimraf": "2.6.3",
+		"shelljs": "^0.8.3",
+		"ts-loader": "^6.2.1",
+		"tslint": "^5.20.1",
+		"typescript": "3.4.5",
+		"vscode-dotnet-runtime-library": "file:../vscode-dotnet-runtime-library",
+		"vscode-dotnet-uninstall-library": "file:../vscode-dotnet-uninstall-library",
```
NTaylorMullen, line 102: Why have a separate uninstall library? Looking through the code is that this looks like it's only used at build time right? If that's the case why not just have a separate node.js script that you call `node` into?

vscode-dotnet-sdk-extension/src/extension.ts
```diff
@@ -0,0 +1,150 @@
...
+            extensionConfigWorker,
+            eventStream,
+            commandName,
+            version,
+        } as IIssueContext;
+    };
+    const timeoutValue = extensionConfiguration.get<number>(configKeys.installTimeoutValue);
+    let storagePath: string;
+    if (os.platform() === 'win32') {
+        storagePath = process.env.APPDATA ? process.env.APPDATA : context.globalStoragePath;
```
NTaylorMullen, line 83: Should leave a comment here as to why we utilize appdata as our storage path on windows

vscode-dotnet-sdk-extension/src/extension.ts
```diff
@@ -0,0 +1,150 @@
...
+                });
+            }
+            if (!version) {
+                displayWorker.showErrorMessage('No .NET SDK version provided', () => { /* No callback needed */ });
+                return undefined;
+            }
+
+            eventStream.post(new DotnetAcquisitionRequested(version!));
+            const resolvedVersion = await versionResolver.getFullSDKVersion(version!);
+            const dotnetPath = await acquisitionWorker.acquireSDK(resolvedVersion);
```
NTaylorMullen, line 118: The resolve existing path API doesn't seem to be used here. Just checking, acquireSDK won't try and re-download an already acquired SDK still correct?

vscode-dotnet-sdk-extension/package.json
sfoslund: I've doubled this to 240 for now, @KathleenDollard let me know if you have a better timeout value.

vscode-dotnet-runtime-library/package.json
```diff
@@ -39,6 +39,7 @@
 		"diff": ">=3.5.0",
 		"eol": "^0.9.1",
 		"es-abstract": "^1.18.0-next.1",
+		"fsevents": "^2.1.2",
```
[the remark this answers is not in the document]
sfoslund: This was a side effect of updating the node/npm version we're using, the older version was warning that this was an optional dependency and the new version turned this into an error.

vscode-dotnet-runtime-library/src/Acquisition/IDotnetCoreAcquisitionWorker.ts
sfoslund: For now I'm not including the existing paths configuration in this PR since it's relatively complex to ensure we're resolving a correct version and we don't particularly want the SDK extension to be used on machines that have global installs anyways. I can definitely reincorporate this, but I would prefer to do it as a separate PR (if nothing else, to make this one less massive :))

wli3:
Sorry I just started to read this PR

What's the interaction between runtime installed by SDK and the runtime installed by runtime extension? They are not in the same location right?

This SDK will only be used by VSCode?

What is the decision on the PATH? If it will be only used by VSCode, put on the PATH is complicated. Since users will not know they need to use VSCode extension to uninstall it. (if they install another SDK in programfiles that push the existing PATH off. It will be even more complicated)

vscode-dotnet-runtime-library/src/Acquisition/VersionResolver.ts
sfoslund: It's easiest to understand by looking at the structure of the file: https://github.com/dotnet/vscode-dotnet-runtime/blob/master/vscode-dotnet-runtime-library/src/test/mocks/mock-releases.json, each channel is associated with a major.minor .NET version. So here, since we filtered for the `channelVersion` we want (which would be `1.0`, `5.0`, `3.1`, etc) channel[0] is taking the first set of data associated with that major.minor version (there should only be one).

vscode-dotnet-sdk-extension/package.json
sfoslund: This library produces the node script that's called here: https://github.com/dotnet/vscode-dotnet-runtime/pull/163/files#diff-eec7f1511c1726a1883bdf93af89f838b72fd3fbf430edca98268b63123a6601R73, which hooks into VSCode's extension uninstall. It's kind of a weird hook in that it requires us to provide a script that's packed with the extension but with a separate entry point, so it's used at more than just build time.

I separated it out into a different library basically because of webpack- this way the extension and the uninstall scripts can be packed separately then inserted into the vsix together (logic for that [here](https://github.com/dotnet/vscode-dotnet-runtime/pull/163/files#diff-eec7f1511c1726a1883bdf93af89f838b72fd3fbf430edca98268b63123a6601R72) and [here](https://github.com/dotnet/vscode-dotnet-runtime/pull/163/files#diff-e29c8cf29668bc44ae83e958c39957ce6f85dcbf66cb35b56242ecbc62a7fc59R49)). I wasn't able to find a way to package them separately when they were sharing a package.json and webpack config, but I can definitely look more into this.

vscode-dotnet-sdk-extension/src/extension.ts
sfoslund: Yes, the resolve existing path logic had to do with the `[existingDotnetPath`](https://github.com/dotnet/vscode-dotnet-runtime/pull/163/files#diff-56eb173bf3821e528c6cb0129059298ebb627e3a9864955269f63684011d4fc1R59) configuration value, not the SDK we acquire.

sfoslund:
> Sorry I just started to read this PR

No problem, it's still in draft form so no need to do an in depth review yet, I still plan on making some changes. 

> What's the interaction between runtime installed by SDK and the runtime installed by runtime extension? They are not in the same location right?

No, they are not in the same location. This PR adds a separate SDK extension but tries not to touch the runtime extension. The runtime extension will continue to install to its global storage (something like `C:\Users\user\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.vscode-dotnet-runtime`, note that this path includes the runtime extension id) and the SDK extension will either install to its separate global storage (on mac or linux) or AppData on windows to avoid long path issues. 

> This SDK will only be used by VSCode? What is the decision on the PATH?

There's been a lot of discussion surrounding this and we have concluded that we will need to add to the PATH so that the SDK can be used on the command line. As a result we're hoping to limit the use of this extension to just the education bundle, which has a target audience that is not expected to have global installs of the SDK. That being said, I expect we'll continue to have discussion surrounding how to minimize the issues that would arise if a global install does exist.

vscode-dotnet-runtime-library/src/EventStream/EventStreamRegistration.ts
sfoslund: I think we already do this, the log only gets written to that file on dispose, which is called before reportIssue and on vscode close: https://github.com/dotnet/vscode-dotnet-runtime/blob/master/vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts#L27

vscode-dotnet-runtime-library/src/Acquisition/IDotnetCoreAcquisitionWorker.ts
NTaylorMullen: Ahhhh, this API is specific to SDK resolution then? In particular it's meant to resolve other SDKs on the path? If that's the case I'm not sure it makes sense for that type of orthogonal API to exist on the worker. Also 100% agree that it makes more sense in another PR

vscode-dotnet-runtime-library/src/Acquisition/VersionResolver.ts
NTaylorMullen: Ahhh so we're depending on index 0 effectively being the "latest" Alright good to know

vscode-dotnet-runtime-library/src/EventStream/EventStreamRegistration.ts
NTaylorMullen: Yes but it's still generating the "this.log" string behind the covers that sits in memory and is massive 😄

vscode-dotnet-sdk-extension/package.json
NTaylorMullen: Holllly hell, so VSCode's way to "hook" uninstall is to have a separate VSCode extension? That's nuts!

Having a separate extension here to enable webpack definitely feels odd. It'd be worth digging more into; I feel like this will be a huge point of confusion for repo onlookers

sfoslund: Not really another extension, just another node script: https://code.visualstudio.com/updates/v1_21#_extension-uninstall-hook. I'll try to look into ways to consolidate these webpacks

vscode-dotnet-runtime-library/src/Acquisition/IDotnetCoreAcquisitionWorker.ts
sfoslund: It's specific to the existing paths configuration value, which right now is only supported for runtimes. It basically just checks to see if the path(s) the user provided in their configuration are valid and if so returns the one we want. Yeah, that's a fair point, it might be better outside of the acquisition worker.

vscode-dotnet-sdk-extension/package.json
sfoslund: Yeah, I can't find a simple way to combine the webpack configs so I think we'll have to leave it for now. If I can figure out how  later when I have more time to look at it I'll definitely follow up with a PR to consolidate.

sfoslund: Made an issue for follow up work, as I imagine there will be a significant amount: https://github.com/dotnet/vscode-dotnet-runtime/issues/165

sfoslund:
@wli3 @NTaylorMullen I think this PR is complete enough to be a first iteration, I have an issue with tasks to follow up with in separate PRs: https://github.com/dotnet/vscode-dotnet-runtime/issues/165. Mind reviewing so we can merge?

For the most recent commit (821c790), I made the change to add the SDK to the PATH but I haven't found a good way to test it, do you have any suggestions? I tried asserting via `dotnet --info` but since the azure images have global installs of the SDK the local install doesn't show up. I also tried asserting the PATH itself but since it's all being done in test processes the environment variable changes aren't picked up. I've tested it manually and confirmed it works on a machine without any global installs, but I'm not sure if we'll be able to add test coverage.

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -3,11 +3,164 @@
...
+
+    context.subscriptions.push(
+        dotnetAcquireRegistration,
+        dotnetUninstallAllRegistration,
+        showOutputChannelRegistration,
+        ensureDependenciesRegistration,
+        reportIssueRegistration);
+    context.subscriptions.push({
+        dispose: () => {
+            for (const observer of eventStreamObservers) {
```
NTaylorMullen, line 161: Can't this now just be 
```TypeScript
const eventStreamObserversDisposable= vscode.Disposable.from(eventStreamObservers)
context.subscriptions.push(
        dotnetAcquireRegistration,
        dotnetUninstallAllRegistration,
        showOutputChannelRegistration,
        ensureDependenciesRegistration,
        reportIssueRegistration,
        eventStreamObserversDisposable);
```

vscode-dotnet-sdk-extension/src/extension.ts
```diff
@@ -0,0 +1,167 @@
...
+        open(url);
+    });
+
+    context.subscriptions.push(
+        dotnetAcquireRegistration,
+        dotnetUninstallAllRegistration,
+        showOutputChannelRegistration,
+        reportIssueRegistration);
+    context.subscriptions.push({
+        dispose: () => vscode.Disposable.from(...eventStreamObservers).dispose(),
```
NTaylorMullen, line 151: I don't think you need to wrap the disposable in another disposable, you can just push it along side the other registrations 😄

vscode-dotnet-sdk-extension/src/extension.ts
```diff
@@ -0,0 +1,167 @@
...
+            if (!version) {
+                displayWorker.showErrorMessage('No .NET SDK version provided', () => { /* No callback needed */ });
+                return undefined;
+            }
+
+            eventStream.post(new DotnetAcquisitionRequested(version!));
+            const resolvedVersion = await versionResolver.getFullSDKVersion(version!);
+            const dotnetPath = await acquisitionWorker.acquireSDK(resolvedVersion);
+            displayWorker.showInformationMessage(`.NET SDK ${version} installed to ${dotnetPath.dotnetPath}`, () => { /* No callback needed */ });
+            setPathEnvVar(path.dirname(dotnetPath.dotnetPath));
```
NTaylorMullen, line 127: Pull the `path.dirname(dotnetPath.dotnetPath)` into its own variable to ease debugability

vscode-dotnet-sdk-extension/src/extension.ts
```diff
@@ -0,0 +1,167 @@
...
+    let pathCommand: string;
+    if (os.platform() === 'win32') {
+        pathCommand = `setx PATH "${pathAddition};%PATH%"`;
+    } else {
+        pathCommand = `export PATH="${pathAddition};$PATH"`;
+    }
+    try{
+        cp.execSync(pathCommand);
+    } catch (error) {
+        throw new Error(`Unable to add SDK to the PATH: ${error}`);
```
NTaylorMullen, line 165: Because of how this is called wont this result in us prompting the user to file an issue? I feel like we should be utilizing a VSCode prompt here to suggest a help article or at the very least provide the path to the user so they can fully qualify their dotnet invocations. Kathleen probably has better opinions here than I though 😄

vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
```diff
@@ -24,14 +24,14 @@ export class LoggingObserver implements ILoggingObserver {
     }
 
     public dispose(): void {
-        fs.writeFileSync(this.logFilePath, this.log);
+        fs.writeFileSync(this.logFilePath, this.log.join('\n'));
```
NTaylorMullen, line 27: 🎉

vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
```diff
@@ -8,7 +8,7 @@ import { IEvent } from './IEvent';
 import { ILoggingObserver } from './ILoggingObserver';
 
 export class LoggingObserver implements ILoggingObserver {
-    private log = '';
+    private log: string[] = [];
```
NTaylorMullen, line 11: readonly

NTaylorMullen (approved):
Looks good. Only real concern was about how to handle the inability to add to the users path but I wont block on that. 👏 super excited to see this come through!

sfoslund:
@NTaylorMullen I've incorporated your feedback, thanks! 

I've done some more digging into adding to the PATH and made some changes there as well. I found that I was overwriting the user path with the system path with the way I had originally implemented this on windows, so I fixed that. For Mac, I've done a lot of research and I haven't found an easy way to do this. As far as I can tell the only ways to add to the path either require writing to an elevated file location or writing a .profile for the user, which feels hacky. @wli3 do you know of any better ways or have thoughts here?

NTaylorMullen:
> Unless anyone disagrees, I'm fine with merging this as is and making a separate PR to add mac support for updating the PATH.

I think that's fine. One question is how do Python/Java education bundles write to a users path on Mac? Maybe they don't but something we should follow up on /cc @mavasani @cartermp

cartermp:
Python doesn't have a pack for macOS, but Java does. I just cleared out my existing JDKs on my mac and verified that it:

1. Installs the [AdoptOpenJDK ](https://adoptopenjdk.net/)if you don't have one on your machine
2. Sets it on the PATH

And after installation in my existing terminal window I typed `java -version` and it had the correct JDK version, referring to what was just installed.

cartermp:
+1 for doing this incrementally BTW, we can figure out the MAC path issues as we go

sfoslund:
@cartermp do you know where the code Java uses for setting the path lives? I'd like to take a look and compare if possible.

sfoslund:
Talked with Will offline and he will add his review later to avoid blocking. Merging this now with Taylor's approval.

cartermp:
@sfoslund I think you could chat with @mavasani, since he has access to the repo.

LLotme (commented):
gh pr checkout 163

LLotme (commented):
gh pr checkout 163

PR #168, opened 2021-02-18
Update pipeline tag trigger pattern

sfoslund:
Missed the tagging pattern when adding the publishing pipeline


NTaylorMullen (approved):
👍

Issue #165, opened 2021-02-16
Follow up work on SDK Extension 


sfoslund:
- [x] If possible, consolidate SDK extension and uninstall library into one directory/ webpack config. 
- [ ] ~~Consider incorporating existing path configuration setting for SDK extension~~
- [x] Update sample extension to expose SDK commands
- [x] Write test plan for vendor test passes
- [x] Audit/ expand repo documentation
- [x] Add support for adding SDK to PATH on mac
- [x] Add test coverage for adding to PATH

sfoslund:
> Consider incorporating existing path configuration setting for SDK extension

For context, the existing path configuration in the runtime extension allows users to point to an existing .NET runtime install to avoid having the extension download a local copy. It's main goal was to be a workaround for any issues with installing a local copy. 

I would argue that we don't need a similar configuration setting for the SDK extension. The idea behind this extension is to install local copies on students' machines which will presumably not have global .NET installs. If we add this feature to support the global install case it feels like it's encouraging crossing streams in a way that could mess up the user's machine. 

@KathleenDollard @NTaylorMullen @wli3 let me know if you disagree, but for now I'm crossing this off the list as a won't fix.

wli3:
agree. For education bundle we should be more opinionated.

PR #171, opened 2021-02-19
Add SDK to PATH on mac and linux

sfoslund:
Fixes one of https://github.com/dotnet/vscode-dotnet-runtime/issues/165

### DIFF
--- a/vscode-dotnet-sdk-extension/src/extension.ts
+++ b/vscode-dotnet-sdk-extension/src/extension.ts
@@ -157,19 +157,24 @@ export function activate(context: vscode.ExtensionContext, extensionContext?: IE
 }
 
 function setPathEnvVar(pathAddition: string, displayWorker: IWindowDisplayWorker) {
-    if (process.env.PATH && process.env.PATH.includes(pathAddition)) {
-        // No need to add to PATH again
-        return;
-    }
-
     let pathCommand: string;
     if (os.platform() === 'win32') {
+        if (process.env.PATH && process.env.PATH.includes(pathAddition)) {
+            // No need to add to PATH again
+            return;
+        }
         pathCommand = `for /F "skip=2 tokens=1,2*" %A in ('%SystemRoot%\\System32\\reg.exe query "HKCU\\Environment" /v "Path" 2^>nul') do ` +
-                      `(%SystemRoot%\\System32\\reg.exe ADD "HKCU\\Environment" /v Path /t REG_SZ /f /d "${pathAddition};%C")`;
+            `(%SystemRoot%\\System32\\reg.exe ADD "HKCU\\Environment" /v Path /t REG_SZ /f /d "${pathAddition};%C")`;
+
     } else {
-        // Adding to the user path is only supported on windows right now
-        return;
+        const profileFile = os.platform() === 'darwin' ? path.join(os.homedir(), '.zshrc') : path.join(os.homedir(), '.profile');
+        if (fs.existsSync(profileFile) && fs.readFileSync(profileFile).toString().includes(pathAddition)) {
+            // No need to add to PATH again
+            return;
+        }
+        pathCommand = `echo 'export PATH="${pathAddition}:$PATH"' >> ${profileFile}`;
     }
+
     try {
         cp.execSync(pathCommand);
     } catch (error) {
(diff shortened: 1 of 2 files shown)


vscode-dotnet-sdk-extension/src/extension.ts
```diff
@@ -157,19 +157,24 @@ export function activate(context: vscode.ExtensionContext, extensionContext?: IE
...
+        }
         pathCommand = `for /F "skip=2 tokens=1,2*" %A in ('%SystemRoot%\\System32\\reg.exe query "HKCU\\Environment" /v "Path" 2^>nul') do ` +
-                      `(%SystemRoot%\\System32\\reg.exe ADD "HKCU\\Environment" /v Path /t REG_SZ /f /d "${pathAddition};%C")`;
+            `(%SystemRoot%\\System32\\reg.exe ADD "HKCU\\Environment" /v Path /t REG_SZ /f /d "${pathAddition};%C")`;
+
     } else {
-        // Adding to the user path is only supported on windows right now
-        return;
+        const profileFile = os.platform() === 'darwin' ? path.join(os.homedir(), '.zshrc') : path.join(os.homedir(), '.profile');
+        if (fs.existsSync(profileFile) && fs.readFileSync(profileFile).toString().includes(pathAddition)) {
```
NTaylorMullen, line 171: For debuggability break this into multiple sections. I'm betting this will be decently error prone just due to the nature of us touching more sensitive files 😄

NTaylorMullen (approved):

vscode-dotnet-sdk-extension/src/extension.ts
```diff
@@ -157,19 +157,24 @@ export function activate(context: vscode.ExtensionContext, extensionContext?: IE
...
+            return;
+        }
         pathCommand = `for /F "skip=2 tokens=1,2*" %A in ('%SystemRoot%\\System32\\reg.exe query "HKCU\\Environment" /v "Path" 2^>nul') do ` +
-                      `(%SystemRoot%\\System32\\reg.exe ADD "HKCU\\Environment" /v Path /t REG_SZ /f /d "${pathAddition};%C")`;
+            `(%SystemRoot%\\System32\\reg.exe ADD "HKCU\\Environment" /v Path /t REG_SZ /f /d "${pathAddition};%C")`;
+
     } else {
-        // Adding to the user path is only supported on windows right now
-        return;
+        const profileFile = os.platform() === 'darwin' ? path.join(os.homedir(), '.zshrc') : path.join(os.homedir(), '.profile');
```
wli3, line 170: Is there a prompt? rc file is very "private". We should not edit that write without explicit permission

vscode-dotnet-sdk-extension/src/extension.ts
```diff
@@ -157,19 +157,24 @@ export function activate(context: vscode.ExtensionContext, extensionContext?: IE
...
-
     let pathCommand: string;
     if (os.platform() === 'win32') {
+        if (process.env.PATH && process.env.PATH.includes(pathAddition)) {
+            // No need to add to PATH again
+            return;
+        }
         pathCommand = `for /F "skip=2 tokens=1,2*" %A in ('%SystemRoot%\\System32\\reg.exe query "HKCU\\Environment" /v "Path" 2^>nul') do ` +
-                      `(%SystemRoot%\\System32\\reg.exe ADD "HKCU\\Environment" /v Path /t REG_SZ /f /d "${pathAddition};%C")`;
+            `(%SystemRoot%\\System32\\reg.exe ADD "HKCU\\Environment" /v Path /t REG_SZ /f /d "${pathAddition};%C")`;
```
wli3, line 167: we should make sure it writes to user level Environment. https://github.com/dotnet/sdk/blob/b91b88aec2684e3d2988df8d838d3aa3c6240a35/src/Cli/Microsoft.DotNet.Cli.Utils/WindowsRegistryEnvironmentPathEditor.cs#L46

If vscode is elevated. It will cause a lot of problem.

And on ${pathAddition};%C. You should make sure it is not expand. So if there is a path

%USERPROFILE%\Other

wli3 (dismissed):
block on windows path until it is verified

vscode-dotnet-sdk-extension/src/extension.ts
sfoslund: Just verified that this writes to the same reg key as in the like, HKCU (the key we're using here) is shorthand for HKEY_CURRENT_USER, where the SDK code writes. 

> it should still be %USERPROFILE%\Other;newpath, instead of c:\user\myname\Other;newpath

Just verified that this is true with the current implementation.

vscode-dotnet-sdk-extension/src/extension.ts
sfoslund: As discussed offline, there isn't a prompt but we've decided that's fine since the extension will only be used by the education bundle which is targeted for students/ new developers who will likely not understand rc files.

wli3 (approved):

PR #184, opened 2021-02-25
Install to user home on unix

sfoslund:
Changing install location to 'match' windows installing to AppData. We won't have the path length problems on mac but when we have to eventually work with a offline bundled SDK we won't be able to do so from the vscode extension storage location.


NTaylorMullen (approved):
Really good catch 👏

Issue #178, opened 2021-02-25
Consider separate SDK update command


sfoslund:
Since we have decided that the education bundle will be the only user of the SDK extension and they will only be requesting one version of the SDK at a time, we could consider adding an explicit update command. With the runtime extension we achieve update behavior with repeated acquire commands but one drawback to this is that we don't remove old patch versions while updating to newer patches. This isn't much of a concern with runtimes, which are relatively small, but if we repeatedly acquire new SDK versions without removing the previous one(s) the size will quickly get unwieldly. 

If we can assume that only one SDK is needed at a time then implementing such a command should be simple as it would be a combination of the existing uninstallAll and acquire commands. 

@NTaylorMullen @KathleenDollard do you have any thoughts here?

NTaylorMullen:
> @NTaylorMullen @KathleenDollard do you have any thoughts here?

That'd absolutely be a valuable thing to have to "force" an update; however, it makes me ask the question: Would this be "instead" of automatic updating?

sfoslund:
> Would this be "instead" of automatic updating?

I'm not sure I follow the question. 

I think in general if we can make the assumption that we will only be managing one SDK at a time it will make things a lot cleaner. Another area I'm thinking about is the install directories and the fact that we're installing to `.dotnet/version` right now. This means that every time we install a new version we have to add it to the PATH separately, which feels like overkill.

NTaylorMullen:
> > Would this be "instead" of automatic updating?
> 
> I'm not sure I follow the question.
> 
> I think in general if we can make the assumption that we will only be managing one SDK at a time it will make things a lot cleaner. Another area I'm thinking about is the install directories and the fact that we're installing to `.dotnet/version` right now. This means that every time we install a new version we have to add it to the PATH separately, which feels like overkill.

oooo It wasn't you it was me. I misunderstood your original question. So I totally agree, there's no reason to support multiple SDKs IMO. Especially since they're supppposedddd to be backwards compat right? Aka a 6.0 can build >= 3.1 <=6.0 right?

So with that being said you could provide an "update my existing version" which in turn would be more like "move me to the next major" I imagine. Sounds reasonable.

sfoslund:
> Especially since they're supppposedddd to be backwards compat right? Aka a 6.0 can build >= 3.1 <=6.0 right?

Yup!

> So with that being said you could provide an "update my existing version" which in turn would be more like "move me to the next major or minor" I imagine. 

[...]

PR #185, opened 2021-02-26
Moving SDK to install directly to .dotnet folder

sfoslund:
Based on our discussion in https://github.com/dotnet/vscode-dotnet-runtime/issues/178, I think we should remove the prefix version from the SDK install path (instead of installing to `%AppData%\.dotnet\{version}`, install directly to `%AppData%\.dotnet`). This will have 2 main benefits: it will allow us to only have to add to the PATH once, regardless of how many SDK installations/ updates we make. With the current model we have to add a new directory to the path with every update, including patch updates, which will quickly get out of hand. Changing the install location will also allow some sharing across SDK versions, for example, if there are multiple SDKs installed they will all use the same dotnet.exe.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
@@ -2,6 +2,7 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
+import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
 import rimraf = require('rimraf');
@@ -21,15 +22,12 @@ import { IDotnetInstallationContext } from './IDotnetInstallationContext';
 
 export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker {
     private readonly installingVersionsKey = 'installing';
-    private readonly installDir: string;
     private readonly dotnetExecutable: string;
     private readonly timeoutValue: number;
 
     private acquisitionPromises: { [version: string]: Promise<string> | undefined };
 
     constructor(private readonly context: IAcquisitionWorkerContext) {
-        const installFolderName = process.env._VSCODE_DOTNET_INSTALL_FOLDER || '.dotnet';
-        this.installDir = path.join(this.context.storagePath, installFolderName);
         const dotnetExtension = os.platform() === 'win32' ? '.exe' : '';
         this.dotnetExecutable = `dotnet${dotnetExtension}`;
         this.timeoutValue = context.timeoutValue;
@@ -41,7 +39,7 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
 
         this.acquisitionPromises = {};
 
-        this.removeFolderRecursively(this.installDir);
+        this.removeFolderRecursively(this.context.installDirectoryProvider.getStoragePath());
 
         await this.context.extensionState.update(this.installingVersionsKey, []);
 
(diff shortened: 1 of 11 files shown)


vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -78,18 +77,21 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
...
+            this.context.eventStream.post(new DotnetAcquisitionPartialInstallation(version));
+            await this.uninstallAll();
         }
 
-        const dotnetInstallDir = this.getDotnetInstallDir(version);
+        const dotnetInstallDir = this.context.installDirectoryProvider.getDotnetInstallDir(version, this.installDir);
         const dotnetPath = path.join(dotnetInstallDir, this.dotnetExecutable);
 
-        if (fs.existsSync(dotnetPath)) {
+        if (this.context.installDirectoryProvider.isBundleInstalled(dotnetPath, version, this.context.extensionState, this.installingVersionsKey)) {
```
NTaylorMullen, line 94: Why does this need to be part of the provider? I'd imagine that the worker could get the install directory and then based off of that determine if it's installed something into it.

vscode-dotnet-runtime-library/src/Acquisition/IInstallationDirectoryProvider.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import { IExtensionState } from '../IExtensionState';
+
+export interface IInstallationDirectoryProvider {
+    getDotnetInstallDir(version: string, installDir: string): string;
+
+    isBundleInstalled(dotnetPath: string, version: string, extensionState: IExtensionState, installingVersionsKey: string): boolean;
```
NTaylorMullen, line 11: As mentioned above, this feels a little odd to have on this API

vscode-dotnet-runtime-library/src/Acquisition/IInstallationDirectoryProvider.ts
```diff
@@ -0,0 +1,12 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import { IExtensionState } from '../IExtensionState';
+
+export interface IInstallationDirectoryProvider {
+    getDotnetInstallDir(version: string, installDir: string): string;
```
NTaylorMullen, line 9: Dotnet is inferred

```suggestion
    getInstallDirectory(version: string, installDir: string): string;
```

NTaylorMullen: Why do you need to take in the installDir here? Isn't that the responsibility of this API?

I'd imagine each directory provider would do [this](https://github.com/dotnet/vscode-dotnet-runtime/pull/185/files#diff-62bfd0034cb508600b5a0b8a61f91033d2a9888830f40ff468155779594d958cR31-R32) magic and this API would then only take the `version: string` parameter

vscode-dotnet-runtime-library/src/Acquisition/SdkInstallationDirectoryProvider.ts
```diff
@@ -0,0 +1,19 @@
...
+import { IInstallationDirectoryProvider } from './IInstallationDirectoryProvider';
+
+export class SdkInstallationDirectoryProvider implements IInstallationDirectoryProvider {
+    public getDotnetInstallDir(version: string, installDir: string): string {
+        return installDir;
+    }
+
+    public isBundleInstalled(dotnetPath: string, version: string, extensionState: IExtensionState, installingVersionsKey: string): boolean {
+        const installingVersions = extensionState.get<string[]>(installingVersionsKey, []);
+        return installingVersions.includes(version) && fs.existsSync(dotnetPath);
```
NTaylorMullen, line 17: I still feel the isBundleInstalled logic should probably not exist on this API but leaving a comment here about this logic. Why do you care about the `installingVersions` member here for SDKs and not for runtimes?

NTaylorMullen (changes requested):

vscode-dotnet-runtime-library/src/Acquisition/SdkInstallationDirectoryProvider.ts
sfoslund: Good call, I've updated to remove it. It doesn't make a difference in terms of functionality for runtimes because runtimes are all installed to different folders so the fs.exists check is enough. Since SDKs are all installed to the same folder, we have to do more when checking if a version is installed.

NTaylorMullen: I see, so is `installingVersions` effectively `installedVersions`? Maybe I was thrown off by the naming too

sfoslund: > I see, so is installingVersions effectively installedVersions

Yes!

NTaylorMullen: Should change that at some point as I'm sure a future onlooker will be thoroughly confused like me 😄

NTaylorMullen (approved):

PR #186, opened 2021-03-02
Adding installed SDK to VSCode PATH

sfoslund:
Test VSIX: [SdkExtPathFix.zip](https://github.com/dotnet/vscode-dotnet-runtime/files/6070769/SdkExtPathFix.zip)

### DIFF
--- a/vscode-dotnet-sdk-extension/src/extension.ts
+++ b/vscode-dotnet-sdk-extension/src/extension.ts
@@ -140,11 +140,6 @@ export function activate(context: vscode.ExtensionContext, extensionContext?: IE
 }
 
 function setPathEnvVar(pathAddition: string, displayWorker: IWindowDisplayWorker, environmentVariables: vscode.EnvironmentVariableCollection) {
-    // Set PATH for VSCode terminal instances
-    if (!process.env.PATH!.includes(pathAddition)) {
-        environmentVariables.append('PATH', path.delimiter + pathAddition);
-    }
-
     // Set user PATH variable
     let pathCommand: string | undefined;
     if (os.platform() === 'win32') {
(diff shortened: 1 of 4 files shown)


JoeRobich (approved):

vscode-dotnet-sdk-extension/src/extension.ts
```diff
@@ -139,7 +139,13 @@ export function activate(context: vscode.ExtensionContext, extensionContext?: IE
         ...eventStreamObservers);
 }
 
-function setPathEnvVar(pathAddition: string, displayWorker: IWindowDisplayWorker) {
+function setPathEnvVar(pathAddition: string, displayWorker: IWindowDisplayWorker, environmentVariables: vscode.EnvironmentVariableCollection) {
+    // Set PATH for VSCode terminal instances
+    if (!process.env.PATH!.includes(pathAddition)) {
+        environmentVariables.append('PATH', path.delimiter + pathAddition);
+    }
+
```
NTaylorMullen, line 147: I wonder, if you set process.env.path here is it isolated to this extension or do others benefit? if it's the latter we should include that too

NTaylorMullen (approved):

vscode-dotnet-sdk-extension/src/extension.ts
sfoslund: Based on the documentation I don't think setting process.env here would have any effect: https://nodejs.org/dist/latest-v8.x/docs/api/process.html#process_process_env

NTaylorMullen: Hmm, I guess it depends on if all extensions in VSCode share the same process (which they might). If that's the case wouldn't O# see the path update and be able to use it?

sfoslund: Oh I assumed they wouldn't be on the same process but I can add that line in case. It will help ensure we don't add the PATH more than once anyways.

PR #189, opened 2021-03-04
Add logic to handle local preinstalled SDKs

sfoslund:
Add logic to pick up offline, local SDK installs from the education bundle installer, located in `%AppData%/.dotnet` on windows and `$HOME/.vscode-dotnet-sdk` on other platforms. cc @mavasani

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
@@ -14,6 +14,8 @@ import {
     DotnetAcquisitionStarted,
     DotnetAcquisitionStatusResolved,
     DotnetAcquisitionStatusUndefined,
+    DotnetPreinstallDetected,
+    DotnetPreinstallDetectionError,
     DotnetUninstallAllCompleted,
     DotnetUninstallAllStarted,
 } from '../EventStream/EventStreamEvents';
@@ -97,7 +99,7 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
     }
 
     private async acquireCore(version: string, installRuntime: boolean): Promise<string> {
-        const installedVersions = this.context.extensionState.get<string[]>(this.installingVersionsKey, []);
+        let installedVersions = this.context.extensionState.get<string[]>(this.installingVersionsKey, []);
         const partialInstall = installedVersions.indexOf(version) >= 0;
         if (partialInstall && installRuntime) {
             // Partial install, we never updated our extension to no longer be 'installing'.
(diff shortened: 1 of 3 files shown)


vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -167,4 +174,21 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
         this.context.eventStream.post(new DotnetAcquisitionDeletion(folderPath));
         rimraf.sync(folderPath);
     }
+
+    private async managePreinstalledVersion(dotnetInstallDir: string, installedVersions: string[]): Promise<string[]> {
```
NTaylorMullen, line 178: Don't love the name but don't have a better idea after trying for a while 😆

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -167,4 +174,21 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
...
+        try {
+            // Determine installed version(s)
+            const versions = fs.readdirSync(path.join(dotnetInstallDir, 'sdk'));
+
+            // Update extension state
+            for (const version of versions) {
```
NTaylorMullen, line 184: Totally curious, does anything catastrophic happen if there's an extraneous folder in that dir that we're unaware of?

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -113,6 +115,11 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
         const dotnetInstallDir = this.context.installDirectoryProvider.getInstallDir(version);
         const dotnetPath = path.join(dotnetInstallDir, this.dotnetExecutable);
 
+        if (fs.existsSync(dotnetPath) && installedVersions.length === 0) {
+            // A local install already exists, add it to our managed installs
```
NTaylorMullen, line 119: I'd expand this comment because it's a little ambiguous but inherently dependent on the idea that an external asset (the edu bundle) has done something that we can react to. It's not that just any pre-install was detected (aka we're not looking for Admin installs) it's that a "local install" was detected as in someone copy and pasted some content into the VSCode folder.

NTaylorMullen (approved):

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
sfoslund: Theoretically no, we will add it to our installed versions even if it's an invalid version (or not a version at all) but then it would never make a functional difference unless an extension requests to acquire that invalid version and somehow it manages to get through version resolution, in which case we would just return the path to the existing dotnet.exe.

wli3 (approved):

PR #198, opened 2021-03-16
Replacing master branch with main

sfoslund:


azure-pipelines.yml
```diff
@@ -3,6 +3,7 @@ trigger:
   branches:
     include:
     - master
```
NTaylorMullen, line 5: Why still include master?

NTaylorMullen (approved):

azure-pipelines.yml
sfoslund: It's part of the instructions, I'll remove it once the final move is finished.

PR #231, opened 2021-05-18
Bump SDK extension timeout

sfoslund:


vscode-dotnet-sdk-extension/package.json
```diff
@@ -59,7 +59,7 @@
 				},
 				"dotnetSDKAcquisitionExtension.installTimeoutValue": {
 					"type": "number",
-					"default": 240,
+					"default": 300,
```
NTaylorMullen, line 62: Interesting, I guess slower networks bring this into play? Out of curiosity is the 1min increase based on a specific metric or a magic number?

NTaylorMullen (approved):

vscode-dotnet-sdk-extension/package.json
sfoslund: No specific metric, just seems to make sense to me

PR #284, opened 2021-08-18
Remove install script redirects

sfoslund:
Hopefully fixing https://github.com/dotnet/vscode-dotnet-runtime/issues/114 based on this suggestion: https://github.com/dotnet/vscode-dotnet-runtime/issues/114#issuecomment-898665972. This should be a no-risk fix since the install scripts team now have test coverage before they release updates to the scripts in this link.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/InstallScriptAcquisitionWorker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/InstallScriptAcquisitionWorker.ts
@@ -18,7 +18,7 @@ import { IInstallScriptAcquisitionWorker } from './IInstallScriptAcquisitionWork
 
 export class InstallScriptAcquisitionWorker implements IInstallScriptAcquisitionWorker {
     protected webWorker: WebRequestWorker;
-    private readonly scriptAcquisitionUrl: string = 'https://aka.ms/dotnet-install-script-';
+    private readonly scriptAcquisitionUrl: string = 'https://dot.net/v1/dotnet-install.';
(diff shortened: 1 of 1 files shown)


vscode-dotnet-runtime-library/src/Acquisition/InstallScriptAcquisitionWorker.ts
```diff
@@ -18,7 +18,7 @@ import { IInstallScriptAcquisitionWorker } from './IInstallScriptAcquisitionWork
 
 export class InstallScriptAcquisitionWorker implements IInstallScriptAcquisitionWorker {
     protected webWorker: WebRequestWorker;
-    private readonly scriptAcquisitionUrl: string = 'https://aka.ms/dotnet-install-script-';
+    private readonly scriptAcquisitionUrl: string = 'https://dot.net/v1/dotnet-install.';
```
NTaylorMullen, line 21: Hmmm, this is an interesting choice. I get the harm in redirects; however, having an aka.ms link allows us to move that link whenever we want. Do we have a guarantee that the `https://dot.net/v1/dotnet-install` is stable?

sfoslund: Yes, when I originally added this redirect there wasn't much infrastructure around the install scripts but now they have a multi-teired system of releasing new changes, so that this link is only updated once the new script has been dogfooded by the other tiers first.

NTaylorMullen (approved):

PR #319, opened 2021-10-18
Escape \ in Windows path to .NET SDK

SvenAelterman:


sfoslund (commented):
@SvenAelterman thanks for your contribution! I'm hesitant to take this since I expect some users will directly copy the json snippet in this sample into their settings files and if we have escaped `\`'s this feature will fail. Did you think this change was necessary due to the bad formatting in the page? Or was there another reason you think it's needed?

Issue #331, opened 2021-11-07
.NET Runtime is downloaded / installed every time


josetr:
`const result = await vscode.commands.executeCommand('dotnet.acquire', { version: '3.1', requestingExtensionId });`

```
Downloading the .NET Runtime.
 Error!
ENOENT: no such file or directory, scandir '%userprofile%\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.vscode-dotnet-runtime\.dotnet\3.1.20\sdk'

Downloading .NET version(s) 3.1.20 ..... Done!
.NET 3.1.20 executable path: %userprofile%\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.vscode-dotnet-runtime\.dotnet\3.1.20\dotnet.exe
```

It happens every time VSCode is restarted. It doesn't happen if I execute the `dotnet.acquire` command twice in a row.

josetr:
Changed the issue title. Added the following workaround and I managed to get rid of the 'sdk' error but the .NET Runtime is still downloaded every time.

```ts
const dotnetDir = path.dirname(dotnetPath)
const dotnetSdkDir = path.join(dotnetDir, 'sdk')

if (!fs.existsSync(dotnetSdkDir))
    fs.mkdirSync(dotnetSdkDir)
```

sfoslund:
@josetr I think this might be the expected behavior. Based on the number of dots, it looks like this 'download' doesn't take too long, is that right? Usually, when the runtime is really being installed, it can take up to 2 minutes. So it looks like what's happening here is that when it says that it's downloading .NET, it's really checking that .NET is installed and checking that it is up to date.

josetr:
The "downloading .NET" message seems pretty misleading then and probably should be changed.

The reason why I thought it was being downloaded over and over is that the acquire command takes 1.5 seconds to complete in my Ubuntu VM and 750ms on my main PC.

What's happening exactly that takes that long? Is it checking every file to see if one is corrupted?

I'm also now getting the error `"Failed to run .NET runtime. You may be missing key Linux libraries. Install them now?"` in my Ubuntu VM which is a bit misleading once again because if I just close it by clicking on 'Cancel', my dotnet app will run without any problems.

sfoslund:
Thanks for your feedback, yes, we could probably update that string. 

> What's happening exactly that takes that long? Is it checking every file to see if one is corrupted?

[...]

PR #335, opened 2021-11-09
Adding acquireStatus command to runtime extension

sfoslund:
Fixes https://github.com/dotnet/vscode-dotnet-runtime/issues/331

### DIFF
--- a/sample/package.json
+++ b/sample/package.json
@@ -21,10 +21,12 @@
 	"activationEvents": [
 		"onCommand:sample.helloworld",
 		"onCommand:sample.dotnet.acquire",
+		"onCommand:sample.dotnet.acquireStatus",
 		"onCommand:sample.dotnet.uninstallAll",
 		"onCommand:sample.dotnet.concurrentTest",
 		"onCommand:sample.dotnet.showAcquisitionLog",
 		"onCommand:sample.dotnet-sdk.acquire",
+		"onCommand:sample.dotnet-sdk.acquireStatus",
 		"onCommand:sample.dotnet-sdk.uninstallAll",
 		"onCommand:sample.dotnet-sdk.showAcquisitionLog"
 	],
@@ -39,6 +41,11 @@
 			{
 				"command": "sample.dotnet.acquire",
 				"title": "Acquire .NET runtime",
+				"category": "Sample"
+			},
+			{
+				"command": "sample.dotnet.acquireStatus",
+				"title": "Check the status of acquiring the .NET runtime",
 				"category": "Sample"
 			},
 			{
@@ -59,6 +66,11 @@
 			{
 				"command": "sample.dotnet-sdk.acquire",
 				"title": "Acquire .NET SDK",
+				"category": "Sample"
+			},
+			{
+				"command": "sample.dotnet-sdk.acquireStatus",
+				"title": "Check the status of acquiring the .NET SDK",
 				"category": "Sample"
 			},
 			{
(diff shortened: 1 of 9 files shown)


NTaylorMullen (approved):
Looks great! I'm a bit hesitant about adding it also to the SDK side but it's minimal work so why not? 😆

sfoslund:
> I'm a bit hesitant about adding it also to the SDK side but it's minimal work so why not?

This command already exists on the SDK side, I'm adding it to the runtime extension since we've had some users requesting it.

josetr:
Thanks for the quick fix. Any idea when this update will be available in the marketplace?

sfoslund:
I'll release a new version at some point this week, probably tomorrow. I'll ping here when it's released.

sfoslund:
Version 1.4.0 has been released, please let us know if you have any problems, thanks!

gulbanana:
I'm not sure the change is working as intended. I debugged the new release, and found that in `DotnetCoreAcquisitionWorker.acquireStatus`, execution reaches this line: https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts#L78
```
        if (installedVersions.includes(version) && fs.existsSync(dotnetPath)) 
```

At this point, `dotnetPath` has been calculated correctly and `fs.existsSync(dotnetPath)` returns `true`, but `installedVersions` is an empty array, so the function returns undefined. It looks like installedVersions gets set by `acquireCore()` - so you'd have to call dotnet.acquire before dotnet.acquireStatus, which defeats the purpose?

sfoslund:
> installedVersions is an empty array, so the function returns undefined. It looks like installedVersions gets set by acquireCore() - so you'd have to call dotnet.acquire before dotnet.acquireStatus, which seems to defeat the purpose.

installedVersions in this context is saved as part of the extension state, so it's persistent even when the extension shuts down/ starts up. Do you have a repro where the new command is behaving unexpectedly?

josetr:
@sfoslund I can confirm that @gulbanana is right. `acquireStatus` only works after calling the `acquire` command. In the following example, the first `dotnetPath not found` message is shown every time you start vscode and the message `dotnetPath still not found` is never shown.

```ts
let commandResult = await vscode.commands.executeCommand('dotnet.acquireStatus', { version: version, requestingExtensionId });
if (!commandResult?.dotnetPath) {
    vscode.window.showInformationMessage("dotnetPath not found");

    commandResult = await vscode.commands.executeCommand('dotnet.acquire', { version: version, requestingExtensionId });
    if (!commandResult?.dotnetPath;)
        throw new Error('Could not resolve dotnet path!');

    commandResult = await vscode.commands.executeCommand('dotnet.acquireStatus', { version: version, requestingExtensionId });
    if (!commandResult?.dotnetPath)
        vscode.window.showInformationMessage("dotnetPath still not found!");
}
```

gulbanana:
i didn't get time to put together a repro yet, sorry, but @josetr's code produces the same result for me. i do see that installedVersions should be coming from state... i wonder why it's empty

sfoslund:
> i do see that installedVersions should be coming from state... i wonder why it's empty

I'm wondering the same thing. Investigating this is on my backlog, I hope to get to it some time this week, sorry for the delay.

sfoslund:
Just released a new version (1.5.0) with this issue fixed, thanks for pointing it out!

PR #344, opened 2021-11-17
Separate extension state installation records

sfoslund:
It looks like the bug reported [here ](https://github.com/dotnet/vscode-dotnet-runtime/pull/335#issuecomment-967717720)was a mix up with how we record installation versions. We were using installingVersions extension state to check if a version was installed but that state actually represented which versions were in the process of being installed. To fix this, I've added a new extension state item to keep track of which versions we have installed.


NTaylorMullen (approved):
Ahhh I recall when this code was originally added. Yes having an installed key makes sense :shipit:

PR #432, opened 2022-04-07
Use axios to handle timed cache of http requests

### LINKED ISSUE #416: Acquisition isn't getting the most recent version of the SDK
The [releases index](https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json) currently advertises 6.0.201 as the most recent version of the SDK, but the extension is currently acquiring 6.0.102 instead.  This should be changed - the [Dotnet Pack](https://github.com/dotnet/vscode-dotnet-pack/blob/main/src/extension.ts#L15) is requesting 6.0, and so it should get the latest 6.0 SDK by default.

baronfel:
This should fix #416, but it needs more testing.


nagilson:
Omg this would've fixed the install scripts issue a while ago

baronfel:
I would love to see this cleaned up and used again - there's still a long-running issue in the extension that the _first_ cached version of the scripts are used, but they can and do float over time.

PR #512, opened 2022-06-28
Remediate npm configuration issues

nagilson:
I will likely be maintaining this codebase now. It's been a while since anyone took a look at it. Arcade folks reached out because we need to use their new registries. The main fix involved was adding this registry to .npmrc files but this change also required updates to many packages (unfortunately a huge change.) Some changes required updates to annotations to allow compiling. Lastly, windows 2016 is deprecating tomorrow so all builds failed, so I ended up having to put that fix in here too so we can push this out swiftly. Sadly, this was going to be a mess of a PR either way to get things building as main is not passing for several reasons and I figured it best to put it all in one instead of pushing several broken PRs to main.


azure-pipelines.yml
```diff
@@ -67,16 +67,14 @@ stages:
...
-        versionSpec: '15.x'
-      displayName: 'Install Node.js'
-    - bash: |
-        npm install -g tslint
-        npm run lint
-      displayName: Run Lint
+    - task: npmAuthenticate@0
+    inputs:
+      workingFile: .npmrc
+      displayName: 'Authenticate NPM'
```
rainersigwald, line 75:
```suggestion
      inputs:
        workingFile: .npmrc
        displayName: 'Authenticate NPM'
```

nagilson:
Closing because this PR is a mess and doing it more cleanly here now https://github.com/dotnet/vscode-dotnet-runtime/pull/526

PR #528, opened 2022-07-15
Switch to `npm ci`

rainersigwald:
This will lock all versions to what is specified in the lock file.


rainersigwald:
This is probably destined to fail until #526 is merged, but it should be helpful afterward.

nagilson (approved):

PR #595, opened 2022-09-15
Pipeline Registry Fixup: Update Chai and Add VSCE to package-lock

nagilson:
- Lock the chai versions to older versions, they weren't locked properly in the .npmrc migration
- Add vsce and dependencies so the build task doesn't encounter 401 unauthorized. 
- Add some stuff to gitignore
- Sync up package lock jsons to the new package json content in upstream


nagilson:
Normally wouldn't merge without review but this needs to get into a working state.

PR #617, opened 2022-10-06
Add Step To Verify Yarn and NPM Packages Are In Sync

nagilson:
MERGE AFTER https://github.com/dotnet/vscode-dotnet-runtime/pull/616

With the new registry changes, we need to have a yarn.lock, otherwise yarn will update dependencies differently from npm and fail with auth errors. The reason we use two package managers is because of unsupported features in VSCE with npm so we can't only use npm, and AzDo does not support yarn, so we cannot only use yarn. We don't want the two to get out of sync, so this adds a checker that checks the diff of a PR to see if any changes made in npm locks are reflected in yarn locks. 

Limitations:
- It doesn't sync them for you, you need to sync them after it fails the pipeline.
- If someone updates the versions incorrectly but both have been updated, the pipeline will not fail. Extracting the version from the package-lock is not super trivial, though also not that hard to do later.
- The target branch is currently fetched from origin/, in a forked branch they must have a copy of the target branch, otherwise the sync will be more aggressive. We could add a custom remote in the script instead of calling into origin/ but that would break PRs on top of forks


NTaylorMullen:
Can you refresh me on why we have both npm and yarn ?

NTaylorMullen:
Forget me, I didn't read yopur description before asking like a dumb dumb: `The reason we use two package managers is because of unsupported features in VSCE with npm so we can't only use npm, and AzDo does not support yarn, so we cannot only use yarn. `

azure-pipelines.yml
```diff
@@ -78,6 +78,23 @@ stages:
         npm install tslint --verbose
         npm run lint
       displayName: Run Lint
+  ##### Verify NPM and Yarn are in sync #####
+  - job: SyncPackageManagers
+    displayName: 'Verify NPM & Yarn In-Sync [Local Copy of Target Branch Must Be Up to Date]'
+    pool:
+      vmImage: 'windows-latest'
+    steps:
+    - task: UsePythonVersion@0
```
NTaylorMullen, line 87: Interesting, out of curiosity why use Python? Familiarity I assume?

NTaylorMullen (approved):

azure-pipelines.yml
nagilson: Yea, familiarity + speed of development time. Writing this in bash sounds... not fun. C# is an option though it's also not really in the repository much so either way it's an introduced new onboarding cost. My hope is someday VSCE will support npm and this script can be eradicated, but since their issues for NPM have been open for a few years now, I don't expect it to happen anytime soon

NTaylorMullen: Makes total sense!

@dbaeumer is there someone we can chat with to better understand VSCE expectations / details around yarn / npm integrations?

nagilson: Thanks for linking, I did reach out to Joao. Here are the relevant issues in VSCE's repo https://github.com/microsoft/vscode-vsce/issues/308
https://github.com/microsoft/vscode-vsce/issues/580
https://github.com/microsoft/vscode-vsce/issues/300 
https://github.com/microsoft/vscode-vsce/issues/381

dbaeumer: @NTaylorMullen the person to talk to is @joaomoreno

joaomoreno: > My hope is someday VSCE will support these parts of npm

What specific features are you referring to?

nagilson: > What specific features?

Thanks for following up (offline as well a while back.) All of the above are relevant but here are the main 2: https://github.com/microsoft/vscode-vsce/issues/308 and https://github.com/microsoft/vscode-vsce/issues/580

Issue #159, opened 2021-01-15
Powershell Script Content and Sh File Contents Get Mixed Up Across Shared Machine Platforms


eliezerchavez:
dotnet-install.sh: 7: Syntax error: newline unexpected

[DotNetAcquisition1610671763422.txt](https://github.com/dotnet/vscode-dotnet-runtime/files/5817746/DotNetAcquisition1610671763422.txt)

brandonh-msft:
New error now:

```shell
**Extension Version:** 1.0.0
**Error Message:** .NET Acquisition Failed: Installation failed: At C:\Users\anonymous\.vscode-insiders\extensions\ms-dotnettools.vscode-dotnet-runtime-1.1.0\dist\install 
scripts\dotnet-install.ps1:25 char:3
+ if [ -t 1 ] && command -v tput > /dev/null; then
+   ~
Missing '(' after 'if' in if statement.
At C:\Users\anonymous\.vscode-insiders\extensions\ms-dotnettools.vscode-dotnet-runtime-1.1.0\dist\install 
scripts\dotnet-install.ps1:25 char:5
+ if [ -t 1 ] && command -v tput > /dev/null; then
+     ~
Missing type name after '['.
At C:\Users\anonymous\.vscode-insiders\extensions\ms-dotnettools.vscode-dotnet-runtime-1.1.0\dist\install 
scripts\dotnet-install.ps1:25 char:13
+ if [ -t 1 ] && command -v tput > /dev/null; then
+             ~~
The token '&&' is not a valid statement separator in this version.
At C:\Users\anonymous\.vscode-insiders\extensions\ms-dotnettools.vscode-dotnet-runtime-1.1.0\dist\install 
scripts\dotnet-install.ps1:28 char:7
+     if [ -n "$ncolors" ] && [ $ncolors -ge 8 ]; then
+       ~
Missing '(' after 'if' in if statement.
At C:\Users\anonymous\.vscode-insiders\extensions\ms-dotnettools.vscode-dotnet-runtime-1.1.0\dist\install 
scripts\dotnet-install.ps1:28 char:9
+     if [ -n "$ncolors" ] && [ $ncolors -ge 8 ]; then
+         ~
Missing type name after '['.
At C:\Users\anonymous\.vscode-insiders\extensions\ms-dotnettools.vscode-dotnet-runtime-1.1.0\dist\install 
scripts\dotnet-install.ps1:28 char:26
[...]

PR #625, opened 2022-10-12
Fix Install Script Mismatches + Prepare for New Release

nagilson:
Resolves: https://github.com/dotnet/vscode-dotnet-runtime/issues/159 

This bug has been a big customer ask for quite a while and impacting the bicep team, we had some help from them reproing it and they confirmed that this fixes the issue. It seems the cache used the same keys between script extensions. 

This also bumps a dependency security update and makes more infrastructure changes. The custom registry which is a pain is not used for things which won't be packed into the extension anymore. 

We plan on releasing this as the 1.6.1 version.


vscode-dotnet-runtime-library/src/Acquisition/InstallScriptAcquisitionWorker.ts
```diff
@@ -25,7 +25,7 @@ export class InstallScriptAcquisitionWorker implements IInstallScriptAcquisition
         const scriptFileEnding = os.platform() === 'win32' ? 'ps1' : 'sh';
         const scriptFileName = 'dotnet-install';
         this.scriptFilePath = path.join(__dirname, 'install scripts', `${scriptFileName}.${scriptFileEnding}`);
-        this.webWorker = new WebRequestWorker(extensionState, eventStream, this.scriptAcquisitionUrl + scriptFileEnding, scriptFileName);
+        this.webWorker = new WebRequestWorker(extensionState, eventStream, this.scriptAcquisitionUrl + scriptFileEnding, scriptFileName + scriptFileEnding);
```
NTaylorMullen, line 28: Ahh good catch! 👍

NTaylorMullen (approved):

vscode-dotnet-runtime-library/src/Acquisition/InstallScriptAcquisitionWorker.ts
nagilson: Thanks, do the changes look good from a releasing standpoint? I talked with Sarah about how to make a new release but wanted to check if you knew of anything.

PR #642, opened 2022-10-20
Move version to 1.6.0

nagilson:
1.6.1 was an internal preview version, the latest released version was 1.5.0, so changing it so the public versioning remains consistent for the release.


nagilson:
Aborting failure of NPM Yarn Sync since Yarn doesn't have package publishing details inside of it. I'll need to fix that at some point

PR #652, opened 2022-11-03
Enable CodeQL + Eliminate Publishing Code

nagilson:
For reasons we aren't allowed to publish publicly anymore and had to change how it worked. We also need to enable CodeQL, if we do not do these steps all of our pipelines will be out of compliance and probably start failing.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,8 +14,12 @@ pr:
     - '*'
 
 variables:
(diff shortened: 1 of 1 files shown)


NTaylorMullen (approved):

nagilson:
@NTaylorMullen  You saw it before I even requested it and messaged you 😳  I am impressed!

NTaylorMullen:
> @NTaylorMullen You saw it before I even requested it and messaged you 😳 I am impressed!

Always on the lookout 😉

PR #799, opened 2023-03-20
Migrate to Axios to fix caching bugs & Reduce development burden + Other QOL Fixes

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/416. The old cache did not ever time out and so we've replaced it with axios cache systems. 

A 'modernized' version of https://github.com/dotnet/vscode-dotnet-runtime/pull/432 
(AKA completed, polished, and integrated with the new infrastructure changes over the past year.)

Also:
- Adds ability to test only a specific subset of tests (sdk, lib, or runtime) via test lib, test sdk, etc (See documentation change)
- Removes unneeded json configuration due to a change in vscode extension handling
- Adds debug logging to better evaluate what's gone wrong in code that is hard to put breakpoints in
- Adds instructions on how to debug the extensions with breakpoints
- Improves documentation on how to manage both yarn and npm successfully
- Updates some out of date packages that needed updates


vscode-dotnet-sdk-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -233,19 +235,30 @@ suite('DotnetCoreAcquisitionExtension End to End', function() {
...
-    const context: IDotnetAcquireContext = { version: '5.0', requestingExtensionId: 'ms-dotnettools.sample-extension' };
+    const context: IDotnetAcquireContext = { version: currentSDKVersion, requestingExtensionId: 'ms-dotnettools.sample-extension' };
     const result = await vscode.commands.executeCommand<IDotnetAcquireResult>('dotnet-sdk.acquire', context);
     assert.exists(result);
     assert.exists(result!.dotnetPath);
     uninstallSDKExtension();
     assert.isFalse(fs.existsSync(result!.dotnetPath));
   }).timeout(100000);
+
+  test('Extension installs latest SDK version', async () => {
```
nagilson, line 254: Why is this new test necessary?

baronfel: I think I added it to have something I could debug easier, but that's just a guess. I think there might even be a bug in this test because it's hard-coding 6.0.201 when the 'currentSDKVersion' might float.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -2,79 +2,140 @@
...
+    protected async makeWebRequest(throwOnError: boolean, retries: number): Promise<string | undefined> {
+        Debugging.log(`makeWebRequest Invoked. Requested URL: ${this.url}`);
+        try
+        {
             this.eventStream.post(new WebRequestSent(this.url));
-            const response = await request.get(options);
-            this.cacheResults(response);
-            return response;
-        } catch (error) {
+            const response = await this.client.get(
```
nagilson, line 107: It would be best to add some sort of promise.race or timeout besides the axios timeout as it doesn't cover all invalid url edge cases if this internal function is ever used incorrectly, but I tried this and it wasn't very successful, so probably not investing more time into that right now

nagilson: Added a timeout wrapper.

nagilson:
I would like to see if the pr builds are consistent at this point.

nagilson:
It appears more consistent today, I think this is ready for review as well.. but itd be good to do another round of careful testing.

baronfel (approved):
The core change of the axios client, retry handler, and cache adapter looks good to me. The test changes make sense as well. :shipit:

vscode-dotnet-runtime-library/src/Utils/Debugging.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    public static log(message : string)
+    {
+        if(Debugging.debugOn)
+        {
+            console.log(message);
+        }
+        else
+        {
+            ; // do nothing to appease ansync?
```
nagilson, line 20:
```suggestion
            ; // do nothing to appease async?
```
Remove this else statement?

vscode-dotnet-runtime-library/src/Acquisition/ReleasesResult.ts
```diff
@@ -0,0 +1,49 @@
...
+//       "releases-index": [ --> Array of release channels
+//        {
+//           "channel-version": "X.X", --> Major.Minor version this channel represents
+//           "latest-runtime": "X.X.X", --> Most recently released full version of the runtime
+//           "latest-sdk": "X.X.X", --> Most recently released full version of the SDK
+//           ...
+//        },
+//        ...
+//    }
+export class ReleasesResult {
```
nagilson, line 18: What is this file doing here?

vscode-dotnet-runtime-library/src/test/unit/WebRequestWorker.test.ts
```diff
@@ -18,82 +18,99 @@ import {
...
+        assert.exists(uncachedResult);
+        assert.deepEqual(uncachedResult, cachedResult);
+
         const requestCount = webWorker.getRequestCount();
-        assert.isBelow(requestCount, requests.length);
-    });
+        assert.isAtMost(requestCount, 1);
+    }).timeout(maxTimeoutTime);
 
-    test('Web Requests are retried', async () => {
```
nagilson, line 90: Why did we remove the requests are retried test again?

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -2,57 +2,164 @@
...
-                private readonly extensionStateKey: string) {}
+    constructor(
+        private readonly extensionState: IExtensionState,
+        private readonly eventStream: IEventStream,
+        private readonly url: string,
+        private readonly websiteTimeoutMs: number, // Match the default timeout time of 10 minutes.
+        private cacheTimeToLive = -1
+        )
+        {
+            this.cacheTimeToLive = this.cacheTimeToLive === -1 ? this.websiteTimeoutMs * 5 : this.cacheTimeToLive; // make things live 5x (arbitrary) as long as it takes to maximally acquire them
```
nagilson, line 50: @baronfel Is this cacheTimeToLive appropriate for the default?

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -2,57 +2,164 @@
...
+            const uncachedAxiosClient = Axios.create({});
```
nagilson, line 51: Why are we using create without setting the timeout and such? It looks like we set it below, so I guess it's not a big deal

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -2,57 +2,164 @@
...
+    return buildStorage({
+        // tslint:disable-next-line
+        set(key: string, value: any) {
+            extensionStorage.update(`${cachePrefix}:${key}`, value);
+        },
+        remove(key: string) {
+            extensionStorage.update(`${cachePrefix}:${key}`, undefined);
+        },
+        find(key: string) {
+            return extensionStorage.get(`${cachePrefix}:${key}`) as StorageValue;
```
nagilson, line 29: does extensionstroage return undefined if not found? Thats what the axios cache interceptor expects

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -2,57 +2,164 @@
...
+     * https://github.com/axios/axios/issues/647 (e.g. bad URL/site down).
+     */
+    private async axiosGet(url : string, options = {})
+    {
+        if(url === '' || !url)
+        {
+            throw new Error(`Request to the url ${this.url} failed, as the URL is invalid.`);
         }
-        return this.cachedData;
+        const abort = Axios.CancelToken.source()
```
nagilson, line 89: This doesnt seem to actually set a timeout like the function suggests?

nagilson: its private and all of the callers do, but according to the comment above, axios timeout does not work, and I agree with those comments. Dont know why axios has not done anything about it

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
nagilson: yes according to vscode docs

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
nagilson: I think I did not want to make it specific to one url, I could go either way on this. Will leave it for now

vscode-dotnet-runtime-library/src/test/unit/WebRequestWorker.test.ts
nagilson: Ah yes: https://github.com/dotnet/vscode-dotnet-runtime/pull/799/commits/ed49b7d52dc571b568decddf3944669ce219cd5c

it was a faulty test. But it would be ideal to make a new retry test that isnt faulty. A bit hard to do with these libraries, though.

vscode-dotnet-runtime-library/src/Acquisition/ReleasesResult.ts
nagilson: Ah I think I included it during merge accidentally a long while back but it got merged into one file because it was needed in several places

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
nagilson: Oh yes its the connection based vs response based timeout fiasco, yes I should ideally add a timeout, will cc offline to see if this is a release blocker

nagilson: I added a custom timeout which seems to work.

vscode-dotnet-sdk-extension/src/extension.ts
```diff
@@ -99,6 +99,7 @@ export function activate(context: vscode.ExtensionContext, extensionContext?: IE
     };
 
     const timeoutValue = extensionConfiguration.get<number>(configKeys.installTimeoutValue);
+    const finalTimeoutValue = timeoutValue === undefined ? defaultTimeoutValue : timeoutValue;
```
baronfel, line 102: This value flows through a different places and at the end of it's journey it's multiplied by 1000 (presumably to generate a number of milliseconds). Could you change the name of these fields to be explicit that this time value in intended to be seconds here (and where the value flows to)?

baronfel (approved):
Also LGTM after the most recent round of changes :shipit: 

 I left a note about one place that I was confused, mostly about naming.

vscode-dotnet-sdk-extension/src/extension.ts
nagilson: That's a good point, thank you. resolvedTimeoutSeconds is the new name.

LLotme (commented):
gh pr checkout 799

Issue #763, opened 2023-02-21
Allow SDK/Runtime extensions to manage machine-global installations


baronfel:
Currently, the [SDK](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-sdk-extension/src/extension.ts#L114-L132) and [Runtime](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-extension/src/extension.ts#L111-L132) `acquire` commands follow the [following contract](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/IDotnetAcquireContext.ts#L8-L12):

```ts
export interface IDotnetAcquireContext {
    version: string;
    requestingExtensionId?: string;
    errorConfiguration?: AcquireErrorConfiguration;
}

export enum AcquireErrorConfiguration {
    DisplayAllErrorPopups = 0,
    DisableErrorPopups = 1,
}

export interface IDotnetAcquireResult {
    dotnetPath: string;
}

export interface AcquireCommand =IDotnetAcquireContext => Promise<IDotnetAcquireResult> 
```

In addition, the `version` string has a specific meaning - it is intended to map to the `channel` of the underlying entry in the [releases index](https://github.com/dotnet/core/blob/main/releases-index.json). This allows for managing user-local installations of the SDK and Runtime, but there's no reason why this extension shouldn't also be able manage machine-global installations if requested.

We should make the following updates to the extension:

* We should allow the `acquire` command for SDKs and Runtimes to specify versions more precisely, using syntax that is familiar to both existing extension users, and users of the setup-dotnet github actions:
  * Versions allowed:
    * specify major version only - '6'
    * specify major.minor - '6.0'
    * specify a fully-qualified version - '6.0.403'
    * specify a feature-band - '6.0.40x' or (more correctly) '6.0.4xx'
  * When a non-specific version is used, the releases-index should be used to look up the corresponding `releases.json` for that 'channel' (major/minor), and that releases.json should be used to find the latest version that matches the given spec
    * 6/6.0 would resolve to the most recent feature band, and most recent patch within that feature band
    * 6.0.40x/6.0.4xx would resolve to the most recent patch in the specified feature band

* We should allow the `acquire` command to manage the installation of machine-global SDKs, ideally via the installers for the particular systems. 
  * This would require the addition of a new flag
  * This is enabled by the `releases.json` for each channel. 
[...]

PR #1046, opened 2023-07-14
Enable Global SDK Installs for Win, Mac, & Ubuntu

nagilson:
! This PR Description is a copy of the document made in `\Documentation\global-installs\early-global-code-changes-overview.md`)

# .NET Global Acquisition Engineering Spec

Resolves: https://github.com/dotnet/vscode-dotnet-runtime/issues/763.

Described below is an overview of the changes made to enable supporting global .NET SDK installs via our VS Code Extensions. This document's original function was to help during PR Review, as this functionality was added in 1-go. (This is terrible and I am sorry: but we had initially struggled to get reviewers during development and were trying to move swiftly.)

# Extension Plan

We have added functionality to the .NET SDK Extension to install the SDK Globally. That extension will also be able to install the .NET Runtime globally for use by the .NET SDK. 

> :apple: The .NET Education Bundle APIs will be unaffected as we have just added a global option, however the extension will be re-branded to remove the remarks of 'do NOT use,' as well as the 'education bundle' marketing. 

The API will change to installing the SDK globally by default in the future. (Because we do not want people to rely on the local SDK install feature in production, as it was intended only for the education bundle.) We are waiting for the education bundle to be prepared for that breaking change.

In the initial set of changes, the branding is not included. The changes to extension documentation, names, etc, will go in a separate PR.

# Overview of How Global Installs are Done

> :exclamation: For Windows and Mac, we can leverage the .NET Installer that we ship. We can determine the correct installer and simply download and then execute it.

For Windows, if we are already running VS Code with admin, we can simply download and run the installer in quiet mode and the user will not see anything.

image

In any other case on windows, and always on mac, we can make the .NET Installer appear above VS Code. That will handle any elevation/UAC prompts for us. It also will handle conflicting SDKs that could be on the machine, setup issues with currently running versions of dotnet, etc, and provides a consistent already-built user experience. We have greatly simplified our job by just downloading the correct installer file from your machine using data available from our releases APIs, and then executing said installer.

> :exclamation: For Linux we essentially had to write our own installer. We have a separate linux document, `[here](linux-global-install-design.md)` which I would suggest checking out.

The TLDR is the following: It uses the distro package manager to install the dotnet package on the system using commands such as `apt-get`. It will use the dotnet packages provided by the distro if it's version includes such packages, otherwise it will attempt to use the Microsoft feed packages.

Our initial push includes Ubuntu. We intend add RHEL and other supported linux versions within VS Code before launch. We also have a policy in place for the community to add support to other distros described in the linux document. :star: In the intial set of changes, only Ubuntu is supported.

A threat model must be developed. It will not be part of the initial PR.

# Changes Made

For those reviewing the code changes, it may be best to look at the isolated files and their tests so the review can be broken up into chunks. The changes should also be investigated from an end-to-end perspective. However, the changes are minimal and copy existing patterns, with a global flag that branches to its own logic. Here are the working pieces that were added/changed:

### Review Suggestions

The code should be under the most scrutiny. We will have PMs review the error messages. For decisions such as, 'WSL is not supported,' or, 'We support XYZ Distros,' or 'We do/don't allow this install,' every decision has been vetted by 3-5+ PMs, so while questioning these decisions may be of use, it is not something a reviewer needs to focus on.

## Version Resolver

The version resolver now has functionality to extract the feature band, patch version, major, minor, and more from modern .NET Core Versions. We don't expect this will be used to install older .NET versions or plan to support that, so this is OK.

## Global Installer Resolver

As defined in the spec here https://github.com/dotnet/vscode-dotnet-runtime/issues/763, users can request the SDK using different version formats when making a global request.

This class uses the new VersionResolver functionality to determine what version the user should install (aka the newest version that matches what they requested.) It should find the correct installer file to download for windows and mac and save a link to that file for others to access.

## Linux Version Resolver

The Linux Version Resolver has a few purposes. One, it determines what distro and version you are running on, and creates the correct distro class for you. Currently there is no special distro logic in place since only Ubuntu is supported, however, we anticipate the factory function this class has will be expanded to support more distros that have unique requirements to do an installation.

This class is also responsible for calling the correct APIs to validate and ensure that the correct action is taken: aka, if we can update instead of installing, if we do not need to install because an install already exists, or if we cannot install, this class handles all of that logic. 

## Linux Installer

The Linux Global Installer exists to provide high level visiblity into the install so code calling the installer can make decisions it needs to make about the install, and mostly so we can match the API that is used by the windows and mac installer.

## GenericDistroProvider

All of the distro specific logic has been extracted out of the linux version resolver and put into this class. It is essentially a 'front' into the distro.json file that contains all of the commands that need to be executed for a particular distro to do an install.

The Generic Distro Provider was written under Ubuntu, but it is also 'generic' in the sense that for most commands, it simply runs the command in the json file for the matching distro and version, so it is relatively platform-agnostic. Ubuntu does not require much special logic to capture the output of the commands or run the commands, and so it is extendable to other distros.

> :exclamation: This interface currently supports more than what's used and exposed by our commands. This code has been tested. For example, we add the ability to uninstall the SDK, however, we have yet to add a global uninstall command. We do not have strong demand for such an API yet. But as we want to accept community contributions, we wanted to future-proof this API so everything that we could possibly need has been implemented.

## Windows & Mac Installer

The Windows and Mac installer provides ability to download the install file correctly and execute it using the correct flags. It will also have sign-checking in the future. 

## File Utilities

This class allows us to check if we are elevated and will host any file related operations needed for managing installs.

## Command Executor

This library runs a command you give it. For linux, it will handle the sudo elevation prompts that are needed to gain elevation.

For sudo elevation we rely on the built-in `@vscode/sudo` library which will show a prompt above the user like that is not super encouraging. We plan to make changes in their library to improve the UX.

MicrosoftTeams-image (1)

One of the key purposes of this class is to make the global installs mockable and testable, as we don't want to leave global install artifacts on dev-box or pipeline machines. It also enables us to fake responses from the OS when running commands or installs.


nagilson:
Closing this in favor of: https://github.com/dotnet/vscode-dotnet-runtime/pull/1081
So we dont have to force push the cleaner commit history :)

Issue #885, opened 2023-05-16
Improve Issue Experience for .NET VS Code Extensions


nagilson:
# Context
Currently, we have an `ErrorHandler` that handles all Errors thrown in the VS Code .NET SDK and VS Code .NET Runtime Extension. 

This `ErrorHandler` has an `IssueReporter` that lets customers create a GH issue on our repo by opening the "create an issue url" when they cancel the error window. They often don't understand what will happen when they click to report the error, and they often do not fill in the needed details. So, we get hundreds of non-issues that are unresolvable like this one: https://github.com/dotnet/vscode-dotnet-runtime/issues/871

We would like to not have to close hundreds of blank issues. We cannot remove the privacy statement or auto-fill the issue with the log file seen in the above issue due to privacy concerns. Another problem is that the logs that this message recommends can end up empty, and that should be fixed. I have not investigated why this is the case.

# What to do? 

There are several ways to remedy this.
Here's the first few things:
1) We have no issue template. We should remedy this by adding an issue template to the repo.
2) We should fix the log file issue so they aren't empty. https://github.com/dotnet/vscode-dotnet-runtime/issues/731 Is an example of someone attaching the log file, yet the log file is empty. This is the case for many issues in the repo. We don't know why this happens yet as it hasn't been investigated.
3) We should redirect users to the common issues page and modify that page to link to the create an issue page if their issue isn't found.
https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md is the page.

I would like to have @baronfel chime in on this before work is started. This is 'low' priority.

cc @marcpopMSFT

I hope this is a good first-issue that is also high-impact. The code for this can be found here: https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/Utils/IssueReporter.ts

nagilson:
A good proposal should not have A, B, C, D, and instead have the one option. I will arrange something @baronfel 
 @JL03-Yue so we can discuss this and come to a final decision.

baronfel:
Definitely agree on fixing 1. and 2.

For the remainder, We probably have two modes of operation for this extension:
A) we are completely in control of runtime/SDK acquisition, 
B) we are a _client_ of some other extension that's just using us as a means to an end

Our error reporting/user interaction experience should differ based on client type (which could be signalled via the API's of our methods).

[...]

PR #1122, opened 2023-07-27
fix logging issue and add a logging unit test

JL03-Yue:
Addressing the error issue of #885 

Fixes the empty log issue and add `LoggingObserver.text.ts` to incorporate it into automated test.

### DIFF
--- a/vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
@@ -31,6 +31,7 @@ export class LoggingObserver implements ILoggingObserver {
     }
 
     private writeLine(line: string) {
-        this.log = this.log.concat(line);
+        //this.log = this.log.push(line);
+        this.log.push(line);
     }
 }
--- a/vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
@@ -31,7 +31,6 @@ export class LoggingObserver implements ILoggingObserver {
     }
 
     private writeLine(line: string) {
-        //this.log = this.log.push(line);
         this.log.push(line);
     }
 }
--- a/vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
@@ -7,7 +7,7 @@ import { IEvent } from './IEvent';
 import { ILoggingObserver } from './ILoggingObserver';
 
 export class LoggingObserver implements ILoggingObserver {
-    private readonly log: string[] = [];
+    private log: string[] = [];
 
     constructor(private readonly logFilePath: string) {}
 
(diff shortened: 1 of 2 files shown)


vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
```diff
@@ -31,6 +31,6 @@ export class LoggingObserver implements ILoggingObserver {
     }
 
     private writeLine(line: string) {
-        this.log.concat(line);
+        this.log = this.log.concat(line);
```
baronfel, line 34: consider using `this.log.push(line)` instead - `concat` is a copying call, so this will generate a lot of additional garbage for the GC to manage. `push` modifies the underlying array and handles updating the size of the array if necessary, so copying is kept to a minimum.

nagilson: Nice catch, thank you :)

vscode-dotnet-runtime-library/src/test/unit/LoggingObserver.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    test('Log file is writing output', async () => {
+        // Create an empty folder
+        if (!fs.existsSync(tempPath)) {
+            fs.mkdirSync(tempPath);
+        }
+        // Create a logging observer
+        const loggingObserver = new LoggingObserver(path.join(tempPath, 'logTest.txt'));
+
+        // Create a fake event and call the post/dispose function
```
nagilson, line 25: nit: the post and dispose are technically called in the next set of actions. I would change this comment and the following one

vscode-dotnet-runtime-library/src/test/unit/LoggingObserver.test.ts
```diff
@@ -0,0 +1,38 @@
...
+        const fakeEvent = new DotnetUninstallAllStarted();
+
+        // Check if the log file content is same as expected content
+        loggingObserver.post(fakeEvent);
+        loggingObserver.dispose();
+
+        fs.readdirSync(tempPath).forEach(file => {
+            const logContent = fs.readFileSync(path.join(tempPath, file)).toString();
+            assert.include(logContent, fakeEvent.eventName, 'Does not include');
```
nagilson, line 34:
```suggestion
            assert.include(logContent, fakeEvent.eventName, 'The log file does not contain the expected content that should be written to it?');
```

nit: It is very handy to include error messages in the assert that say why they might fail! Nice thinking on that :) Though I would try to use a more descriptive message : imagine seeing this 1 year from now and it fails in a pipeline with 'Does not include.' It would be much more helpful to have more context and what couldve gone wrong

vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
nagilson, line 34: This is pretty subtle! Thank you for having such a concise fix and figuring it out! 😄

nagilson (approved):
After the following suggestions are added, it looks good to me :) I think we should even consider making a release with this so we can get better logs for current customers. We could work together on that! Thanks for this fix, it will help resolve customer issues a lot!

PR #1212, opened 2023-08-09
Verify that PowerShell Install Scripts can Execute

nagilson:
Add an error to tell users how to install PowerShell if it's not on the PATH. Also check execution policy to see if the elevation level needed to run dotnet install is available or not.

Also changes the timeout value for those who have slower internet -> assuming 8 mbps/sec for downloading .NET is roughly 4.5 minutes or 270 seconds + internet SP ramp up time.

Changes messaging for timeout to include instructions on how to change timeout value and also makes button more likely to be clicked on, or that's the hope.

Didn't add tests as it's kind of a pain to test. The MockCommandExecutor in my Global PR solves this but we'd rather not add global code right now, IMO.

### DIFF
--- a/sample/yarn.lock
+++ b/sample/yarn.lock
@@ -3,771 +3,771 @@
 
 
 "@babel/code-frame@^7.0.0":
-  "integrity" "sha1-BuKrGb21NThVWaq7W6WXKUgoAPg="
-  "resolved" "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@babel/code-frame/-/code-frame-7.0.0.tgz"
-  "version" "7.0.0"
+  version "7.0.0"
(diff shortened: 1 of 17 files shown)


baronfel (approved):
The code changes look good to me, the messages are certainly more informative :+1: 

Are the lockfile changes required as part of this? As far as I can tell it looks like just metadata changes to the existing lockfile format, but it does sort of muddy the waters.

nagilson:
> The code changes look good to me, the messages are certainly more informative 👍
> 
> Are the lockfile changes required as part of this? As far as I can tell it looks like just metadata changes to the existing lockfile format, but it does sort of muddy the waters.

Thanks :) The lock file changes aren't necessary, but I actually think I should upgrade one package before hand which I am doing now. Any time you execute ./build.cmd it will change it automatically to this, so while it's not ideal it would come in eventually no matter what PR.

nagilson:
I can see it causing nasty conflicts in the future, though regrettably I do actually think it's required :/ Let me know if you want more info

baronfel (approved):
The new messages should be very useful for end users!

nagilson:
Thanks for taking the time to review

Issue #1305, opened 2023-08-22
Investigate if `chmod` of the scripts 'takes'


baronfel:
Related to #1304, we might need to check if `chmod` took after setting the executable bit on the script. Do this one after the race condition in case this is merely a symptom of that issue.

Issue #1304, opened 2023-08-22
Investigate and solve repeated writes of the install-scripts


baronfel:
[This call](https://github.com/dotnet/vscode-dotnet-runtime/blob/3250aa7d4706c733a5b8e7c286ca3d569ccdb646/vscode-dotnet-runtime-library/src/Acquisition/InstallScriptAcquisitionWorker.ts#L56-L63) writes the script file on every acquisition. In parallel-call scenarios (multiple extensions initializing) it's possible to race here and end up overwriting the script file with incorrect permissions while another invocation is attempting to execute the script.

Issue #1303, opened 2023-08-22
Execute the install scripts with verbose logging for better diagnostics


baronfel:
The scripts emit information about their output on verbose logging channels - we should opt into that so that we have better diagnostics information. This would mean adding `-Verbose` in powershell and `--verbose` on the sh script.

PR #1334, opened 2023-08-29
Version 1.7.3 - Introduce File Locking for Install Script

nagilson:
Introduces https://www.npmjs.com/package/proper-lockfile to eliminate possible race conditions in install script file writes.

Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1304
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1303/
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1309
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1305

Btw, here is evidence that the 'chmod' takes. Before and after chmod exec on mac.
image


vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -151,7 +151,7 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
             timeoutValue: this.timeoutValue,
             installRuntime,
         } as IDotnetInstallationContext;
-        this.context.eventStream.post(new DotnetAcquisitionStarted(version));
+        this.context.eventStream.post(new DotnetAcquisitionStarted(version, this.context.acquisitionContext?.requestingExtensionId));
```
baronfel, line 154: 💯

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
```diff
@@ -14,30 +14,38 @@ export class DotnetAcquisitionStarted extends IEvent {
...
     public readonly eventName = 'DotnetRuntimeAcquisitionStarted';
     public readonly type = EventType.DotnetRuntimeAcquisitionStart;
 
+    constructor(public readonly requestingExtensionId = '') {
+        super();
+    }
+
     public getProperties() {
-        return undefined;
+        return {extensionId : this.requestingExtensionId};
```
baronfel, line 35: same - this needs to be hashed since extension Ids can be freetext.

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
```diff
@@ -260,6 +268,35 @@ export class DotnetFallbackInstallScriptUsed extends DotnetAcquisitionMessage {
     public readonly eventName = 'DotnetFallbackInstallScriptUsed';
 }
 
+export abstract class DotnetFileEvent extends DotnetAcquisitionMessage
+{
+    constructor(public readonly eventMessage: string, public readonly time: string) { super(); }
```
baronfel, line 273: one downside to taking the formatted messages as parameters is that you lose what the inputs to the formatted strings were. filepaths, etc go into these messages and so we have a PII concern. Can you look at what it would take to pass in parameters to these constructors and scrub the sensitive fields in `getProperties()`?

vscode-dotnet-runtime-library/src/Acquisition/InstallScriptAcquisitionWorker.ts
```diff
@@ -40,7 +46,7 @@ export class InstallScriptAcquisitionWorker implements IInstallScriptAcquisition
             }
 
             Debugging.log('Writing the dotnet install script into a file.');
-            this.writeScriptAsFile(script, this.scriptFilePath);
+            await this.writeScriptAsFile(script, this.scriptFilePath);
```
baronfel, line 49: So this code change successfully ensures serialized access to the writing of the file, but we still attempt to write the file on every acquisition request, right? It would be best if we could avoid doing that work unless necessary (script file content changed, file checksum on disk is incorrect, etc).

baronfel (changes requested):
I think we have to address the telemetry hashing on the event payload data before merge, but the logic of the lock looks good. I'd like to not do the write at all if at all possible.

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
nagilson: That's a really good point. Definitely. On it!

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -98,23 +98,23 @@ export function activate(context: vscode.ExtensionContext, extensionContext?: IE
...
         installDirectoryProvider: new RuntimeInstallationDirectoryProvider(context.globalStoragePath),
     });
     const existingPathResolver = new ExistingPathResolver();
-    const versionResolver = new VersionResolver(context.globalState, eventStream, finalTimeoutValue);
+    const versionResolver = new VersionResolver(context.globalState, eventStream, resolvedTimeoutSeconds);
 
     const dotnetAcquireRegistration = vscode.commands.registerCommand(`${commandPrefix}.${commandKeys.acquire}`, async (commandContext: IDotnetAcquireContext) => {
         const dotnetPath = await callWithErrorHandling<Promise<IDotnetAcquireResult>>(async () => {
-            eventStream.post(new DotnetRuntimeAcquisitionStarted());
+            eventStream.post(new DotnetRuntimeAcquisitionStarted(commandContext.requestingExtensionId));
```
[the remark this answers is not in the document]
nagilson: Thank you a bunch for catching this. Thats really important. 

We don't need to hash it here since I made it get hashed inside of the EventStreamEvents get properties. This means the log can contain the true info but the telemetry we collect will not

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
nagilson: Same here, covered in Event Stream Events now

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
nagilson: Done. Also scrubbed the dotnet install script output of paths since that has 'em too.

vscode-dotnet-runtime-library/src/Acquisition/InstallScriptAcquisitionWorker.ts
nagilson: Yep. Fixed 👌

nagilson:
@baronfel It should be hashed now. Let me know if there's any other concerns. Thanks for taking a look 🔥

vscode-dotnet-runtime-library/src/EventStream/TelemetryUtilities.ts
```diff
@@ -0,0 +1,17 @@
+ /* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+ import * as crypto from 'crypto';
+import { TextEncoder } from 'util';
+
+export class TelemetryUtilities
+{
+    public static HashData(dataToHash: string) : string
```
baronfel, line 12: nit: I notice in a few places that you have to ternary-check the data passed to this method in case it is null - could you instead change the signature here to `HashData(datatoHash: string?) : string` and have this method check for `null` and in that case return `''`? Might make your callsites less painful.

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -49,10 +50,10 @@ You will need to restart VS Code after these changes. If PowerShell is still not
                         async (error, stdout, stderr) => {
                     if (error) {
                         if (stdout) {
-                            this.eventStream.post(new DotnetAcquisitionScriptOuput(installContext.version, stdout));
+                            this.eventStream.post(new DotnetAcquisitionScriptOuput(installContext.version, TelemetryUtilities.HashAllPaths(stdout)));
```
baronfel, line 53: I think it'll take a more detailed approach to properly de-PII the stdout/stderr streams from the install-scripts. IMO that should be unrelated to this PR, and this PR should mostly be focused on making sure that we safely treat the extension Id.

baronfel (commented):
Approved, but I think we could handle anonymizing the install-scripts output in a separate unit of work - right now the API is looking at the entire giant string of the stdout/stderr streams, and I think we'll need to move to a more incremental checking approach that uses the events raised by the [ChildProcess](https://nodejs.org/api/child_process.html#class-childprocess) that's returned by spawning the scripts.

baronfel (approved):

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
nagilson: That makes sense. I added this as stopgap to improve it with what we can do for now. This isn't great if there's a space in the path. We could take a dynamic programming approach to fix that. But that's for another time.

vscode-dotnet-runtime-library/src/EventStream/TelemetryUtilities.ts
nagilson: Good point, thank you

PR #1369, opened 2023-09-12
Explicitly call the powershell.exe when running the install script

MiYanni:
Resolves all the [powershell not found](https://github.com/dotnet/vscode-dotnet-runtime/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Acause%3Apowershell-not-found) tagged issues.

## Summary
Instead of relying on the console to resolve the `PATH` variable, which can be modified by a user or other applications, we explicitly call `powershell.exe` from the file location that exists on the system. `%SystemRoot%` will always resolve to the Windows installation location, for example, `C:\Windows`. This removes the dependency on the PATH containing this directory.

**Additionally:**
- Removed an extraneous space in the command call
- Added [`-NonInteractive`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe?view=powershell-5.1#-noninteractive) and [`-NoLogo`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe?view=powershell-5.1#-nologo) since we're running an install script

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
@@ -39,7 +39,7 @@ You will need to restart VS Code after these changes. If PowerShell is still not
         const installCommand = await this.getInstallCommand(installContext.version, installContext.installDir, installContext.installRuntime);
         return new Promise<void>((resolve, reject) => {
             try {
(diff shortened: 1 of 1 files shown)


nagilson:
Thank you @MiYanni for helping an trying this out, I think it's a great idea to try this path 😄 I was also working on changes last week related to this and merged out ideas together. Now it will try powershell, then your path idea, then pwsh and others after. We can use the data we have to see if people get past the path and go to pwsh, since this path is now the 2nd thing we try

MiYanni:
Closing in favor of: https://github.com/dotnet/vscode-dotnet-runtime/pull/1373

Issue #1246, opened 2023-08-14
Apostrophes in user ID Break download


poneill3:
I have a user ID set by my corporation that is my name.  My name contains an apostrophe.  So every time I start the vscode, I see 

.NET Acquisition Failed: Installation failed: & : The term 'c:\Users\PaulO' is not recognized as the name of a cmdlet, function, script file, or operable program. 
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:138
+ ... l -bor [Net.SecurityProtocolType]::Tls12 ; & 'c:\Users\PaulO'Neill\.v ...

So it's just pulling in the first bit.  

This actually broke my entire environment today, so now I've got to troubleshoot an alternative.

baronfel:
Thanks for the report - this is definitely something we should correct

nagilson:
This is a high priority issue. We have someone working on a fix now. Thanks for reporting it.

MiYanni:
There is already code that escape the single quotes in the path to the install script. I'm a bit perplexed why it isn't working in this situation. Code is here:
https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts#L103

nagilson:
That's a good point, @JL03-Yue saw that as well. I think it's the way it's escaped? Like, it sees there is a ', so it ends the string prematurely. It's probably double ''.

MiYanni:
> That's a good point, @JL03-Yue saw that as well. I think it's the way it's escaped? Like, it sees there is a ', so it ends the string prematurely. It's probably double ''.

Almost certainly. And yes, you use double single quotes to escape a single quote within a PowerShell string.

JL03-Yue:
Thank you for the input! This is a great point

nagilson:
@poneill3 This should be fixed in 1.8.0, may you please see if this still occurs for you? If you uninstall and reinstall online, vscode should pick up the updates

PR #1372, opened 2023-09-12
escape the single quotes in path when installing .NET

JL03-Yue:
Per issue #1246, this PR escapes the single quotes in user's folder path when installing .NET


vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -101,8 +101,7 @@ You will need to restart VS Code after these changes. If PowerShell is still not
     private escapeFilePath(path: string): string {
         if (os.platform() === 'win32') {
             // Need to escape apostrophes with two apostrophes
-            const dotnetInstallDirEscaped = path.replace(/'/g, `''`);
-
+            const dotnetInstallDirEscaped = path.replace(/'/g, `\'`);
```
MiYanni, line 104: This *might* work but it depends on how it is called. I'd be interested to see a unit test that resolves the command call (with path to the install script) and checks to see it works properly. Not sure what unit testing is available in this repo.

JL03-Yue: Thanks for the comment! A unit test is added with the changes we discussed

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -39,7 +39,7 @@ You will need to restart VS Code after these changes. If PowerShell is still not
         const installCommand = await this.getInstallCommand(installContext.version, installContext.installDir, installContext.installRuntime);
         return new Promise<void>((resolve, reject) => {
             try {
-                const windowsFullCommand = `powershell.exe -NoProfile -ExecutionPolicy unrestricted -Command "& { [Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12 ; & ${installCommand} }`;
+                const windowsFullCommand = `powershell.exe -NoProfile -ExecutionPolicy unrestricted -Command "& { [Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12 ; & ${installCommand} }"`;
```
MiYanni, line 42: I'm not sure this will actually solve the issue. But the double quotes should definitely be there.

MiYanni (commented):
I think we'd need to check if executing the command with a single quote in the path works in this scenario. I have a hunch that we're missing something in how this command is executed.

vscode-dotnet-runtime-library/src/test/unit/DotnetCoreAcquisitionWorker.test.ts
```diff
@@ -213,4 +213,11 @@ suite('DotnetCoreAcquisitionWorker Unit Tests', function() {
         const events = eventStream.events.filter(event => event instanceof DotnetAcquisitionStarted);
         assert.equal(events.length, 1);
     });
+
+    test('Get Expected Path', async () => {
+        const [acquisitionWorker, eventStream, context] = getTestAcquisitionWorker(true);
```
nagilson, line 220: I like the use of the `testAcquisitionWorker`, I think that will indeed use the actual install script writer to write the install script out. But I think this will not test the issue the user was facing by having ' in the path.

You could make a mock IInstallationDirectoryProvider that writes to a path with ' such as the users path in the issue. But the install script acquisition worker does not write to the directory provided by the installationdirectoryprovider, so you'd have to change that as well. 

Instead, I would recommend looking into `InstallScriptAcquisitionWorker` and making a mock of it that lets you change the `scriptFilePath,` and then change that to include the apostrophe in the install path to test it ;)

JL03-Yue:
./azp run

nagilson (approved):
LGTM, thanks for making these changes :) 

I will check in with the user since we could not get a repro with the username ourselves to see as well

Issue #1338, opened 2023-08-28
Architecture mismatch between VSCode and Install-Scripts for downloaded Runtime/SDK


4ColorDesign:
# Describe the bug
  Failed to find dotnet info from path, falling back to acquire runtime via ms-dotnettools.vscode-dotnet-runtime
No compatible .NET runtime found. Minimum required version is 7.0.
[Error - 4:13:01 PM] Microsoft.CodeAnalysis.LanguageServer client: couldn't create connection to server.
Error: Could not resolve the dotnet path!
    at t.DotnetRuntimeExtensionResolver.<anonymous> (c:\Users\atulm\.vscode\extensions\ms-dotnettools.csharp-2.0.413-win32-x64\dist\extension.js:2:1382303)
    at Generator.next (<anonymous>)
    at s (c:\Users\atulm\.vscode\extensions\ms-dotnettools.csharp-2.0.413-win32-x64\dist\extension.js:2:1380683)

  
  
  ## **Please attach log file(s) here if you have any:**
  >❗ Privacy Alert! The contents in our log files may contain personal data. Prior to posting to GitHub, please remove any personal data which should not be publicly viewable. https://privacy.microsoft.com/en-US/privacystatement
  
  
  
  If you came here from VS Code, the log file location should be copied to the clipboard.
  
  ## Other Info:
  Errors the extension may have encountered have likely been added to your terminal or VS Code windows. Please paste it/them here and describe what happened and the problem you are facing.
  
  **Reproduction Steps**
  
  
  **Expected behavior**
  
  
  **Exceptions & Errors (if any)**
  
  
  **Further technical details**
  - Your VSCode Version
  
  - The version of the extension that's on your machine
  
  - Your operating system details

ReyZRX:
Howdy. I have a similar err on which I spent the day futzing with after deciding to try C# extension on VS Code. Started w/Net SDK 7 and uninstalled that and C#, C# dev kit and problem still happens. Will now go back to C# 2010 Express as all I wanted to do was read the Beginning C# 2008 Objects book to get round to reading on OOP..

Been retired for 5 yrs now and now this...

Thanks for any assistance you/others may have on fixing this,

Rey

[CSharp_DevKit_err.txt](https://github.com/dotnet/vscode-dotnet-runtime/files/12501060/CSharp_DevKit_err.txt)

dibarbet:
@ReyZRX 

[...]

PR #1384, opened 2023-09-14
1.8.0 Use Node Architecture instead of Environment Variable Architectures

nagilson:
Use node.arch() to determine the architecture in hopes to improve x32 install experience per resolving https://github.com/dotnet/vscode-dotnet-runtime/issues/1338 

Note that installs are now managed by version-architecture, aka an `installKey`. This is a 'breaking' change of existing installs as they were managed using a key based on their fully specified version only, so existing installs will be removed when they are requested and then reinstalled upon success of an architecture-specific install. Please see the changes and commit comments for more details.

Also adds DotnetTotalSuccess event to keep track of the true success rate of installs.


vscode-dotnet-runtime-library/src/test/unit/DotnetCoreAcquisitionWorker.test.ts
```diff
@@ -180,6 +205,47 @@ suite('DotnetCoreAcquisitionWorker Unit Tests', function () {
...
+        const [sdkWorker, sdkEvents, sdkContext] = getTestAcquisitionWorker(false, null);
+        await AssertInstallSDK(sdkWorker, sdkContext, sdkEvents, sdkV5);
+        await AssertInstallSDK(sdkWorker, sdkContext, sdkEvents, sdkV6);
+
+        // Install 5.0 runtime with an architecture. Share the same event stream and context.
+        runtimeWorker.installingArchitecture = os.arch();
+        await AssertInstallRuntime(runtimeWorker, context, events, runtimeV5);
+
+        // 5.0 legacy runtime should be replaced, but 6.0 runtime should remain, and all SDK items should remain.
+        // todo: make test so install keys use fll version via extension e2e
```
nagilson, line 230:
```suggestion

```

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
```diff
@@ -15,12 +15,14 @@ export class DotnetAcquisitionStarted extends IEvent {
...
 
-    constructor(public readonly installKey: string, public readonly requestingExtensionId = '') {
+    constructor(public readonly installKey: string, public readonly startingVersion: string, public readonly requestingExtensionId = '') {
         super();
     }
 
     public getProperties() {
-        return {AcquisitionStartInstallKey : this.installKey, extensionId : TelemetryUtilities.HashData(this.requestingExtensionId)};
+        return {AcquisitionInstallKey : this.installKey,
+                AcquisitionStartVersion : this.installKey,
```
baronfel, line 24: i think this was meant to be `startingVersion`

baronfel (commented):
Couple comments, but overall looks good. I'll mark 'comment' for now and we can take another look.

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -172,7 +176,58 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
...
+     *
+     * @param allInstalls all of the existing installs.
+     * @returns All existing installs made by the extension that dont include a - for the architecture.
+     */
+    private existingLegacyInstalls(allInstalls : string[]) : string[]
+    {
+        let legacyInstalls : string[] = [];
+        for(const install of allInstalls)
+        {
+            if(!install.includes('-'))
```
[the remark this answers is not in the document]
nagilson: Oh, that's a really good point. They definitely do. We can't use `/` because the install key becomes part of the path, similarly `:` is an invalid path character. So instead I think I'll use `~` which we don't use anywhere.

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
nagilson: Good catch, ty ;)

baronfel (approved):
Let's do that hockey!

PR #1394, opened 2023-09-19
Dont let proxy check fail if registry access is disabled

nagilson:
The proxy library fails if registry access is disabled, wrap it in a try catch.
Bump version to 1.7.4.

### DIFF
--- a/vscode-dotnet-runtime-extension/CHANGELOG.md
+++ b/vscode-dotnet-runtime-extension/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning].
 
 ## [Unreleased]
 
+## [1.7.4] - 2023-08-24
+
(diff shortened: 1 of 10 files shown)


vscode-dotnet-runtime-extension/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning].
 
 ## [Unreleased]
 
+## [1.7.4] - 2023-08-24
+
+Install using node.arch() instead of environment variable architecture from root terminal to prevent mismatching shell architecture with vs code architecture.
```
nagilson, line 12: The node.arch() part will be added in another pr.

baronfel (approved):
This LGTM - for the new event we should filter that out in the telemetry and not consider it a terminal error (for overall success calculations), right?

Issue #1413, opened 2023-09-26
Error with .NET acquisition on C# Extension update


SusanTWhite:
# Describe the bug
Unable to connect to server

## **Please attach log file(s) here if you have any:**
>❗ Privacy Alert! The contents in our log files may contain personal data. Prior to posting to GitHub, please remove any personal data which should not be publicly viewable. https://privacy.microsoft.com/en-US/privacystatement

[DotNetAcquisition-ms-dotnettools.vscode-dotnet-runtime-1695731444737.txt](https://github.com/dotnet/vscode-dotnet-runtime/files/12727578/DotNetAcquisition-ms-dotnettools.vscode-dotnet-runtime-1695731444737.txt)

If you came here from VS Code, the log file location should be copied to the clipboard. 

c:\Users\xxxxxxx\AppData\Roaming\Code\logs\20230926T083039\window1\exthost\ms-dotnettools.vscode-dotnet-runtime\DotNetAcquisition-ms-dotnettools.vscode-dotnet-runtime-1695731444737.txt. Note that this file may contain personal data.

## Other Info:
Errors the extension may have encountered have likely been added to your terminal or VS Code windows. Please paste it/them here and describe what happened and the problem you are facing.

**Extension Version:** 1.0.0
**Error Message:** .NET Acquisition Failed: EPERM: operation not permitted, unlink 'c:\Users\anonymous\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.vscode-dotnet-runtime\.dotnet\7.0.11\dotnet.exe'
**Error Stack:** Error: .NET Acquisition Failed: EPERM: operation not permitted, unlink 'c:\Users\anonymous\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.vscode-dotnet-runtime\.dotnet\7.0.11\dotnet.exe'
    at c:\Users\anonymous\.vscode\extensions\vscode-dotnet-runtime-library\dist\Acquisition\DotnetCoreAcquisitionWorker.js:95:27
    at processTicksAndRejections (node:internal/process/task_queues:95:5)

**Reproduction Steps**
Recently updated C# extension to latest in Visual Code

**Expected behavior**
No errors

**Exceptions & Errors (if any)**

Failed to find dotnet info from path, falling back to acquire runtime via ms-dotnettools.vscode-dotnet-runtime
No compatible .NET runtime found. Minimum required version is 7.0.
[...]

PR #1443, opened 2023-10-03
1.8.1 Add Garbage Collection for Failed Uninstalls + Fix Uninstalls

nagilson:
Users can experience failure to uninstall, which our extension reports as an 'acquisition failure', however, based on the code and logs it returns as success and finishes installing successfully. It never tries to uninstall a 'legacy' no arch install unless the install succeeded. We would like to remove this failure and report it silently for our records, then try to uninstall again later when .net may not be in use. 

resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1413

We dont just delete the promise even if uninstall fails because it could be picked up as a legitimiate install or partial install and cause a lot of bugs, so it gets moved to a separate 'graveyard' for GCollection

### DIFF
--- a/vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
+++ b/vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
@@ -80,7 +80,7 @@ suite('DotnetCoreAcquisitionExtension End to End', function() {
     assert.include(result!.dotnetPath, context.version);
     await vscode.commands.executeCommand<string>('dotnet.uninstallAll', context.version);
     assert.isFalse(fs.existsSync(result!.dotnetPath));
-  }).timeout(40000);
+  }).timeout(400000);
 
   test('Install and Uninstall Multiple Versions', async () => {
     const versions = ['2.2', '3.0', '3.1'];
--- a/vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
@@ -22,6 +22,7 @@ import {
     DotnetPreinstallDetectionError,
     DotnetUninstallAllCompleted,
     DotnetUninstallAllStarted,
+    SuppressedAcquisitionError,
 } from '../EventStream/EventStreamEvents';
 import { IDotnetAcquireResult } from '../IDotnetAcquireResult';
 import { IAcquisitionWorkerContext } from './IAcquisitionWorkerContext';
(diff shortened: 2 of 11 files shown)


baronfel (approved):
The new event is a good call - we'll have to do GDPR classification on it once this merges and releases.

vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -80,7 +80,7 @@ suite('DotnetCoreAcquisitionExtension End to End', function() {
     assert.include(result!.dotnetPath, context.version);
     await vscode.commands.executeCommand<string>('dotnet.uninstallAll', context.version);
     assert.isFalse(fs.existsSync(result!.dotnetPath));
-  }).timeout(40000);
+  }).timeout(400000);
```
nagilson, line 83:
```suggestion
  }).timeout(40000);

```

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -15,28 +15,36 @@ import {
...
+    // These will become marked for attempted 'garbage collection' at the end of every acquisition.
+    private readonly installPathsGraveyardKey = 'installPathsGraveyard';
     public installingArchitecture : string | null;
     private readonly dotnetExecutable: string;
     private readonly timeoutValue: number;
 
-    private acquisitionPromises: { [version: string]: Promise<string> | undefined };
+    private acquisitionPromises: { [installKeys: string]: Promise<string> | undefined };
+
+
```
nagilson, line 47:
```suggestion
```

vscode-dotnet-runtime-library/src/test/unit/DotnetCoreAcquisitionWorker.test.ts
```diff
@@ -34,17 +36,19 @@ import {
...
+const expectedTimeoutTime = 6000;
+
 suite('DotnetCoreAcquisitionWorker Unit Tests', function () {
     const installingVersionsKey = 'installing';
     const installedVersionsKey = 'installed';
     const dotnetFolderName = `.dotnet O'Hare O'Donald`;
 
-    function getTestAcquisitionWorker(runtimeInstall: boolean, arch : string | null | undefined = undefined,
-        customEventStream? : MockEventStream , customContext? : MockExtensionContext): [DotnetCoreAcquisitionWorker, MockEventStream, MockExtensionContext]
+    function getTestAcquisitionWorker(runtimeInstall: boolean, arch? : string | null,
```
JL03-Yue, line 46: I appreciate the change here. It makes thing simpler

JL03-Yue (approved):
Ran the feature and the test and it looks good to me 🙂

vscode-dotnet-runtime-library/src/test/unit/DotnetCoreAcquisitionWorker.test.ts
nagilson: Thanks for your review! :)

PR #1469, opened 2023-10-18
Verify Global Install Files & Set SDK Telemetry to Match VS Code Setting

nagilson:
- Adds ability to detect when VS Code's telemetry setting has been changed
- Disables the .NET SDK telemetry when VS Code or VS Code Extension Telemetry settings are off on windows, warns on osx and linux
- Does not edit SDK telemetry if its already off 
- Also verifies hashes of .NET SDK Installer files and their download url, adds a box to accept the risk if no hash exists
- Adds a warning pop up if we fail to disable SDK telemetry 
- Fixes Ubuntu Package Feed Repository Key downloading by adding sudo, so it can pass
- Improves error message about how to install manually on WSL
- Updates telemetry setting to mention you should restart VS Code for the setting to take effect


nagilson:
The pipeline is currently failing to download vscode, which I am investigating, but it is not really a problem with the changes. It builds and tests fine on other machines

nagilson:
@JL03-Yue @joeloff @dsplaisted, Please leave your review comments on this PR https://github.com/nagilson/vscode-dotnet-runtime/pull/8. I am unable to request your review on that fork. This includes all of the global code, which @baronfel is reviewing. Since it is not merged, I am doing a PR into that PR.

@dsplaisted Please look at the changes to `TelemetryObserver`.

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    public async getExpectedGlobalSDKPath(specificSDKVersionInstalled : string, installedArch : string) : Promise<string>
+    {
+        if(os.platform() === 'win32')
+        {
+            // The program files should always be set, but in the off chance they are wiped, we can try to use the default as backup.
+            // Both ia32 and x64 machines will use 'Program Files'
+            // We don't anticipate a user would need to install the x86 SDK, and we don't have any routes that support that yet.
+            return process.env.programfiles ? path.resolve(path.join(process.env.programfiles, 'dotnet', 'sdk')) : path.resolve(`C:\\Program Files\\dotnet\\sdk\\`);
```
joeloff, line 170: This assumes that Program Files reside on a specific drive. That's not necessarily the case. Also, what about x64 SDK on arm64?

nagilson:
Note the acknowledgement on this PR https://github.com/nagilson/vscode-dotnet-runtime/pull/8 which I have closed, as it is no longer needed to have two PRs.

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
nagilson: This is true. This path will not be used unless the environment variable programfiles is not set. Basically, it is a last ditch effort if someone deleted the environment variables on their machine somehow, which for some reason there exists people out there will do that...

This currently only installs the .NET SDK that matches the machine architecture, so for now we don't need to worry about the case where there's an x64 SDK on arm64. Is there a large contingent of users that need that for some reason? If so, it would be good to know. Would you suggest adding the x64 to the path here in that corner-case?

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -36,6 +36,8 @@ import {
     registerEventStream,
     RuntimeInstallationDirectoryProvider,
     VersionResolver,
```
nagilson, line 38: per @/dsplaisted, we should also show the message about telemetry if it is enabled, because the first run message may get swallowed by CDK or C# extensions. We should add an env var for the future .NET SDKs and for CDK to pass to opt out of SDK Telemetry just for them, and also so we dont let the first run message get swallowed, so we can show SDK telemetry after that globally on the system outside of CDK or C# in vs code

nagilson: Done.

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
joeloff: Emulation was a big deal during .NET 6.0, but now that we have VS support for arm64, it's probably less of a concern for SDKs, and mostly just going to be runtimes where users get applications that were built for x64 and want to run those on arm64 machine.

joeloff (approved):

PR #1487, opened 2023-10-25
Add Global SDK Acquistion to the Runtime Extension

nagilson:
Rebrands the runtime extension to the .NET Install Tool so users dont need two extensions to download both the local runtime and global SDK separately.

PR changes look like a lot, but github is not showing the diff super well -- almost all of it is a 1-1 copy/move of functions to shared places / different files, with some slight modifications to allow calling from either a runtime context or sdk context.

### DIFF
--- a/vscode-dotnet-runtime-extension/package.json
+++ b/vscode-dotnet-runtime-extension/package.json
@@ -9,7 +9,7 @@
 	},
 	"license": "MIT",
 	"author": "Microsoft Corporation",
-	"displayName": ".NET Runtime Install Tool",
+	"displayName": ".NET Install Tool",
 	"description": "This extension installs and manages different versions of the .NET Runtime.",
 	"appInsightsKey": "02dc18e0-7494-43b2-b2a3-18ada5fcb522",
 	"icon": "images/dotnetIcon.png",
@@ -47,17 +47,17 @@
 		"commands": [
 			{
 				"command": "dotnet.reportIssue",
-				"title": "Report an issue with the .NET Runtime Install Tool.",
-				"category": ".NET Runtime Install Tool"
+				"title": "Report an issue with the .NET Install Tool.",
+				"category": ".NET Install Tool"
 			}
 		],
 		"configuration": {
-			"title": ".NET Runtime Install Tool",
+			"title": ".NET Install Tool",
 			"properties": {
 				"dotnetAcquisitionExtension.enableTelemetry": {
 					"type": "boolean",
 					"default": true,
-					"description": "Enable Telemetry for the .NET Runtime Install Tool. Restart VS Code to apply changes."
+					"description": "Enable Telemetry for the .NET Install Tool. Restart VS Code to apply changes."
 				},
 				"dotnetAcquisitionExtension.installTimeoutValue": {
 					"type": "number",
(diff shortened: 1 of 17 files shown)


vscode-dotnet-runtime-extension/package.json
```diff
@@ -66,7 +67,11 @@
 				},
 				"dotnetAcquisitionExtension.existingDotnetPath": {
 					"type": "array",
-					"description": "File Path to an existing installation of .NET."
+					"description": "File Path to an existing installation of a .NET Runtime."
+				},
+				"dotnetAcquisitionExtension.existingDotnetSDKPath": {
+					"type": "array",
+					"description": "File Path to an existing installation of a .NET SDK Folder."
```
baronfel, line 74: Big question - these are going to effectively be the same thing for a lot of users. Meaning I think lots of folks are setting these to the existing global install locations. Do we need both settings? Can we mark one deprecated and point to the more general one? How are these used distinctly in the code?

nagilson: Hm. When developing the SDK extension I didn't know what the app host was so I made it return the folder such as C:\Program Files\dotnet\sdk\7.0.113 because I thought there'd be a .NET exe in there, obviously there is not.

Some of the validation of 'did it install the right version of the sdk' is, checking for that versioned folder to exist.
We can return the same folder as the dotnet runtime such as C:\Program Files\dotnet, (this currently returns the same thing but with sdk at the end) though that doesnt guarantee the apphost will use the older sdk that someone might've requested if there's a new sdk on the machine (in fact, it wont, though not like adding the sdk folder to the path changes that.)

Someone could have used an old path to an app host that doesn't have any .NET SDK installs... I guess that'd still be fine and hopefully the app host will find it... given this information, do you still think its ok for them to share the existing variable? If so, I see no need to deprecate it.

nagilson: TLDR there is some mismatch between using the apphost folder as output for installing a specific sdk version, but I think this is the best choice so Ill make that change now and remove this variable

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -190,18 +217,69 @@ export function activate(context: vscode.ExtensionContext, extensionContext?: IE
...
+            return new Promise((resolve) => {
+                resolve(existingPath);
+            });
+        }
+        return new Promise((resolve) => {
+            resolve(null);
+        });
+    }
+
+    function getAcquisitonWorker(isRuntimeWorker : boolean) : DotnetCoreAcquisitionWorker
```
baronfel, line 246: nit: `Acquisition` is typo'd here.

baronfel (approved):
I went in expecting a lot more churn than there ended up being!  This is a good 'merge' of the two extensions, and I didn't find anything that looked wrong on its face. I think there could be opportunity for refactoring some of the larger methods later, but this PR is correctly scoped to getting the initial squishing of functionality together.

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: Thank you very much :) Resolved this feedback and the above. Please see the latest commit and description and let me know if it looks good, its a fairly significant change, small but important to get right

vscode-dotnet-runtime-extension/package.json
nagilson: I made them the same which required some changes

baronfel:
:shipit:

Issue #1485, opened 2023-10-24
Alter command structures for Linux package managers to use string[] instead of raw strings


baronfel:
I'd really suggest doing a string[] for commands here, and make sudo execution a boolean flag for each command. Without a string array, you're going to have to split and parse command lines per-token, and that gets very hairy very quickly. not to mention quoted arguments or multi-word arguments. trust me that moving to an array will make your life easier.

consider a signature like this for each command here:

```typescript
type Command = string[];
type FullCommand = { sudo: boolean; singleCommand: Command} | { sudo: boolean; compoundCommands: Command[] }
```

then in code you can pass the string[] of tokens to spawn/spawnSync, and if you need to orchestrate compoundcommands you can just concat the arrays for each sub-Command together, joining them with a "&&" token between.

_Originally posted by @baronfel in https://github.com/dotnet/vscode-dotnet-runtime/pull/1081#discussion_r1367469979_

PR #1490, opened 2023-10-26
Rewrite Command Execution API & Command Parsing

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1485 to avoid parsing spaces and commands by adding a new command type.

### DIFF
--- a/vscode-dotnet-runtime-library/distro-data/distro-support.json
+++ b/vscode-dotnet-runtime-library/distro-data/distro-support.json
@@ -1,17 +1,90 @@
 {
 	"$schema": "../../Documentation/json-schema/distro-support-schema.json",
 	"Ubuntu": {
-		"installCommand": "sudo apt-get update && sudo apt-get install -y {0}",
-		"uninstallCommand": "sudo apt-get remove {0}",
-		"updateCommand": "sudo apt-get update && apt-get upgrade -y {0}",
-		"isInstalledCommand": "sudo apt list --installed {0}",
-		"packageLookupCommand": "dpkg -l {0}",
+		"installCommand":
+		[
+			{
+				"runUnderSudo": true,
+				"commandRoot": "apt-get",
+				"commandParts": ["update"]
+			},
+			{
+				"runUnderSudo": true,
+				"commandRoot": "apt-get",
+				"commandParts": ["install", "-y", "{0}"]
+			}
+		],
+		"uninstallCommand":
+		[
+			{
+				"runUnderSudo": true,
+				"commandRoot": "apt-get",
+				"commandParts": ["remove", "{0}"]
+			}
+		],
+		"updateCommand":
+		[
+			{
+				"runUnderSudo": true,
+				"commandRoot": "apt-get",
+				"commandParts": ["update"]
+			},
+			{
+				"runUnderSudo": true,
+				"commandRoot": "apt-get",
+				"commandParts": ["upgrade", "-y", "{0}"]
+			}
+		],
+		"isInstalledCommand":
+		[
+			{
+				"runUnderSudo": true,
+				"commandRoot": "apt",
+				"commandParts": ["list", "--installed", "{0}"]
+			}
+		],
+		"packageLookupCommand":
+		[
+			{
+				"runUnderSudo": false,
+				"commandRoot": "dpkg",
+				"commandParts": ["-l", "{0}"]
+			}
+		],
 		"expectedDistroFeedInstallDirectory" : "/usr/lib/dotnet/sdk",
         "expectedMicrosoftFeedInstallDirectory" : "/usr/bin/dotnet",
-		"installedSDKVersionsCommand": "dotnet --list-sdks",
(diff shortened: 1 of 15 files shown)


vscode-dotnet-runtime-library/distro-data/distro-support.json
```diff
@@ -1,17 +1,90 @@
 {
 	"$schema": "../../Documentation/json-schema/distro-support-schema.json",
 	"Ubuntu": {
-		"installCommand": "sudo apt-get update && sudo apt-get install -y {0}",
-		"uninstallCommand": "sudo apt-get remove {0}",
-		"updateCommand": "sudo apt-get update && apt-get upgrade -y {0}",
-		"isInstalledCommand": "sudo apt list --installed {0}",
-		"packageLookupCommand": "dpkg -l {0}",
+		"installCommand":
```
baronfel, line 4: the schema file needs to be updated for this change, I believe.

vscode-dotnet-runtime-library/distro-data/distro-support.json
```diff
@@ -1,17 +1,90 @@
...
+		[
+			{
+				"runUnderSudo": true,
+				"commandRoot": "apt-get",
+				"commandParts": ["update"]
+			},
+			{
+				"runUnderSudo": true,
+				"commandRoot": "apt-get",
+				"commandParts": ["install", "-y", "{0}"]
```
baronfel, line 14: These `{0}` placeholders are .NET format string style, but they don't give any semantic meaning for review - should we have a named placeholder like `{packageName}` instead that we can explicitly substitute in the command processor?

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -137,33 +138,45 @@ You will need to restart VS Code after these changes. If PowerShell is still not
...
             {
                 knownError = true;
                 const err = Error(this.noPowershellError);
                 error = err;
             }
             else
             {
-                command = commandWorking[0];
+                command = possibleCommands[workingCommandIndex];
             }
```
baronfel, line 166: nit: instead of returning indexes, why not return the found command to prevent any change of array indexing errors?

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -194,33 +194,39 @@ We cannot verify .NET is safe to download at this time. Please try again later.`
...
+                CommandExecutor.makeCommand(`/usr/bin/open`, [])
+            ];
 
-            const openAvailable = await this.commandRunner.TryFindWorkingCommand([`command -v open`, `/usr/bin/open`]);
-            if(openAvailable[1] && openAvailable[0] !== 'command -v open')
-            {
-                commandToExecute = openAvailable[0];
-            }
-            else if(!openAvailable[1])
+            const workingCommandIndex = await this.commandRunner.tryFindWorkingCommand(possibleCommands);
```
baronfel, line 203: similar comment here - why return the index instead of a nullable object?

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -87,114 +85,120 @@ Please install the .NET SDK manually by following https://learn.microsoft.com/en
...
                 {
-                    this.eventStream.post(new CommandExecutionUserCompletedDialogueEvent(`The command ${commandFollowUps} successfully ran under sudo.`));
+                    this.eventStream.post(new CommandExecutionUserCompletedDialogueEvent(`The command ${fullCommandString} successfully ran under sudo.`));
                     resolve(commandResultString);
                 }
             });
         });
     }
 
+    public async executeMultipleCommands(commands: CommandExecutorCommand[], options?: any): Promise<string[]> {
```
baronfel, line 106: 💯 love this

baronfel (approved):
Excellent set of changes! Left just a few notes.

vscode-dotnet-runtime-library/distro-data/distro-support.json
nagilson: Indeed, I will need to rewrite it again for when I add the searchCommand, which I am doing now! :)

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
nagilson: Good point, that makes it much cleaner. The way it was initially written didnt allow for this but now it can be done.

vscode-dotnet-runtime-library/distro-data/distro-support.json
nagilson: Hm, I like this placeholder idea a lot better. Done! Thanks for the suggestion.

Issue #1516, opened 2023-11-06
[NETE2ESDK] It still keep downloading the older SDK even .NET acquisition failed with version conflicts.


NicoleWang001:
# Describe the bug
It still downloading the older SDK even .NET acquisition failed with version conflicts.

## Other Info:

**Reproduction Steps**
1. Run the sample command ‘Install .NET SDK Globally via .NET Install Tool (Former Runtime Extension)‘.
2.  Use the version it provided, which is 7.0.403. Hit enter.
3. After .NET 7.0.403 installed, then try to install the older SDK 7.0.402

**Expected behavior**
On windows, the older SDK installation should be failed with version conflicts.

**Exceptions & Errors (if any)**
Even .NET acquisition failed, but still keep downloading the older SDK e
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/cfd73702-0073-4815-aa50-52d153b07010)

**Further technical details**
OS: Win 11 X64 22H2(OS Build 22621.2428)
VSCode: 1.84
Extensions: dotnet-install-tool-2.0.0

nagilson:
Closing this as resolved by the above PR.

PR #1524, opened 2023-11-08
Remove Infinite Print Loop After Error, Refactor

nagilson:
- Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1516 by fixing the code to stop printing ....... after any error, not just a version error. this requires a large refactor because we need to know the installKey of the command that fails so we can delete it from the installs in progress that the error handler keeps track of.

- While doing this, I decided to de-dupe some of the other code by spreading the worker context around to all classes.

I also fixed a lot of typos that were bugging me...


nagilson:
This PR is becoming the everything PR. I am going to split it up

baronfel (approved):
_whew_ this is a lot. I appreciated the commit-by-commit on this one. The main cause of change seems to be that installkey tracking, which does make sense to me.  Let's see if we can get this tested!

nagilson:
Thank you for your review!

PR #1532, opened 2023-11-09
Migrate to x64 VS Code instances for testing as 32-bit is unsupported

nagilson:
- Migrate to 64 bit vscode testing platform as 32-bit support is dropped and was the old default, which now causes failure on all tests: https://github.com/microsoft/vscode/issues/197554 and not committed to updating to a newer version of vscode yet where it's fixed.


JL03-Yue (approved):
Looks straightforward to me🙂

PR #1534, opened 2023-11-09
Log Installer File Verification

nagilson:
This will improve our ability to repudiate and view logs when failures with sign checking occur

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/GlobalInstallerResolver.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/GlobalInstallerResolver.ts
@@ -10,6 +10,7 @@ import * as os from 'os';
 import * as path from 'path';
 import { VersionResolver } from './VersionResolver';
 import { DotnetFeatureBandDoesNotExistError,
+        DotnetFileIntegrityCheckEvent,
         DotnetInvalidReleasesJSONError,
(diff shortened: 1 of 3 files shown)


vscode-dotnet-runtime-library/src/Acquisition/GlobalInstallerResolver.ts
```diff
@@ -249,6 +250,10 @@ Please report this issue so it can be remedied or investigated.`));
                             this.eventStream.post(releaseJsonErr);
                             throw releaseJsonErr.error;
                         }
+                        else
+                        {
+                            this.eventStream.post(new DotnetFileIntegrityCheckEvent(`This installer file is hosted on an expected domain https://download.visualstudio.microsoft.com/.`));
```
JL03-Yue, line 255: https://download.visualstudio.microsoft.com/ Does it redirect to an empty page?

JL03-Yue (approved):
Looks good to me

vscode-dotnet-runtime-library/src/Acquisition/GlobalInstallerResolver.ts
nagilson: That's a good thing to point out. This is in the log so it should be fine. I am a bit surprised its empty, but it is just a download server, so kinda makes sense the root is 404

Issue #1515, opened 2023-11-06
[NETE2ESDK][WinArm64].NET acquisition failed while installing .NET SDK on Win Arm64


NicoleWang001:
# Describe the bug

.NET acquisition failed while installing .NET SDK on Win Arm64, but the SDK is installed successfully.

## Other Info:

**Reproduction Steps**
1. Run the sample command ‘Install .NET SDK Globally via .NET Install Tool (Former Runtime Extension)‘.
2. Use the version it provided, which is 7.0.103. Hit enter.
3. Observe

**Expected behavior**
.NET SDK could be downloaded and installed successfully.

**Exceptions & Errors (if any)**
NET acquisition failed
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/a663ef7d-51e3-4a25-8cdd-f224bf6bc7f7)

**Further technical details**
OS: Win 11 Arm64 22H2(OS Build 22621.2428)
VSCode: 1.84
Extensions: dotnet-install-tool-2.0.0

Note:
1. This issue is only repro on Win Arm64

nagilson:
It did not have permission to delete the installer file for some reason on x64. This should be cleaned up later so I will make this a non terminal failure

PR #1535, opened 2023-11-09
Dont fail install if we dont have permission to clean up installer file

nagilson:
This is in the dist folder, it should* be cleaned up worst case by upgrade/uninstall of the extension, we dont want to fail everything if we dont have perm to delete, this is a win arm64 problem for whichever reason

- Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1515 by not failing the install if we dont have permission to delete the installer file we download, which happens on arm64


JL03-Yue (approved):
Looks good to me

PR #1540, opened 2023-11-10
Fix Axios Storage Handler to Handle Timeout & TLS Handshake Properly

nagilson:
https://axios-cache-interceptor.js.org/guide/storages The storage handler is supposed to prevent timed out requests from going into the cache per the documentation which it does not do.

Also, VS Code will try to unwrap a circular reference which the request will be circular in the TLS handshake. This is a 'bug' in serializeRequestArguments in extensionHostProcess.js from VS Code. This will prevent us from removing from the extension state in the future as trying to access it will cause an exception and cancel the request. Unfortunately, there is no good way I could find to filter things out of the cache.

I've changed to a memory based cache and @baronfel is OK with this change. This means the cache will not persist over different vscode windows, which is not optimal. However, until vs code does not fail with circular json references or non json objects with this cache, we do not seem to have ways to filter what goes into the cache and what does not. Even for one request, the no cache method still ends up putting things into the storage, see: https://axios-cache-interceptor.js.org/config/request-specifics#cache.

Then, the request object given to set does allow us to filter some of the content per request, but even then, one request seems to cause multiple potential cache entries. In addition, I looked at the code of the axios-cache-interceptor- and the loading state never gets updated until something is in the cache, so my previous code was based on incorrect assumptions, even if the logic would have worked if we had that kind of graunlar control. 

Perhaps I will raise an issue on the vscode repo, once that is complete, we could go back to the older more performant caching method.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
@@ -27,10 +27,21 @@ const mementoBasedAxiosCacheInterceptorStorage = (extensionStorage: IExtensionSt
         // tslint:disable-next-line
         set(key: string, value: any, request? : CacheRequestConfig)
          {
-             try
-             {
+            if(value.state === 'loading')
+            {
+                setTimeout(() =>
+                {
+                    if(value.state === 'loading')
+                    {
+                        return; // The web request is a timeout, do NOT cache its result, as the result is junk
+                    }
+                }, websiteTimeoutMs); // Give web requests
+            }
+
+            try
+            {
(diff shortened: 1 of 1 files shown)


JL03-Yue (approved):
Thanks for the detailed explanation. It's great to know the intention behind switching to memory-based controls. Does @baronfel would like to take a look?

nagilson:
Thanks for taking a look :) He did approve of the idea, but lets wait and see if he has any feedback in a bit. If not, then we will merge this.

nagilson:
Actually, this bug is becoming very problematic for testing anything else in the extension, so I am going to merge it now so I can import it into other PRs.

PR #1549, opened 2023-11-14
Improve timeout error handling

nagilson:
Axios already has its own timeout, so I dont think we need a 2nd timeout handler,  I was able to confirm it works by using https://httpstat.us/524 with a miniscule timeout value, and it does report the time it took to timeout in a similar fashion. We will get more information from errors this way when it is available.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
@@ -38,8 +38,9 @@ export namespace errorConstants {
 }
 
 export namespace timeoutConstants {
-    export const timeoutMessage = `.NET installation timed out. You may need to change the timeout time if you have a slow connection. Please see: https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md#install-script-timeouts.
-Note that our websites may be blocked in China or experience significant timeouts, and you may want to install .NET manually.`;
+    export const timeoutMessage = `.NET install timed out.
+You should change the timeout if you have a slow connection. See: https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md#install-script-timeouts.
+Our CDN may be blocked in China or experience significant slowdown, in which case you should install .NET manually.`;
     export const moreInfoOption = 'Change Timeout Value';
 }
 
--- a/vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
@@ -231,15 +231,15 @@ export class WebRequestWorker
             {
                 if(isAxiosError(error))
                 {
-                    let formattedError = error as AxiosError;
-                    const genericError = new Error(
-`Request to ${this.url} Failed: ${formattedError.message}. Aborting.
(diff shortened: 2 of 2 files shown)


vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
```diff
@@ -38,8 +38,9 @@ export namespace errorConstants {
 }
 
 export namespace timeoutConstants {
-    export const timeoutMessage = `.NET installation timed out. You may need to change the timeout time if you have a slow connection. Please see: https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md#install-script-timeouts.
```
JL03-Yue, line 41: That's great explanation. I wonder if Chet has any comment on this.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -208,22 +198,28 @@ export class WebRequestWorker
...
-                let formattedError = error as Error;
-                if ((formattedError.message as string).toLowerCase().includes('block')) {
-                    formattedError = new Error(`Software restriction policy is blocking .NET installation: Request to ${this.url} Failed: ${formattedError.message}`);
+                if(isAxiosError(error))
+                {
+                    const axiosBasedError = error as AxiosError;
+                    const summarizedError = new Error(
+`Request to ${this.url} Failed: ${axiosBasedError.message}. Aborting.
+${axiosBasedError.cause? `Error Cause: ${axiosBasedError.cause!.message}` : ``}
+Please ensure that you are online.
```
JL03-Yue, line 211:
```suggestion
Please ensure that you have network connection?
```

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -208,22 +198,28 @@ export class WebRequestWorker
...
+If you're on a proxy and disable registry access, you must set the proxy in our extension settings. See https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md.`);
+                    this.eventStream.post(new WebRequestError(summarizedError));
+                    throw summarizedError;
                 }
                 else
                 {
-                    formattedError = new Error(`Please ensure that you are online: Request to ${this.url} Failed: ${formattedError.message}.
-If you are on a corporate proxy with a result of 400, you may need to manually set the proxy in our extension settings. Please see https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md for more details.
-If your proxy requires credentials and the result is 407: we cannot currently handle your request, and you should configure dotnet manually following the above link.`);
+                    const genericError = new Error(`Web Request to ${this.url} Failed: ${error.message}. Aborting. Please ensure that you are online.`);
```
JL03-Yue, line 219:
```suggestion
                    const genericError = new Error(`Web Request to ${this.url} Failed: ${error.message}. Aborting. Please ensure that you have network connection.`);
```
I think it's great explanations! I wonder if "network connection" is more intuitive, but you have more understanding of the context🙂

nagilson: Thank you, I think this is a good suggestion, but I always try to keep errors as small as possible, as the longer the message is, the less likely someone is to read them.

PR #1554, opened 2023-11-15
Improve bugs with installKey management

nagilson:
I did an audit of the existing code shared between global sdk installs, and local sdk and local runtime installs that keeps track of what we are installing. I did this audit because if you uninstall the global sdk via something like program files, it would still think it was installed. 

I found several big gaps in the existing logic, though maybe I am missing something, because I am surprised it worked this well given these flaws:
- Uninstall does not do anything for global installs, yet it removes them from the list of known installed versions or installing versions
- Global installs get added to the 'grave yard' for collection even though they cant be collected
- The INSTALLING version key does NOT get removed upon error, so it can just stay there forever
- The acquisition promise that appears to be part of a 'we are currently installing this already' check is NOT removed when the install succeeds

I find it alarming that none of the tests failed when I changed this behavior. Those should probably be updated too to cover some of these cases. All of these early 4 commits at minimum should have triggered test failure


nagilson:
This can be reviewed once https://github.com/dotnet/vscode-dotnet-runtime/pull/1524 is merged.

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -79,8 +82,14 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
...
 
-        await this.context.extensionState.update(this.installingVersionsKey, []);
-        await this.context.extensionState.update(this.installedVersionsKey, []);
+        // This does not uninstall global things yet, so don't remove their keys.
+        const installingVersions = this.context.extensionState.get<string[]>(this.installingVersionsKey, []);
+        const remainingInstallingVersions = installingVersions.filter(x => this.isGlobalInstallKey(x));
+        await this.context.extensionState.update(this.installingVersionsKey, remainingInstallingVersions);
+
+        const installedVersions = this.context.extensionState.get<string[]>(this.installedVersionsKey, []);
+        const remainingInstalledVersions = installedVersions.filter(x => this.isGlobalInstallKey(x));
```
JL03-Yue, line 91: Does here repetitively define the `installedVersions` and `remainingInstallingVerisons` twice?  and call `this.context.extensionState.update` a second time?

nagilson: That's a good question. They are a bit different. There are keys for 'installed' versions and 'installing' versions and those are separate, since one maintains things that have not yet finished installing, and the other is for successfully completed installs in the past. The variable names are rather similar which could probably be improved.

JL03-Yue (approved):
Looks great to me.

PR #1555, opened 2023-11-15
Dont fail for unknown versions of Ubuntu by estimating a closest version

nagilson:
The theory is a version with the same major but with the maximum lower minor will be the closest in behavior, then the maximum version that is lower than the current version is most likely to have the same behavior. If there is no smaller known version, it will just use the minimum version that is higher.

This logic may not always work. But with the given circumstances, we are covered by including the first version where preinstall commands were not needed


JL03-Yue (approved):
looks good to me

nagilson:
Thank you :)

PR #1588, opened 2023-11-27
update default download timeout to 10 minutes and update documentatio…

Shadowyfigurings87:
…n universally


nagilson:
No worries, but this change is just whitespace if you see the diff ;)

PR #1590, opened 2023-11-27
Add PMC Feed Earlier & Fix PMC Feed Addition with Apt-Cache

nagilson:
If we dont run sudo apt-get update first then the search command will have no idea PMC packages exist.

In addition, regarding when we add the PMC and how we check for what packages of dotnet are available, the old code is incorrect and Im surprised it worked correctly but see how it happened. 

The following code relies on the function myVersionPackages

-  uninstall, update, install, should update (dotnetpackageexistsonsystem), getglobalsdkpath, getrecommendeddotnetversion

Some of those may have incorrect information based on whether or not the PMC was added before by the install command first before they were called...

This fixes said issue.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/IDistroDotnetSDKProvider.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/IDistroDotnetSDKProvider.ts
@@ -166,7 +166,7 @@ export abstract class IDistroDotnetSDKProvider {
         return supportedType && this.versionResolver.getFeatureBandFromVersion(fullySpecifiedVersion) === '1';
     }
 
-    protected async myVersionPackages(installType : LinuxInstallType) : Promise<LinuxPackageCollection[]>
+    protected async myVersionPackages(installType : LinuxInstallType, haveTriedFeedInjectionAlready = false) : Promise<LinuxPackageCollection[]>
     {
         const availableVersions : LinuxPackageCollection[] = [];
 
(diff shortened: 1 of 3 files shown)


vscode-dotnet-runtime-library/src/Acquisition/IDistroDotnetSDKProvider.ts
```diff
@@ -197,9 +197,60 @@ export abstract class IDistroDotnetSDKProvider {
             }
         }
 
+        if(availableVersions.length === 0 && !haveTriedFeedInjectionAlready)
+        {
+            // PMC is only injected and should only be injected for MSFT feed distros.
+            // Our check runs by checking the feature band first, so that needs to be supported for it to fallback to the preinstall command check.
+            const fakeVersionToCheckMicrosoftSupportStatus = '6.0.1xx';
```
JL03-Yue, line 204: Is there chances that the version might be lower than '6.0.1xx'?

JL03-Yue (commented):
Overall it's great work! The code is very clear and straightforward to me. I have one question related but everything else looks great!

vscode-dotnet-runtime-library/src/Acquisition/IDistroDotnetSDKProvider.ts
nagilson: That's a good question -- it looks like older versions that are now unsupported do support older versions of dotnet. But the 6.0 is really just a placeholder version, and it will get mapped to whatever is closest. Rereading this a month later, this code is clear but a bit more complicated than perhaps necessary, and the comment above appears to be meant to allude to the fact that 6.0 is just a placeholder to pass the unsupported check in `getDotnetVersionSupportStatus`

baronfel (approved):

Issue #1520, opened 2023-11-07
[NETE2ESDK][Ubuntu22.04] Installing different version .net SDK failed with distro issue on Ubuntu22.04


NicoleWang001:
# Describe the bug
 Installing different version .net SDK failed with distro issue on Ubuntu22.04

## Other Info:

**Reproduction Steps**
1. On Ubuntu 22.04 x64
2. Run the sample command ‘Install .NET SDK Globally via .NET Install Tool (Former Runtime Extension)‘ and enter 6.0.1xx
3. After 6.0.124 is installed
4. Run the sample command ‘Install .NET SDK Globally via .NET Install Tool (Former Runtime Extension)‘ and enter 7.0.103
5. Observe

**Expected behavior**
1. The .net SDK 7.0.103 is installed successfully

**Exceptions & Errors (if any)**
Installing .net SDK 7.0.103 failed with distro issue
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/d0a8c8a9-9fce-4c38-8b63-e9146b0cd458)

**Further technical details**
OS: Ubuntu 22.04
VSCode: 1.84
Extensions: dotnet-install-tool-2.0.0

v-xiaofchen:
This issue also repro on Ubuntu 22.04 Arm64.

NicoleWang001:
Met the same error if installing the same .NET SDK again after restarted VSCode.

v-xiaofchen:
This issue also repro on Ubuntu 23.10.
![ubuntu23 101](https://github.com/dotnet/vscode-dotnet-runtime/assets/99376012/16ecbb53-7644-468e-ad76-a9906f5f00ad)

nagilson:
The problem here is we should be reading the trimmed output of `readlink -f /usr/bin/dotnet` once we know we can validate that the share or lib folder is the true location for this. The bin location is used by both install types as a symlink

nagilson:
This should be resolved now.

NicoleWang001:
Close it as fixed.
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/8dca579c-b7a3-458e-b985-c9a030f7c719)
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/e3367d74-a434-41d6-8efe-a6c36544732c)

PR #1612, opened 2024-01-03
Use Installer Type Conforming Installation Directory for Ubuntu

nagilson:
For https://github.com/dotnet/vscode-dotnet-runtime/issues/1520
The /bin/ folder is a symlink that gets used by both distro support types, PMC and Distro Built In.
I got confirmation from Cannonical that lib is where they always put their files and we should expect our PMC debs to place them into share. This is a much simpler change that reading the symlink.

I think I put bin there when testing because thats what I saw the OS doing but it was in reality a symlink.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/GenericDistroSDKProvider.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/GenericDistroSDKProvider.ts
@@ -13,6 +13,8 @@ import { IDistroDotnetSDKProvider } from './IDistroDotnetSDKProvider';
 
 export class GenericDistroSDKProvider extends IDistroDotnetSDKProvider
 {
+    protected resolvePathAsSymlink = true;
+
     public async installDotnet(fullySpecifiedVersion : string, installType : LinuxInstallType): Promise<string>
     {
         await this.injectPMCFeed(fullySpecifiedVersion, installType);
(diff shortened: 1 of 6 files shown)


nagilson:
This is not quite good enough because it will still think BIN is a custom folder, it needs to resolve the symlink.

vscode-dotnet-runtime-library/src/Acquisition/GenericDistroSDKProvider.ts
```diff
@@ -37,6 +39,18 @@ export class GenericDistroSDKProvider extends IDistroDotnetSDKProvider
         {
             commandResult[0] = commandResult[0].trim();
         }
+
+        if(commandResult && this.resolvePathAsSymlink)
+        {
+            let symLinkReadCommand = this.myDistroCommands(this.readSymbolicLinkCommandKey);
+            symLinkReadCommand = CommandExecutor.replaceSubstringsInCommands(symLinkReadCommand, this.missingPathKey, commandResult[0]);
```
JL03-Yue, line 46: Looks good to me. Could you explain the `missPathKey` variable here?

vscode-dotnet-runtime-library/src/test/unit/LinuxDistroTests.test.ts
```diff
@@ -104,7 +104,7 @@ Microsoft.NETCore.App 7.0.5 [/usr/lib/dotnet/shared/Microsoft.NETCore.App]`;
         if(shouldRun)
         {
             await provider.getInstalledGlobalDotnetPathIfExists(installType);
-            assert.equal(mockExecutor.attemptedCommand, 'which dotnet');
+            assert.equal(mockExecutor.attemptedCommand, 'readlink -f /usr/bin/dotnet');
```
JL03-Yue, line 107: Thanks for the explanation for the tests!

JL03-Yue (approved):

vscode-dotnet-runtime-library/src/Acquisition/GenericDistroSDKProvider.ts
nagilson: ah yes, the missing Path Key is the string {path} that gets resolved to become the path that needs to be converted from a symlink to the true path! Thank you for your review :)

PR #1613, opened 2024-01-03
Fix SDK Linux Updates, & Add Cancellation Messages - 2.0.1 Release

nagilson:
Contains the final changes and version bumps for 2.0.1.

Besides the READ.ME and version bump, this PR includes a few changes that are smaller:
- Fixes to the update logic on linux, where we update instead of installing all over again. This did not used to work.
- Improves logging so we can tell in logs where people are hitting failures, e.g. if they declined the password prompt, or if they cancelled the installer we now have a specific message for that
- Fix a bug that allowed you to downgrade the SDK 
- Added a cancellation message and removed the failure message for when we decide the SDK cant be installed or the user cancels the operation. Made the status bar not flash red when this happens, which it used to before.
- Resolves symlinks on RHEL because it also can use symlinks for the dotnet path

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/LinuxVersionResolver.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/LinuxVersionResolver.ts
@@ -318,7 +318,7 @@ export class LinuxVersionResolver
         {
             return '0';
         }
-        return updateOrRejectState;
+        return String(updateOrRejectState);
     }
 
     /**
--- a/vscode-dotnet-runtime-library/src/Acquisition/LinuxVersionResolver.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/LinuxVersionResolver.ts
@@ -10,6 +10,7 @@ import
     DotnetAcquisitionDistroUnknownError,
     DotnetConflictingLinuxInstallTypesError,
     DotnetCustomLinuxInstallExistsError,
+    DotnetInstallLinuxChecks,
     DotnetUpgradedEvent
 } from '../EventStream/EventStreamEvents';
 import { GenericDistroSDKProvider } from './GenericDistroSDKProvider'
--- a/vscode-dotnet-runtime-library/src/Acquisition/LinuxVersionResolver.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/LinuxVersionResolver.ts
@@ -9,7 +9,8 @@ import
 {
     DotnetAcquisitionDistroUnknownError,
     DotnetConflictingLinuxInstallTypesError,
-    DotnetCustomLinuxInstallExistsError
+    DotnetCustomLinuxInstallExistsError,
+    DotnetUpgradedEvent
 } from '../EventStream/EventStreamEvents';
 import { GenericDistroSDKProvider } from './GenericDistroSDKProvider'
 import { VersionResolver } from './VersionResolver';
@@ -67,6 +68,7 @@ export class LinuxVersionResolver
     protected distroSDKProvider : IDistroDotnetSDKProvider | null = null;
     protected commandRunner : ICommandExecutor;
     protected versionResolver : VersionResolver;
+    public okUpdateExitCode = 11188; // Arbitrary number that is not shared or used by other things we rely on as an exit code
 
     public conflictingInstallErrorMessage = `A dotnet installation was found which indicates dotnet that was installed via Microsoft package feeds. But for this distro and version, we only acquire .NET via the distro feeds.
     You should not mix distro feed and microsoft feed installations. To continue, please completely remove this version of dotnet to continue by following https://learn.microsoft.com/dotnet/core/install/remove-runtime-sdk-versions?pivots=os-linux.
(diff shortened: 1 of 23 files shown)


vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
```diff
@@ -62,11 +64,14 @@ export async function callWithErrorHandling<T>(callback: () => T, context: IIssu
     catch (caughtError)
     {
         const error = caughtError as Error;
-        context.eventStream.post(isAcquisitionError ?
-            new DotnetCommandFailed(error, context.commandName, getInstallKeyFromContext(acquireContext?.acquisitionContext)) :
-            // The output observer will keep track of installs and we don't want a non-install failure to make it think it should -=1 from the no. of installs
-            new DotnetNotInstallRelatedCommandFailed(error, context.commandName)
-        );
+        if(!(error instanceof DotnetInstallExpectedAbort) && error && error.constructor && error.constructor.name !== 'UserCancelledError')
```
JL03-Yue, line 67: I really like how new error messages are added. It adds new error types and reduce the number of irrelevant error posts for the user. Nit: I wonder if it is easier to write this way
```suggestion
        if(!(error instanceof DotnetInstallExpectedAbort) && !(error instanceof UserCancelledError))
```

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -111,6 +115,14 @@ ${stderr}`));
                     this.context?.eventStream.post(new CommandExecutionUserCompletedDialogueEvent(`The command ${fullCommandString} failed to run under sudo.`));
                     if(terminalFailure)
                     {
+                        if(error.code === 126)
```
JL03-Yue, line 118: I tried and really like how this adds new error handling checks if command invoked cannot execute if user rejects the password prompt. Is it possible to know what `126` refers to in this case?

vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
```diff
@@ -127,6 +127,23 @@ export class OutputChannelObserver implements IEventStreamObserver {
         // Nothing to dispose
     }
 
+    private updateDownloadIndicators(installKey : string | null | undefined)
+    {
+        if(installKey && installKey !== 'null')
```
JL03-Yue, line 132: Is it possible that installKey could be empty?
```suggestion
        if(installKey && installKey !== 'null' && installKey !== '')
```

vscode-dotnet-runtime-library/src/Acquisition/LinuxVersionResolver.ts
```diff
@@ -250,34 +254,43 @@ export class LinuxVersionResolver
...
             const existingGlobalInstallSDKVersion = await this.distroSDKProvider!.getInstalledGlobalDotnetVersionIfExists();
             if(existingGlobalInstallSDKVersion && Number(this.versionResolver.getMajorMinor(existingGlobalInstallSDKVersion)) ===
                 Number(this.versionResolver.getMajorMinor(fullySpecifiedDotnetVersion)))
             {
-                if(Number(this.versionResolver.getMajorMinor(existingGlobalInstallSDKVersion)) > Number(this.versionResolver.getMajorMinor(fullySpecifiedDotnetVersion)))
+                const isPatchUpgrade = Number(this.versionResolver.getFeatureBandPatchVersion(existingGlobalInstallSDKVersion)) <
+                    Number(this.versionResolver.getFeatureBandPatchVersion(fullySpecifiedDotnetVersion));
+
+                if(Number(this.versionResolver.getMajorMinor(existingGlobalInstallSDKVersion)) > Number(this.versionResolver.getMajorMinor(fullySpecifiedDotnetVersion))
+                || Number(this.versionResolver.getFeatureBandFromVersion(existingGlobalInstallSDKVersion)) > Number(this.versionResolver.getFeatureBandFromVersion(fullySpecifiedDotnetVersion)))
```
JL03-Yue, line 272: This changes really make sense to me as it has additional checks for feature band versions besides checking major and minor versions!

JL03-Yue (approved):
The new release looks great to me overall. It really inspired my learning and there are some comment for reference. Look forward to seeing the new release of the runtime extension🙂

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
nagilson: Thanks :) And 126 is what vscode sudo-prompt returns if the user refuses the password request, I dont think there is any other meaning to it besides that, but it might be wise to add a comment stating this

vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
nagilson: '' has a 'truthiness' value of 'false', so this additional condition is unnecessary, but this is good thinking because it could be '' :)

Issue #1589, opened 2023-11-27
Extension asks for password 3-6 times on ubuntu and rhel


nagilson:
This is due to a limitation with sudo-prompts, we should elevate differently so users do not have this bad experience

PR #1620, opened 2024-01-09
Reduce Linux Prompt to Single Ask

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1589
Currently, when the Linux SDK Global installs ask for permission, they need to ask it every single time a command is executed. This will change that problem by spawning a master process. The master process then communicates with the VS Code Extension over a shared user, non-world writeable folder.

The files exchanged are 'output', 'ok', 'command'. The OK file is used to ensure the process is alive, and if not, spawn it again.
The 'command' file is used to pass a command to the master process, which the process then cleans up. The 'output' is a txt file that is output by the process tells the extension to read the files representing stderr, stdout, and return status, like how current command execution in the extension works.

The master process is a shell script so that it can be platform agnostic and does not need a binary.

In addition, we validate that the commands requested are of a fixed set of commands passed to the master process upon creation and change the permissions of files in webpack so they are set from the very beginning.

### DIFF
--- a/vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
+++ b/vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
@@ -2,56 +2,36 @@
 EXECFOLDER=$1 # First argument is the working folder as this is launched with cwd of /root
 OKSIGNALFILE="$EXECFOLDER/ok.txt"
 COMMANDTORUNFILE="$EXECFOLDER/command.txt"
-OUTPUTFILE="/home/viru/vm.txt"
+OUTPUTFILE="/home/viru/om.txt"
+end=$((SECONDS+3600))
 while true
 do
         stop=false
-        until ((stop))
+        while [ $SECONDS -lt $end ];
         do
-            sleep 5
-            echo $(ls) >> "$OUTPUTFILE"
-            echo "$EXECFOLDER" >> "$OUTPUTFILE"
-            echo $(ls EXECFOLDER) >> "$OUTPUTFILE"
-            echo "$OKSIGNALFILE" >> "$OUTPUTFILE"
-            echo "$COMMANDTORUNFILE" >> "$OUTPUTFILE"
             if test -f "$COMMANDTORUNFILE"; then
-                echo "COMMAND FILE FOUND" >> "$OUTPUTFILE"
+                # echo "COMMAND FILE FOUND" >> "$OUTPUTFILE" # Leave this here as an example of debugging
                 COMMAND="$(cat "$COMMANDTORUNFILE" | awk '{$1=$1;print}')"
                 for validCmd in "${@:2}"
                 do
-                    echo "$COMMAND" >> "$OUTPUTFILE"
-                    echo "$validCmd" >> "$OUTPUTFILE"
                     if [ "$COMMAND" == "$validCmd" ]; then
                         IFS=' ' read -ra COMMANDARGS <<< "$COMMAND"
-                        echo "VALID CMD FOUND" >> "$OUTPUTFILE"
                     fi
                 done
                 if [ -z "$COMMANDARGS" ]; then
-                    echo "INVALID CMD FOUND" >> "$OUTPUTFILE"
+                    rm "$COMMANDTORUNFILE"
                     exit 111777 # Special exit code - arbitrarily picked for when the command is not expected
                 fi
-                OUT=$(sudo "${COMMANDARGS[@]}" 2> errFile)
+                OUT=$(sudo "${COMMANDARGS[@]}" 2> "$EXECFOLDER/stderr.txt")
                 STATUSCODE=$?
-                ERR=$(<errFile)
-                rm "errFile"
                 rm "$COMMANDTORUNFILE"
-                cat >| "$EXECFOLDER/stderr.txt" << EOF
-$ERR
-EOF
-                cat >| "$EXECFOLDER/stdout.txt" << EOF
-$OUT
-EOF
-                cat >| "$EXECFOLDER/status.txt" << EOF
-$STATUSCODE
-EOF
-                cat >| "$EXECFOLDER/output.json" << EOF
-" "
-EOF
+                $OUT | tee "$EXECFOLDER/stdout.txt"
+                $STATUSCODE | tee "$EXECFOLDER/status.txt"
+                " " | tee "$EXECFOLDER/output.json"
             fi
             if test -f "$OKSIGNALFILE"; then
-                echo "OK SIGNAL FILE FOUND" >> "$OUTPUTFILE"
                 rm "$OKSIGNALFILE"
             fi
-            echo "RELOOP" >> "$OUTPUTFILE"
+            sleep 5
         done
 done
--- a/vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
+++ b/vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
@@ -2,7 +2,7 @@
 EXECFOLDER=$1 # First argument is the working folder as this is launched with cwd of /root
 OKSIGNALFILE="$EXECFOLDER/ok.txt"
 COMMANDTORUNFILE="$EXECFOLDER/command.txt"
-OUTPUTFILE="/home/viru/om.txt"
+OUTPUTFILE="/home/test_output_.txt"
 end=$((SECONDS+3600))
 while true
 do
--- a/vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
+++ b/vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 OKSIGNALFILE=./ok.txt
 COMMANDTORUNFILE=./command.sh
 while true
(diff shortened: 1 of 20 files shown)


vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+EXECFOLDER=$1 # First argument is the working folder as this is launched with cwd of /root
+OKSIGNALFILE="$EXECFOLDER/ok.txt"
+COMMANDTORUNFILE="$EXECFOLDER/command.txt"
+OUTPUTFILE="/home/viru/om.txt"
```
nagilson, line 5:
```suggestion
OUTPUTFILE="/home/test_output_.txt"
```

vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
```diff
@@ -0,0 +1,36 @@
...
+        stop=false
+        while [ $SECONDS -lt $end ];
+        do
+            if test -f "$COMMANDTORUNFILE"; then
+                # echo "COMMAND FILE FOUND" >> "$OUTPUTFILE" # Leave this here as an example of debugging
+                COMMAND="$(cat "$COMMANDTORUNFILE" | awk '{$1=$1;print}')"
+                for validCmd in "${@:2}"
+                do
+                    if [ "$COMMAND" == "$validCmd" ]; then
+                        IFS=' ' read -ra COMMANDARGS <<< "$COMMAND"
```
nagilson, line 18: Eventually we should split the cmd file to be line by line instead of space separated, but it works for now because of the commands were running under sudo

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
```diff
@@ -488,6 +500,18 @@ export class DotnetGlobalVersionResolutionCompletionEvent extends DotnetCustomMe
     public readonly eventName = 'DotnetGlobalVersionResolutionCompletionEvent';
 }
 
+export class CommandProcessesExecutionFailureNonTerminal extends DotnetCustomMessageEvent {
```
nagilson, line 503: are we still using these?

nagilson: Yes!

nagilson:
We got the green light on this, please reach out to me for further explanation or comment on teams.

vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+EXECFOLDER=$1 # First argument is the working folder as this is launched with cwd of /root
+OKSIGNALFILE="$EXECFOLDER/ok.txt"
+COMMANDTORUNFILE="$EXECFOLDER/command.txt"
+OUTPUTFILE="/home/test_output_.txt"
```
baronfel, line 5:
```suggestion
#OUTPUTFILE="/home/test_output_.txt"
```

vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
baronfel:
```suggestion
                        # Eventually we should split the cmd file to be line by line instead of space separated,
                        # but it works for now because the commands are running under sudo
                        IFS=' ' read -ra COMMANDARGS <<< "$COMMAND"
```

vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
```diff
@@ -0,0 +1,36 @@
...
+                # echo "COMMAND FILE FOUND" >> "$OUTPUTFILE" # Leave this here as an example of debugging
+                COMMAND="$(cat "$COMMANDTORUNFILE" | awk '{$1=$1;print}')"
+                for validCmd in "${@:2}"
+                do
+                    if [ "$COMMAND" == "$validCmd" ]; then
+                        IFS=' ' read -ra COMMANDARGS <<< "$COMMAND"
+                    fi
+                done
+                if [ -z "$COMMANDARGS" ]; then
+                    rm "$COMMANDTORUNFILE"
```
baronfel, line 22: I see this cleanup-style stuff in multiple places, consider using a [bash 'exit trap' ](http://redsymbol.net/articles/bash-exit-traps/) as a way to consolidate all of the cleanup. This will also help if someone terminates the process outside of your control. This could cover removing the ok file and the command to run file at minimum.

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -84,62 +107,277 @@ Please install the .NET SDK manually by following https://learn.microsoft.com/en
...
-                if (stdout)
-                {
-                    this.context?.eventStream.post(new CommandExecutionStdOut(`The command ${fullCommandString} encountered stdout, continuing
+        // Launch the process under sudo
+        this.context?.eventStream.post(new CommandExecutionUserAskDialogueEvent(`Prompting user for command ${fullCommandString} under sudo.`));
+
+        // The '.' character is not allowed for sudo-prompt so we use 'NET'
+        let sanitizedCallerName = this.context?.acquisitionContext?.requestingExtensionId?.replace(/[^0-9a-z]/gi, ''); // Remove non-alphanumerics per OS requirements
+        sanitizedCallerName = sanitizedCallerName?.substring(0, 69); // 70 Characters is the maximum limit we can use for the prompt.
+        const options = { name: `${sanitizedCallerName ?? '.NET Install Tool'}` };
```
baronfel, line 143: Isn't the `.` in `.NET Install Tool` invalid per the comment above?

vscode-dotnet-runtime-library/src/Utils/IFileUtilities.ts
```diff
@@ -8,12 +8,12 @@ import { IEventStream } from '../EventStream/EventStream';
...
 export abstract class IFileUtilities
 {
-    public abstract writeFileOntoDisk(scriptContent: string, filePath: string, eventStream : IEventStream) : void;
+    public abstract writeFileOntoDisk(scriptContent: string, filePath: string, alreadyHoldingLock : boolean, eventStream? : IEventStream) : void;
 
     /**
      * @param directoryToWipe the directory to delete all of the files in if privilege to do so exists.
      */
-    public abstract wipeDirectory(directoryToWipe : string, eventStream : IEventStream) : void;
+    public abstract wipeDirectory(directoryToWipe : string, eventStream? : IEventStream, fileExtensionsToDelete? : string[]) : void;
```
baronfel, line 16: what led to the eventStreams being nullable here? We should have some stream the entire time for logging/telemetry purposes, right?

baronfel (approved):
A few notes about the communicator script, a few notes about things I saw in the utility files. Overall this looks solid to me. One thing I did note is that this is hard to test in an automated way - we probably need to have a few test plans that stress this communication specifically so that we can get some manual test coverage.

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
nagilson: Very good call, thank you.

JL03-Yue (approved):
It looks good to me🙂

nagilson:
One thing I dont like about this PR is it will cause a definite failure in the future whenever someone tries to install dotnet 8, unless we've added the packages. Maybe we should add the package names now, hell, even for 9 and 10, to prevent this from happening.

vscode-dotnet-runtime-library/src/Utils/IFileUtilities.ts
nagilson: The `CommandExecutor` `AcquisitionContext` with the `EventStream` may be `null` in rare scenarios be null because:

The `TelemetryUtilities` needs to call the `CommandExecutor` to edit the environment, and in creating the `Executor`, needs to pass its `Context` -> The Context of the `TelemetryUtilities` may be `null`, because it needs to be created before the global event stream can be created, so that the `TelemetryUtilities` can be added as a listener to that event stream -> The `context` of `TelemetryUtilities` containing the `EventStream` is added to it later after creation, so there is a point in time where it is null between when it is created, when the event stream is created and subscribed to it, and then when the event stream is added back in to the `TelemetryUtilities`. That point 'expanded' during the refactor, where almost every class needed to take a dependency on `AcquisitionContext` to correctly post an Error, though most classes already had it as one long ago.

The commands that do not create an `AcquisitionContext` may take actions that call the `TelemetryUtilities` before it has an `AcquisitionContext` with an `EventStream`, so theoretically some of that would not be recorded if they are called before an Acquisition based command is called.

It should not be this way, I agree. Fixing the Context class is non trivial but likely doable

nagilson: TLDR: If someone changes the VS Code telemetry setting before ever running an acquisition command, then that would not show up in our logs, nor would someone enabling telemetry come up in our telemetry.

vscode-dotnet-runtime-library/install scripts/interprocess-communicator.sh
nagilson: Oh, that's pretty handy. Thank you!

nagilson:
@baronfel I've responded to your feedback, thank you. Does this all look good to you?

vscode-dotnet-runtime-library/src/Utils/IFileUtilities.ts
baronfel: This makes perfect sense, thank you for walking through that!

baronfel:
> @baronfel I've responded to your feedback, thank you. Does this all look good to you?

Yes, let's :shipit:

PR #1641, opened 2024-01-30
Dont treat stderr as error for Install Scripts

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1635

Output in stderr from install scripts should not be treated as errors. Most of the code base follows that methodology, this code was never written to. Failures are still treated as such and fail with including the stderr.


JL03-Yue (approved):
Looks great to me.

nagilson:
thank you for your review :)

Issue #1658, opened 2024-02-06
[NETE2ESDK][Regression]There is no message saying that the .NET SDK is already installed if try installing it again.


NicoleWang001:
### Describe the bug.

There is no message saying that the .NET SDK is already installed if try installing it again.

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

1. Run the sample command ‘Install .NET SDK Globally via .NET Install Tool (Former Runtime Extension)‘.
2. Use the version it provided, which is 7.0.103. Hit enter.
3. After 7.0.103 is installed succeed, try to install 7.0.103 again and observe the output

### Expected Behavior

The extension should notice the SDK is already installed and give you a message saying it is already installed such as  ‘A global install is already on the machine’, then return success as normal.

### Exceptions & Errors (if any)

There is only a success message with the path to dotnet
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/f8ed44f7-3218-4f9c-98ec-f18f3d865830)

### .NET Install Tool Version

2.0.2

### Your operating system? Please provide more detail, such as the version and or distro above.

N/A

### VS Code Version

1.86.0

nagilson:
This is simply about fixing the messaging, it recognized the install was there and did not do any extra work, it just didnt communicate that.

NicoleWang001:
This issue is sill repro
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/ce6f60d4-c7cb-4627-a6de-589ac49fc29d)

nagilson:
This issue needs to be fixed eventually, but it is not a release blocker. We will need the logs to investigate (please send over the thread) as I am not getting this behavior.

v-xiaofchen:
On windows x64, can see a message saying the SDK is already installed but it seems install 7.0.103 again.
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/99376012/530bb09d-76ad-453b-ba25-2e24279cfd00)
On windows arm64. it the same behavior  as Linux: There is only a success message with the path to dotnet
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/99376012/8cd458d3-941f-44d1-8e02-3d71bbe3107e)

v-xiaofchen:
Run the command multiple times: .NET Install Tool - Install the .NET SDK System-Wide.
[...]

Issue #1657, opened 2024-02-06
[NETE2ESDK] [Linux]Cancelling .NET SDK installation does not work


NicoleWang001:
### Describe the bug.

Cancelling .NET SDK installation does not work on Linux

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

1. On Ubuntu 22.04 or RHEL 9.3, run the sample command ‘Install .NET SDK Globally via .NET Install Tool (Former Runtime Extension)‘.
2. Use the version it provided, which is 7.0.103. Hit enter.
3. On the Authentication Required prompt, click 'cancel'
4. Observe

### Expected Behavior

You should see a failure message saying the installation was cancelled and installation is cancelled successfully.

### Exceptions & Errors (if any)

On Ubuntu 22.04, no failure message and .NET SDK is installed successfully still.
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/5c0e9ea2-75c1-4347-885b-f02ed437c26b)

On RHEL, there is a failure message 'The user refused the password prompt', but .NET is installed successfully still.
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/28556f17-1db6-47be-86cf-2de7ad6f23da)

### .NET Install Tool Version

2.0.2

### Your operating system? Please provide more detail, such as the version and or distro above.

Linux

### VS Code Version

1.86.0

NicoleWang001:
With below steps, the installation is succeed still if canceling:
Install the .NET SDK -> Uninstall it -> try re-install it then cancel.
But if install the .NET SDK at the first time then cancel, the installation is failed with other errors.
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/099107aa-1eb8-4374-993b-6a6d1f950fc4)

nagilson:
Interesting, I think I know why this is happening. It works successfully even after cancellation because some parts of the process get left behind from the older run, so the elevation from the previous attempt is still working and accepting tasks. That is not a big problem in my eyes. We can probably update the testing doc to say this 'working after cancelling the password prompt' behavior is acceptable. Do you agree @baronfel?

What IS a problem is the last error with stdout. It's a complicated bug but I think a simple fix.

v-xiaofchen:
On ubuntu 22.04;Sometimes, the cancelling message is displayed as bellow, but the SDK is installed succeed.
[...]

PR #1660, opened 2024-02-06
Fix Bug where Password Prompt Failure to Launch could be considered Success

nagilson:
We did not consider failure to run apt-get update as a 'terminal failure', aka a failure that should cause the extension to fail the install request. But updating is often the first thing we do. So included in that request is the portion that launches the elevated master process. If that gets cancelled it will recognize it failed to launch but consider that non terminal so it will timeout. This should fix that.
 
This should be reviewed after https://github.com/dotnet/vscode-dotnet-runtime/pull/1641 is reviewed and merged, which is why it includes those changes.

This should resolve the bad part of https://github.com/dotnet/vscode-dotnet-runtime/issues/1657
Also adds a message to resolve https://github.com/dotnet/vscode-dotnet-runtime/issues/1658


vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -79,6 +80,8 @@ export class WinMacGlobalInstaller extends IGlobalInstaller {
                 if(conflictingVersion === this.installingVersion)
                 {
                     // The install already exists, we can just exit with Ok.
+                    this.acquisitionContext.eventStream.post(new DotnetAcquisitionAlreadyInstalled(`The .NET SDK ${this.installingVersion} is already installed.`,
```
nagilson, line 83:
```suggestion
                    this.acquisitionContext.eventStream.post(new DotnetAcquisitionAlreadyInstalled(` `,
```

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -79,6 +80,9 @@ export class WinMacGlobalInstaller extends IGlobalInstaller {
                 if(conflictingVersion === this.installingVersion)
                 {
                     // The install already exists, we can just exit with Ok.
+                    this.acquisitionContext.eventStream.post(new DotnetAcquisitionAlreadyInstalled(getInstallKeyFromContext(this.acquisitionContext.acquisitionContext) ?? '',
```
JL03-Yue, line 83: This provides very clear information about what to do if the version is installed. Does the user need to know that the version they required is not downloaded again?

JL03-Yue (approved):
Looks great to me 🙂

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
nagilson: That's a good point -- it is already there, just in the OutputChannelObserver! :)

```ts
                    this.outputChannel.append(`${
                        (event as DotnetAcquisitionAlreadyInstalled).requestingExtensionId
                    } wants to install .NET ${
                        (event as DotnetAcquisitionAlreadyInstalled).installKey
                    } but it already exists. No downloads or changes were made.\n`);
```

Issue #1661, opened 2024-02-08
[NETE2ESDK][RHEL].NET installation is failed with invalid version if using the command ‘Sample: Acquire .NET Runtime’ and select version ‘7.0.103’


NicoleWang001:
### Describe the bug.

On RHLE 9.3, .NET installation is failed with invalid version if using the command ‘Sample: Acquire .NET Runtime’ and select version ‘7.0.103’

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

1. On RHEL 9.3,
2. Run the command ‘Sample: Acquire .NET Runtime’ and select version ‘7.0.103’,
3. Observe.

### Expected Behavior

See that the install has a success message.

### Exceptions & Errors (if any)

.NET installation is failed with invalid version.
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/c2e0e8f1-3b91-4f49-bca6-f17072f77d89)

### .NET Install Tool Version

2.0.2

### Your operating system? Please provide more detail, such as the version and or distro above.

Linux

### VS Code Version

1.86.0

nagilson:
This should be fixed with the latest changes :)

PR #1668, opened 2024-02-14
Fix Version Parsing & Test Activation

JL03-Yue:
#1661 

Includes an additional fix as well for when we added on startupfinished activation, which is elaborated upon in the commit history,


nagilson:
@JL03-Yue  I think that other failure we were seeing in 'acquire' was due to a different problem, hopefully what I pushed just now will have fixed it though it may be a problem in another way, need to see how it works out.

Definitely go back to your original change and let's see if it works :)

nagilson:
Hm, my fix did not work. There is something going on here outside your change that is broken. Ill look into this.

nagilson:
Quick thought: I think this might be broken because of the way activation now works. If we activate the extension after startup, it could be that this was technically always broken because it tries to bundle the extension again when it runs the test, so it double registers things.

JL03-Yue:
@nagilson The current code passed all tests and passed the scenarios on Red Hat. I wonder if I should rebase my commits? It is failing on packaging and publishing, which I'm looking into[smile]

nagilson:
I think I fixed upstream.

nagilson (approved):

vscode-dotnet-runtime-library/src/Acquisition/VersionResolver.ts
```diff
@@ -159,7 +159,7 @@ export class VersionResolver implements IVersionResolver {
         }
         Debugging.log(`Semver parsing passed: ${version}.`, this.context.eventStream);
 
-        if (version.split('.').length !== 2 || !parsedVer)
+        if (version.split('.').length !== 2 && version.split('.').length !== 3 || !parsedVer)
```
nagilson, line 162:
```suggestion
        if (!parsedVer || (version.split('.').length !== 2 && version.split('.').length !== 3))
```

nagilson (approved):

PR #1679, opened 2024-02-27
Migrate to 1ES Pipeline

nagilson:
Per company policy, we must migrate to this new pipeline format soon.
For further information from reviewers, please contact me for an internal link to this migration effort.


nagilson:
ok cool github is bugged

1es-azure-pipeline.yml
```diff
@@ -0,0 +1,84 @@
+trigger:
+  batch: true
+  branches:
+    include:
+    - main
```
smitpatel, line 5: nit: Generally - feature/* is also good addition here

1es-azure-pipeline.yml
```diff
@@ -0,0 +1,84 @@
+trigger:
```
smitpatel, line 1: What is this pipeline used for?

Also good to have link to pipeline at top like this https://github.com/dotnet/project-system/blob/34eb57b35962367b71c2a1d79f6c486945586e24/eng/pipelines/pull-request.yml#L3-L4
@MiYanni started the trend.

azure-pipelines.yml
```diff
@@ -14,139 +14,48 @@ pr:
     - '*'
```
smitpatel, line 14: May be make file name singular.

azure-pipelines.yml
```diff
@@ -14,139 +14,48 @@ pr:
...
-    value: true
+- name: Codeql.Enabled
+  value: true
+
+parameters:
+- name: pools
+  type: object
+  default:
+  - name: NetCore-Public
+    vmImage: ubuntu-latest # The public PR pool schema is different from the 1ES one. We use 'vmImage' which is a default azure pipeline schema.
```
smitpatel, line 25: Comment on all of these seems different but not able to correlate with direct line. May be specify at the top together?

pipeline-templates/build-test.yaml
```diff
@@ -0,0 +1,38 @@
...
+- job: ${{ parameters.pool.os }}_Build
+  pool:
+    ${{ if eq(parameters.oes, 'true') }}:
+        ${{ parameters.pool }}
+    ${{ else }}:
+        vmImage: ${{ parameters.pool.vmImage }}
+  displayName: '🔨 ${{ parameters.pool.os }} Build and Test'
+  templateContext:
+    outputs:
+    - output: pipelineArtifact
```
smitpatel, line 14: 🎉

pipeline-templates/package-vsix.yaml
```diff
@@ -0,0 +1,56 @@
...
+  pool:
+    ${{ if eq(parameters.oes, 'true') }}:
+        ${{ parameters.pool }}
+    ${{ else }}:
+        vmImage: ${{ parameters.pool.vmImage }}
+  displayName: '📦 Package and Publish'
+  dependsOn:
+  - ${{ parameters.pool.os }}_Build
+  - TSLint
+  condition: succeeded()
```
smitpatel, line 15: nit: succeeded is default

smitpatel (approved):

azure-pipelines.yml
MiYanni, line 25: I'd be surprised if this is working as intending. Meaning, I don't believe `ubuntu-latest` exists in the `NetCore-Public` pool. It might be falling back to the Microsoft-hosted pool (`Azure Pipelines`), but the logs for the PR build would show that. If you take an image from [here](https://helix.dot.net/#1ESHostedPoolImagesWestUS-rg-Public), that should work for ya. I'd recommend: `build.ubuntu.2204.amd64.open`

azure-pipelines.yml
```diff
@@ -14,139 +14,48 @@ pr:
...
+    os: linux
+  - name: NetCore-Public
+    vmImage: macOS-latest # 1ES uses 'image' but defining an image this way using our public pools does not work as they aren't configured this way.
+    os: macOS
```
MiYanni, line 29: I'd just use the Microsoft-hosted one with certainty. Therefore, update `name` to be `Azure Pipelines`.

azure-pipelines.yml
```diff
@@ -14,139 +14,48 @@ pr:
...
+  - name: NetCore-Public
+    vmImage: windows-latest # We add the flag 'oes' or one engineering system: false or true in all tasks to determine if we use vmImage or a 1es pool object.
```
MiYanni, line 31: Same here for the `NetCore-Public` images for Windows. I'd recommend: `windows.vs2022preview.amd64.open`

azure-pipelines.yml
```diff
@@ -14,139 +14,48 @@ pr:
...
+    - template: pipeline-templates/build-test.yaml
+      parameters:
+        pool:
+          vmImage: ${{ image.vmImage }}
+          os: ${{ image.os }}
+        oes: false
+  - template: pipeline-templates/upstream-verify.yaml
+    parameters:
+      pool:
+        vmImage: windows-latest
```
MiYanni, line 47: Same for this and the other 2 jobs below. Use the `NetCore-Public` pool with an image from the list, such as `windows.vs2022preview.amd64.open`.

pipeline-templates/build-test.yaml
```diff
@@ -0,0 +1,38 @@
+parameters:
+  pool: ''
+
+jobs:
+- job: ${{ parameters.pool.os }}_Build
+  pool:
+    ${{ if eq(parameters.oes, 'true') }}:
+        ${{ parameters.pool }}
+    ${{ else }}:
+        vmImage: ${{ parameters.pool.vmImage }}
```
MiYanni, line 10: Oh, I see how to resolved the "you can't use the same templates" problem. That works just fine. I'd likely recommend using an actual parameter name instead of an acronym like `oes` which isn't very descriptive, and kinda looks like a typo for `os`.

MiYanni (approved):

.vscode/settings.json
```diff
@@ -13,5 +13,6 @@
         "programfiles",
         "REGHEXVALUE",
         "REGTYPE"
-    ]
+    ],
+    "azure-pipelines.1ESPipelineTemplatesSchemaFile": true
```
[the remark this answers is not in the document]
nagilson: Yes, this is to enable the extension to code complete 1es features.

1es-azure-pipeline.yml
```diff
@@ -0,0 +1,84 @@
...
+    type: git
+    name: 1ESPipelineTemplates/1ESPipelineTemplates
+    ref: refs/tags/release
+
+parameters:
+- name: pools
+  type: object
+  default:
+  - name: NetCore1ESPool-Internal
+    image: 1es-ubuntu-2204-pt
```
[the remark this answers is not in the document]
nagilson: Ah I did not see that, thank you!

pipeline-templates/build-test.yaml
nagilson: Changed to `useOneEngineeringPool`, good feedback.

azure-pipelines.yml
nagilson: Those pools dont seem to support this; hence the vm IMage. Name is not even used, its just there as a placeholder, which is kinda gross

1es-azure-pipeline.yml
nagilson: Thanks a convenient pattern, I like it. Thanks for leaving your feedback :)

Regarding this pipeline, it is the internal one which uses the 1es template, while the other one is for public pipelines not using 1es templates.

azure-pipelines.yml
nagilson: That's a good point, these comments are kind of in the middle of nowhere. Moving them.

pipeline-templates/package-vsix.yaml
nagilson: I'm not sure I trust the yaml team to not write a bug in the future that breaks this 😂 

I could not find any documentation from them stating that it is the intended behavior. https://learn.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml%2Cstages. So while I think this is good feedback to remove unnecessary code, I would prefer to keep it as a safety measure.

1es-azure-pipeline.yml
nagilson: I think this is a good idea, but we've never used a feature branch before in this repo, and Yanni is the first one to use it on the SDK repo. This is one interesting way to try to prevent a public PR that fails private tests from being merged, but TBH I don't think I would remember to check it. For now, I will leave this out, but if you have other reasons for doing this then let me know!

nagilson:
@smitpatel and @MiYanni Thank you both for taking the time to review :) That was rather speedy!

azure-pipelines.yml
nagilson: The name does not actually matter at all. But this is probably the most theoretically accurate. Except it might be misleading, maybe do something else like `Unused Pool Name`

nagilson: Itd be better to just get rid of the name but I dont know if we can actually do that or not.

nagilson:
ok I undid the rename so we can merge this as azure devops does not seem to let me select the new pipeline file until it is in main

azure-pipelines.yml
nagilson: Ok based on the docs Name is required. Super awkward that Name becomes the name of the pool using 1es format because that is not the intended use of Name in a non 1es format yaml file. The name is meant to be the name of a paramter, not the pool name, but since we use two templates with different meanings for the same thing, it's going to be weird either way. We could call it 'Fake 1ES Pool Object with Vm Image Windows'

nagilson:
Merging without some checks because GH seems to be waiting on the old checks that are in main, which no longer exist

PR #1713, opened 2024-03-19
Use official templates so we get counted on the internal pie chart

nagilson:
The migration tool and doc examples all use the unofficial one ecks dee

Also create a dupe of the PR pipeline so we can migrate main to that file once it is checked in and then delete the other one to rename it to a more descriptive name.

### DIFF
--- a/1es-azure-pipeline.yml
+++ b/1es-azure-pipeline.yml
@@ -47,7 +47,7 @@ parameters:
     os: windows
 
 extends:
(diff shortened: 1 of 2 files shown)


nagilson:
@MiYanni For me at least, the old pipeline stages are showing up in the CI on GH 
image

I tried to delete them and read this article https://stackoverflow.com/questions/57927115/delete-a-workflow-from-github-actions but this repo does not have github actions enabled. Any idea how to remove these old checks?

MiYanni (approved):

PR #1714, opened 2024-03-20
Make Recommended Version API Return Version instead of String

nagilson:
This new API we introduced in 2.0.2 will be changed to return IDotnetVersion instead of a string. This will be the only change in 2.0.3.

This will make calling the API simpler for external users. It's a breaking change, but we have not documented this and its relatively new, so should be OK.


vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -237,20 +237,21 @@ export function activate(context: vscode.ExtensionContext, extensionContext?: IE
...
-                return '';
+                return [];
             }
-            return availableVersions[0].version;
+            return [availableVersions[0]];
         }
 
         // The first item will be the newest version.
-        return activeSupportVersions[0].version;
+        return[ activeSupportVersions[0]];
```
nagilson, line 246:
```suggestion
        return [activeSupportVersions[0]];
```

JL03-Yue (approved):
Looks good to me.

Issue #1240, opened 2023-08-13
Set existingDotnetPath for all extensions?


sliekens:
Hello,

I'm currently working on enhancements to the Dev Container feature to install .NET inside a Dev Container (https://github.com/devcontainers/features/pull/628).

One of my objectives is to simplify the configuration process for using the pre-installed .NET Runtime across all extensions by configuring `dotnetAcquisitionExtension.existingDotnetPath`.

### Current configuration

Currently I have this in my `devcontainer-feature.json`:

``` json
{
  "customizations": {
    "vscode": {
      "settings": {
        "dotnetAcquisitionExtension.existingDotnetPath": [
          {
            "extensionId": "ms-dotnettools.csharp",
            "path": "/usr/share/dotnet/dotnet"
          },
          {
            "extensionId": "ms-dotnettools.csdevkit",
            "path": "/usr/share/dotnet/dotnet"
          }
        ]
      }
    }
  }
}
```

I have identified a few problems with this configuration:

1. I need to know ahead of time which extensions will require .NET and add them to the array in `devcontainer-feature.json`
2. If a user wants to add an extension to the configuration in their own `devcontainer.json`, it will overwrite the existing configuration from `devcontainer-feature.json`. This leads to complete replacement, rather than merging.

### Ideal configuration approach

I would like a simplified configuration for both current and future extensions. Here are the proposed alternatives, any of which would help me achieve my goal:

1. Simplified global configuration

``` json
{
    "dotnetAcquisitionExtension.existingDotnetPath": "/usr/share/dotnet/dotnet"
}
```

2. Using a wildcard to keep the configuration schema intact

``` json
{
  "dotnetAcquisitionExtension.existingDotnetPath": [
    {
      "extensionId": "*",
      "path": "/usr/share/dotnet/dotnet"
    }
  ]
}
```

3. Support object syntax to facilitate configuration merging

``` json

{
  "dotnetAcquisitionExtension.existingDotnetPath": {
    "ms-dotnettools.csharp": "/usr/share/dotnet/dotnet",
    "ms-dotnettools.csdevkit": "/usr/share/dotnet/dotnet"
  }
}
```
Thanks for your time and consideration.

baronfel:
This is a useful idea, but I think it's almost entirely unnecessary with the next major version of this extension, which optionally allows for using the standard globally-installed .NET SDK/Runtime instead of per-extension.

sliekens:
[...]

PR #1726, opened 2024-03-27
Add global shared file path for extension

JL03-Yue:
Add global shared file path setting (`sharedExistingDotnetPath`) required in #1240

This PR adds a global shared file path for extension apart from the local extension path. Specifically, it created a data interface `IExistingPaths` that contains both a list of local paths and a shared path, adding corresponding tests. When resolving paths in `ExistingPathResolver`, it returns the local path if the extension id is specified, return the global shared path if no extension id is specified and the global shared path is specified, and gives a warning if no path is specified. Corresponding tests are added to validate the behavior. 

# What is this setting:

Users can set their own install of .NET as the path to use for C# DevKit, C#, etc. In the past they had to set the setting for every single extension. This setting will apply the existing dotnet path as a 'shared' existing path so all extensions that call our extension will get the path.

The individual settings will take precedence over the shared/'global' path setting. For more info, see the readme edits.


vscode-dotnet-runtime-library/src/Acquisition/ExistingPathResolver.ts
```diff
@@ -5,21 +5,28 @@
 
 import { IWindowDisplayWorker } from '../EventStream/IWindowDisplayWorker';
 import { IDotnetAcquireResult } from '../IDotnetAcquireResult';
-import { IExistingPath } from '../IExtensionContext';
+import { IExistingPaths, ILocalExistingPath } from '../IExtensionContext';
```
nagilson, line 8: nit: unused import ILocalExistingPath

vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
```diff
@@ -126,17 +125,15 @@ export async function callWithErrorHandling<T>(callback: () => T, context: IIssu
 
 async function configureManualInstall(context: IIssueContext, requestingExtensionId: string): Promise<void> {
     const manualPath = await context.displayWorker.displayPathConfigPopUp();
-    if (manualPath && fs.existsSync(manualPath))
-    {
-        try
-        {
-            let configVal: IExistingPath[] = [{ [ExistingPathKeys.extensionIdKey]: requestingExtensionId, [ExistingPathKeys.pathKey] : manualPath}];
+
+    if (manualPath && fs.existsSync(manualPath)) {
```
nagilson, line 129:
```suggestion
    if (manualPath && fs.existsSync(manualPath))
    {
```

nit: preserve the style choices of the original document (same for the try catch block)

vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
```diff
@@ -126,17 +125,15 @@ export async function callWithErrorHandling<T>(callback: () => T, context: IIssu
...
     const manualPath = await context.displayWorker.displayPathConfigPopUp();
-    if (manualPath && fs.existsSync(manualPath))
-    {
-        try
-        {
-            let configVal: IExistingPath[] = [{ [ExistingPathKeys.extensionIdKey]: requestingExtensionId, [ExistingPathKeys.pathKey] : manualPath}];
+
+    if (manualPath && fs.existsSync(manualPath)) {
+        try {
+            let configVal: ILocalExistingPath[] = [{ [ExistingPathKeys.extensionIdKey]: requestingExtensionId, [ExistingPathKeys.pathKey] : manualPath}];
```
nagilson, line 131: This was not in the original spec, so I apologize, but I think it would be better here to set the global path here. We failed to install on the users machine which would likely be extension independent, at least more likely than not. So we'd rather them configure it for all extensions in 1 go instead of for just 1 and have to redo this process.
```suggestion
            let configVal: ILocalExistingPath[] = [{ [ExistingPathKeys.extensionIdKey]: requestingExtensionId, [ExistingPathKeys.pathKey] : manualPath}];
```

nagilson (commented):
>  

We should also update the read.me page to show how to set the global path. The read.me page is the main page on the extension in the vs code marketplace; we already have a section for setting the path, but we should demonstrate how to do so globally.

vscode-dotnet-runtime-library/src/IExtensionContext.ts
```diff
@@ -21,7 +21,12 @@ export namespace ExistingPathKeys {
...
 
-export interface IExistingPath {
+export interface ILocalExistingPath {
     [ExistingPathKeys.extensionIdKey]: string;
     [ExistingPathKeys.pathKey]: string;
 }
+
+export interface IExistingPaths{
+    localExsitingPaths: ILocalExistingPath[];
+    globalExistingPathKey? : string;
```
nagilson, line 31: Thank you for using nullability ;)

vscode-dotnet-sdk-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -54,6 +57,9 @@ suite('DotnetCoreAcquisitionExtension End to End', function()
   const mockDisplayWorker = new MockWindowDisplayWorker();
   const environmentVariableCollection = new MockEnvironmentVariableCollection();
   let extensionContext: vscode.ExtensionContext;
+  const mockExistingPaths: IExistingPaths = {
+    localExsitingPaths: [{extensionId: 'alternative.extension', path: 'foo'}]
```
nagilson, line 61: We should ad some more test data here, by adding a global path setting, and several mock extension ids, which:

1 does have its own local setting - make sure it takes precedence over the global one
and another which does not set the path, (not sure if this is correct and will work or not) and will either pick up the global one or fail. 

unfortunately there is no easy way to test without adding this weird test case behavior

vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -42,6 +43,15 @@ suite('DotnetCoreAcquisitionExtension End to End', function() {
   const mockDisplayWorker = new MockWindowDisplayWorker();
   let extensionContext: vscode.ExtensionContext;
 
+  const mockExistingPaths: IExistingPaths = {
```
nagilson, line 46: nit: delete this unused var

vscode-dotnet-sdk-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
nagilson: Oh whoops this was done, sorry! It's in the runtime extension and we dont need to test the sdk one so much ;)

vscode-dotnet-runtime-library/src/Utils/ExtensionConfigurationWorker.ts
```diff
@@ -2,21 +2,28 @@
...
     constructor(private readonly extensionConfiguration: IExtensionConfiguration,
                 private readonly pathConfigValueName: string | undefined) {}
 
-    public getPathConfigurationValue(): IExistingPath[] | undefined {
+    public getPathConfigurationValue(): IExistingPaths | undefined {
         return this.pathConfigValueName ? this.extensionConfiguration.get(this.pathConfigValueName) : undefined;
     }
 
-    public async setPathConfigurationValue(configValue: IExistingPath[]): Promise<void> {
+    public async setLocalPathConfigurationValue(configValue: ILocalExistingPath[]): Promise<void> {
```
nagilson, line 16: This code looks great, but I am concerned about whether VS Code will fail or not. The configuration is updated using the same key whether its local or global, so if the settings change b2b, then we could set the configuration to a string type, and later try to put a ILocalExistingPath[] obj in it. I'm not sure if VS Code will fail or how it will handle that, so we should either test it out, look at the vs code code, or use a dseparate configValueName for both global and local path

nagilson (commented):
Nice work independently looking through all of this code and making the updates! :smile: The overall changes are great, but there are some bugs or other improvements I'd like to discuss first which I've denoted above :)

vscode-dotnet-runtime-library/src/Utils/ExtensionConfigurationWorker.ts
JL03-Yue: Thanks for the explanation. The `setGlobalPathConfigurationValue` is got rid of for now.

vscode-dotnet-runtime-library/src/Utils/ExtensionConfigurationWorker.ts
```diff
@@ -2,21 +2,28 @@
...
         if (!this.pathConfigValueName) {
             throw Error('Existing path configuration not supported.');
         }
-        await this.extensionConfiguration.update<IExistingPath[]>(this.pathConfigValueName, configValue, true);
+        await this.extensionConfiguration.update<ILocalExistingPath[]>(this.pathConfigValueName, configValue, true);
+    }
+
+    public async setGlobalPathConfigurationValue(configValue: string): Promise<void> {
+        if (!this.pathConfigValueName) {
```
nagilson, line 24: `pathConfigValueName` is storing the setting for local paths. If we use that here, it will override the local path setting. We should use the string `sharedExistingDotnetPath` instead which would be stored as the globalPathKey

nagilson (commented):
The new logic in the path resolver looks good! :) I think we'll need some more changes to allow users to set the setting though. I've left some comments.

nagilson (commented):
We should also update the read.me page for the extension as mentioned so users know this setting exists.

My typo correction extension fixed a typo for me, so I pushed that up since it was already done. 😉 But it might need to be fixed in one more place per the build (sorry!).

JL03-Yue:
Thank you very much for the review! 🙂 I updated the `extensionConfigurationWorker` and the interfaces in `IExtensionContext`. The feature is also manually tested in the runtime acquisition extension. The `README` is also updated.

nagilson:
Thank you @JL03-Yue ! :) 

The changes you made looked good! I did some minor text updates and refactoring, (my initial hope was to help get this merged by standup), and I think with that, the PR is complete. 🥳

nagilson (approved):

vscode-dotnet-runtime-extension/package.json
```diff
@@ -73,7 +73,16 @@
...
-					"description": "File Path to an existing installation of .NET. Used for both the .NET SDK and .NET Runtime.",
+					"description": "File Path to an existing installation of .NET to be used by one extension. Used for both the .NET SDK and .NET Runtime. Restart VS Code to apply changes.",
+					"examples": [
+						"C:\\Program Files\\dotnet\\dotnet.exe",
+						"/usr/local/share/dotnet/dotnet",
+						"/usr/lib/dotnet/dotnet"
+					]
+				},
+				"dotnetAcquisitionExtension.sharedExistingDotnetPath": {
+					"type": "string",
```
tmeschter, line 84: What's the difference between this and the existing `dotnetAcquisitionExtension.existingDotnetPath` setting?

tmeschter:
Is the idea that the specified runtime would be used instead of automatically downloading and installing one?

vscode-dotnet-runtime-extension/package.json
nagilson: `existingDotnetPath` is a setting that has existed for about 5 years. With that setting, you need to specify a path and an extension id. The path is only used for that extension. If you want MAUI, C#, C#DK to all use the same path, you need to set it for all 3 extension ids individually. Many users dont realize this.

`sharedExistingDotnetPath` does not require the extension id. It's only a path string, and it will apply to all extensions at once that call us.

nagilson:
> Is the idea that the specified runtime would be used instead of automatically downloading and installing one?

Yes, although the calling extension still needs to call us. And instead of downloading a runtime and returning the path to it, we just return the preexisting path. This has already existed, just not with the above caveat. 

For the record, C#DK has its own existing path check before they call us, which just looks at the PATH. This has caused confusion. But that was a problem before this PR, and would need to be resolved there.

tmeschter (approved):

PR #1730, opened 2024-03-28
Focus on Output Channel to Indicate Install Progress for Global SDK

nagilson:
Currently in DevKit when people try to install the SDK they have no indication of progress unless they manually find our tab in the output window, which no one will do. This requires user intervention to occur, so when they click download, we'd like to navigate them to our output window, and it should not be so disruptive. 

We cant do this in DevKit since VS Code does not have an API to call into others output channels.


JL03-Yue (approved):
Looks good to me

PR #1745, opened 2024-04-10
Collect stack trace on axios error to diagnose better

nagilson:
This is a very simple change to allow our logs to get an error stack trace when axios fails. We have seen a lot of n r split is not a function and I believe this is a type error in either the axios or axios cache interceptor library. I cant tell where it is until we can get a repro with this info.

Manually tested it and it works.


JL03-Yue (approved):
Looks good to me

PR #1748, opened 2024-04-12
Track which Extensions Own Which Installs

nagilson:
# Summary
The initial design years ago did not keep track of which extensions installed which versions of .NET, and it has not been updated since. This makes it hard for us to responsibly uninstall .NET without breaking other extensions.

This PR is meant to add that tracking information. We will change the extension state keys that used to store strings into an `InstallRecord` object, which holds more info about the install. 

We also used to store the extension install data with a 'Install Key' which was just a string which held all of the information. This is a volatile pattern. Instead, we introduce `dotnetInstall` as a type to hold the install data. We introduce the `installMode` concept as well to allow for considering `aspnet`, `windowsdesktop`, and other installs and not just `runtime` or `sdk` installs.

# Assumptions in the change

This is a bit tricky to do because there will be pre-existing keys of just string type (for old versions, just a string of the version, for newer ones, a string with -global and the architecture) inside the extension state for most users. 

I went with a philosophy of 'protected access' -- I.E. work under the assumption that there are no longer any 'string' types stored in the extension by ensuring that any time we access the extension state, we update it before allowing any use of it. That is the job of the `InstallerTracker` -- of which much of the code is ported from the `AcquisitionWorker` class.

# More info on the old concept of `InstallKey`

The 'InstallKey' is meant to represent a single install of .NET and used to just be a 'version' string. Then we introduced '-global' and the architecture to it. I also updated this to become a type, `DotnetInstall` which contains this information to 
a - get us out of the business of string parsing
b - allows us to auto update an install in the future, so the path of the install is not directly correlated to the version.

# More Changes

This also includes some bug fixes I noticed in the code, for example:

Before when we found a 'partial install' we would delete ALL local runtimes. I am astounded this exists. There are other extensions that might depend on those other runtimes and this just deletes it from underneath them. A partial install is rare, I think you need to pkill vscode or something since the install script runs out of proc. We consider an install partial if our code never got to mark it as installed after calling the install script. Its rare but does happen, at 2010 occurrences out of 470k in recent days.


vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -69,7 +71,7 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
     constructor(protected readonly context: IAcquisitionWorkerContext, private readonly utilityContext : IUtilityContext, extensionContext : IVSCodeExtensionContext) {
         const dotnetExtension = os.platform() === 'win32' ? '.exe' : '';
         this.dotnetExecutable = `dotnet${dotnetExtension}`;
-        this.acquisitionPromises = {};
+        this.acquisitionPromises = new InProgressInstallManager();
```
baronfel, line 74: I was thinking about the lifecycle of this component - I almost think it should be created when the extension initializes, and upon creation this manager should 'clean' the data in the cache, bringing it up to the current data format. this way we can ensure that all of the data is in the new format (almost like a database migration) and that the rest of the codebase doesn't have to think about the different formats.

baronfel (commented):
Left a few comments about the approach, but I like a lot of what I see here. I like extracting the installation tracking to a separate component, but have a few questions about the data structure and how it impacts telemetry, as well as how/when we should migrate old data.

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -59,6 +59,7 @@ import {
 } from 'vscode-dotnet-runtime-library';
 import { dotnetCoreAcquisitionExtensionId } from './DotnetCoreAcquisitionId';
 import { IAcquisitionWorkerContext } from 'vscode-dotnet-runtime-library/dist/Acquisition/IAcquisitionWorkerContext';
+import { DotnetInstall, GetDotnetInstallInfo } from 'vscode-dotnet-runtime-library/dist/Acquisition/IInstallationRecord';
```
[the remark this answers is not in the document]
nagilson: Ah, thank you for catching that. That was copilot. It was smart enough to recognize the types werent exported yet but not smart enough to make the change in the index.ts file to export them and then import them like the rest...

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
```diff
@@ -20,14 +21,16 @@ export class DotnetAcquisitionStarted extends IEvent {
...
 
-    constructor(public readonly installKey: string, public readonly startingVersion: string, public readonly requestingExtensionId = '') {
+    constructor(public readonly installKey: DotnetInstall, public readonly startingVersion: string, public readonly requestingExtensionId = '') {
         super();
     }
 
     public getProperties() {
-        return {AcquisitionInstallKey : this.installKey,
+        return {
+                ...InstallToStrings(this.installKey),
```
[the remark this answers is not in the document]
nagilson: Yes, it will. With that said, it will actually standardize the events a bit more and make it easier to query them. Unfortunately, it will mean the new data is not the same as the old data.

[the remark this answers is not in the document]
nagilson: Actually we can probably keep the old data format to prevent churn on the old data :thinking:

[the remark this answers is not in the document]
nagilson: Thank you for pointing this out :) To make our lives easier I decided to add back in and keep the pre-existing properties.

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
nagilson: I think that's a wonderful idea! I've converted the class into a singleton, which I believe is the clearest way to do this pattern.

vscode-dotnet-runtime-library/src/Acquisition/IInstallationRecord.ts
```diff
@@ -0,0 +1,222 @@
...
+ * @remarks
+ * A string representing the installation of either a .NET runtime or .NET SDK.
+ */
+export interface DotnetInstall
+{
+    installKey: string;
+    version: string;
+    architecture: string;
+    isGlobal: boolean;
+    isRuntime: boolean;
```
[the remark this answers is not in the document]
nagilson: I've changed this, ironically I ended up calling the component the mode because 'type' was used for global and local in the previous interface.

vscode-dotnet-runtime-library/src/Utils/InstallKeyGenerator.ts
```diff
@@ -4,15 +4,24 @@
...
 
-    return DotnetCoreAcquisitionWorker.getInstallKeyCustomArchitecture(ctx.version, ctx.architecture,
-        ctx.installType ? ctx.installType === 'global' : false);
+    return {
+        installKey : DotnetCoreAcquisitionWorker.getInstallKeyCustomArchitecture(ctx.version, ctx.architecture,
+            ctx.installType ? ctx.installType === 'global' : false),
+        version: ctx.version,
+        architecture: ctx.architecture,
+        isGlobal: ctx.installType ? ctx.installType === 'global' : false,
+        isRuntime: looksLikeRuntimeVersion(ctx.version)
```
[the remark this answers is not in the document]
nagilson: That's a great point, I've updated this so that we no longer need to do this guesswork. Thank you :) That could have resulted in some nasty bugs

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
nagilson: Hm, I cant actually think of a good way to do this. The class needs the context to access the extension id of the current requesting call and for other things. We could give the context in every single function call to the object but that does not necessarily seem cleaner to me. The data in the extension state is still stored in a single place and the lock that is held to access it is shared even amongst multiple instances of this class, so both of the concerns we have *a) not thinking about different formats and *b) protecting against bad shared state is good.

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -239,94 +254,109 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
...
-            this.context.installationValidator.validateDotnetInstall(installKey, dotnetPath);
-            this.context.eventStream.post(new DotnetAcquisitionAlreadyInstalled(installKey,
+            // We don't do this check with global acquisition, since external sources can more easily tamper with installs.
+            this.context.installationValidator.validateDotnetInstall(install, dotnetPath);
+
+            this.context.eventStream.post(new DotnetAcquisitionAlreadyInstalled(install,
                 (this.context.acquisitionContext && this.context.acquisitionContext.requestingExtensionId)
                 ? this.context.acquisitionContext!.requestingExtensionId : null));
+
+            await this.installTracker.trackInstalledVersion(install);
```
nagilson, line 285: The fact that this line wasnt here before was a bug.

nagilson (commented):
I've looked through everything and it looks good to me now and ready for another review, barring one test fix I need to push.

vscode-dotnet-runtime-library/src/Acquisition/InstallTracker.ts
```diff
@@ -0,0 +1,302 @@
...
+        fs.writeFileSync(lockPath, '', 'utf-8');
+
+        this.context.eventStream?.post(new DotnetLockAttemptingAcquireEvent(`Lock Acquisition request to begin.`, new Date().toISOString(), lockPath, lockPath));
+        try
+        {
+            if(alreadyHoldingLock)
+            {
+                return await f(...(args));
+            }
+            const release = await lockfile.lock(lockPath, { retries: { retries: 10, maxTimeout: 1000 } });
```
nagilson, line 65: We might want to use the user timeout setting here instead of 1000.

vscode-dotnet-runtime-library/src/Acquisition/InstallTracker.ts
```diff
@@ -0,0 +1,302 @@
...
+    public constructor(protected readonly context : IAcquisitionWorkerContext)
+    {
+
+    }
+
+    protected executeWithLock = async <A extends any[], R>(alreadyHoldingLock : boolean, f: (...args: A) => R, ...args: A): Promise<R> =>
+    {
+
+        const trackingLock = 'tracking.lock';
+        const lockPath = path.join(__dirname, trackingLock);
```
baronfel, line 55: I get a little weirded out when writing files to the extension install directory - should we be using the state directory here instead?

baronfel (approved):
This was a big set of changes 😓 I think I agree with many of the main decisions here, and it definitely helps that there are dedicated tests to use as starting points for reviewing this!

vscode-dotnet-runtime-library/src/Acquisition/InstallTracker.ts
nagilson: Thank you for taking a look at the code. Sorry it is such a big change!

I'm not sure what you mean by the state directory? (If you mean the extension state, this needs to be an actual file on disk.)

I totally understand your hesitation, with webpack and such this directory can change, and it is volatile. This file is only a temporary one used by the local code however, and `__dirname` is the safest way I know of to get a non-hardcoded directory that we manage.

nagilson: I'm going to get this merged for now since it looks like you approved the broad swath of changes, but we can definitely keep following up on this one!

LLotme (commented):
gh pr checkout 1748

PR #1758, opened 2024-04-15
Update Is Online Lookup

nagilson:
We rely on a library is-online that has not been managed in about 2 years. The check that it does seems to rely on apples hotspot website which is kind of rudimentary https://github.com/sindresorhus/is-online/blob/main/index.js. We can instead try to rely on 8.8.8.8 or the google dns server, which is one of the highest up time servers. Note that we are partially doing this to update the dependencies, and is-online has a breaking change which also makes it not ideal.

This is a crude check but it only happens when we've already timed out and is simply there to improve error diagnosability. I did test pulling the plug on the internet a few times with this one.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,20 +8,10 @@
 			"license": "MIT",
 			"dependencies": {
 				"rimraf": "^5.0.1",
-				"vsce": "2.11.0"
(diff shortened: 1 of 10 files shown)


JL03-Yue (approved):
Looks good to me.

PR #1768, opened 2024-04-24
Protect against no versions of .NET to recommend 

nagilson:
The API contract for returning a `recommendedVersion` is to return an `IDotnetListVersionsResult` which is an array of versions. If there are no packages of .NET available to install, we should have a specific message for that and return an empty array instead of what we do right now, which is:

a - returning a bad version
b - returning undefined
c - erroring out

I would ignore the earlier commits in the history. GitHub seems to be buggy today.


JoeRobich (approved):
LGTM

nagilson:
Thank you! :)

vscode-dotnet-runtime-library/src/Acquisition/GenericDistroSDKProvider.ts
```diff
@@ -209,6 +210,14 @@ export class GenericDistroSDKProvider extends IDistroDotnetSDKProvider
             }
         }
 
+        if(maxVersion === '0')
+        {
+            const err = new DotnetVersionResolutionError(new Error(`No packages for .NET are available.
+Please refer to https://learn.microsoft.com/en-us/dotnet/core/install/linux if you'd link to install .NET.`), null);
```
tmeschter, line 216: Typo: I think you meant "like" instead of "link".

PR #1770, opened 2024-04-24
Improve the error message for failing to install the .NET SDK.

nagilson:
This error message is misleading. We could detect the distro, it just wasn't supported yet in our list of supported distros. We should give the user more action here.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/IDistroDotnetSDKProvider.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/IDistroDotnetSDKProvider.ts
@@ -82,7 +82,7 @@ export abstract class IDistroDotnetSDKProvider {
         if(!distroVersion || !this.distroJson || !((this.distroJson as any)[this.distroVersion.distro]))
         {
             const error = new DotnetAcquisitionDistroUnknownError(new Error(`Automated installation for the distro ${this.distroVersion.distro} is not yet supported.
-Please install the .NET SDK manually: dotnet.microsoft.com/download.`),
+Please install the .NET SDK manually: https://dotnet.microsoft.com/download`),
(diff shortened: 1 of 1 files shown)


vscode-dotnet-runtime-library/src/Acquisition/IDistroDotnetSDKProvider.ts
```diff
@@ -81,7 +81,8 @@ export abstract class IDistroDotnetSDKProvider {
         this.distroJson = JSON.parse(fs.readFileSync(distroDataFile, 'utf8'));
         if(!distroVersion || !this.distroJson || !((this.distroJson as any)[this.distroVersion.distro]))
         {
-            const error = new DotnetAcquisitionDistroUnknownError(new Error('We are unable to detect the distro or version of your machine'),
+            const error = new DotnetAcquisitionDistroUnknownError(new Error(`Automated installation for the distro ${this.distroVersion.distro} is not yet supported.
+Please install the .NET SDK manually: dotnet.microsoft.com/download.`),
```
MiYanni, line 85: I think you should leave off the `.` on the end so copying the download link is easy. Also, the link should be a full link, meaning `https://dotnet.microsoft.com/download`.

MiYanni (approved):

vscode-dotnet-runtime-library/src/Acquisition/IDistroDotnetSDKProvider.ts
nagilson: Good point, thank you

nagilson, line 85:
```suggestion
Please install the .NET SDK manually: https://dotnet.microsoft.com/download`),
```

PR #1801, opened 2024-05-14
Update distro schema file to account for changes in the file.

baronfel:
I was going to prospectively add support for Arch and found that 

a) the schema file wasn't working for me, and
b) once I got it working, it was out of date with the actual contents of the file due to some enhancements made recently.

So I updated the schema file to account for the primary change, which was making sure the preInstallCommands followed the same command structure as all the other commands.

I also used the ability of JSON Schema to name and reuse definitions to extract out the 2 most common types in use in this file:

* the description of a command
* the description of the package names to use for each component of the SDK/Runtime(s)

This makes it much easier to reuse the content.

### DIFF
--- a/Documentation/json-schema/distro-support-schema.json
+++ b/Documentation/json-schema/distro-support-schema.json
@@ -2,51 +2,127 @@
     "$schema": "http://json-schema.org/draft-04/schema#",
     "type": "object",
     "properties": {
-      "DISTRO": {
+        "$schema": {
+            "type": "string"
+        }
+    },
+    "$defs": {
+        "executable_command": {
+            "$schema": "http://json-schema.org/draft-04/schema#",
+            "type": "object",
+            "properties": {
+                "runUnderSudo": {
+                    "type": "boolean",
+                    "description": "true if this command should be executed under sudo on linux"
+                },
+                "commandRoot": {
+                    "type": "string",
+                    "description": "The first word in the command, e.g. if apt-get install foo, it would be the string apt."
+                },
+                "commandParts": {
+                    "type": "array",
+                    "description": "The remaining parts of the command in an array, for apt-get install {0} it would be [install, {0}] as strings"
+                }
+            },
+            "additionalProperties": false
+        },
+        "required_packages": {
+            "$schema": "http://json-schema.org/draft-04/schema#",
+            "type": "object",
+            "properties": {
+                "version": {
+                    "type": "string",
+                    "description": "The version major.minor of dotnet"
+                },
+                "sdk": {
+                    "type": "array",
+                    "items": [
+                        {
+                            "type": "string",
+                            "description": "a collection of all potential package names for this version of the dotnet sdk on this distro"
+                        }
+                    ]
+                },
+                "runtime": {
+                    "type": "array",
+                    "items": [
+                        {
+                            "type": "string",
+                            "description": "a collection of all potential package names for this version of the dotnet runtime on this distro"
+                        }
+                    ]
+                },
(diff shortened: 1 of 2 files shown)


Documentation/json-schema/distro-support-schema.json
```diff
@@ -2,51 +2,127 @@
...
+                "items": {
+                    "$ref": "#/$defs/executable_command"
+                },
+                "description": "The command(s) needed that will return true if a given package is installed on the machine"
+            },
+            "readSymLinkCommand": {
+                "type": "array",
+                "items": {
+                    "$ref": "#/$defs/executable_command"
+                },
```
baronfel, line 125: This is a new kind of command that was in the actual file but not reflected in the schema.

Documentation/json-schema/distro-support-schema.json
```diff
@@ -2,51 +2,127 @@
     "$schema": "http://json-schema.org/draft-04/schema#",
     "type": "object",
     "properties": {
-      "DISTRO": {
+        "$schema": {
+            "type": "string"
+        }
+    },
+    "$defs": {
```
baronfel, line 9: named groups can be defined under `$defs` and then reused

Documentation/json-schema/distro-support-schema.json
```diff
@@ -2,51 +2,127 @@
     "$schema": "http://json-schema.org/draft-04/schema#",
     "type": "object",
     "properties": {
-      "DISTRO": {
+        "$schema": {
```
baronfel, line 5: I had to describe the schema field because I turned on turbo-enforcement of the schema using `additionalProperties: false` everywhere.

Documentation/json-schema/distro-support-schema.json
```diff
@@ -2,51 +2,127 @@
...
+                "commandRoot": {
+                    "type": "string",
+                    "description": "The first word in the command, e.g. if apt-get install foo, it would be the string apt."
+                },
+                "commandParts": {
+                    "type": "array",
+                    "description": "The remaining parts of the command in an array, for apt-get install {0} it would be [install, {0}] as strings"
+                }
+            },
+            "additionalProperties": false
```
baronfel, line 27: this is a key change: each structure added this directive to it to force the validation to reject any additional properties found in the structures in the distro support file. this means we must exactly describe the structure here in order to validate.

MiYanni (approved):
I didn't see anything that looked out-of-sorts. 👍

nagilson:
Thanks for updating this, I didn't know about these formatting conventions for json schema and how you could define a type (hence why most of them dont use the `CommandExecutorCommand` type.) Is there a tool you're using for this that makes all of this useful?

baronfel:
Nope, just my past experiences :) a lot of this is shared from hand-writing Swagger schemas about 10 years ago, so the idea of generating and reusing the shapes is very natural to me.

Documentation/json-schema/distro-support-schema.json
nagilson: This is super useful and safeguards us for future contributor edits. Thank you a ton Chet, this is pretty awesome and I never would have thought of it!

Documentation/json-schema/distro-support-schema.json
nagilson: Somehow I forgot to add this one. Good catch!

nagilson (approved):
This looks great to me. Thank you for introducing me to the schema validator that's build into vscode and updating these files, I appreciate that a lot. This is pretty rad!

Issue #1734, opened 2024-03-30
vscode for mac os sonoma silicon m1 and Avalonia UI project folder - Error Failed to download .NET 8.0:


aspepper:
### Describe the bug.

I hope everyone is okay.

When opening VSCode, with the folder of an Avalonia UI Coss Platform project, vscode tries to install net 8, although the net 8 sdk is already installed on my mac. So I always get the error message below:

AvaloniaTeam.vscode-avalonia requested to download the .NET Runtime.
Downloading .NET version(s) 8.0.3~arm64 ....Error
Failed to download .NET 8.0:
Web Request to https://dot.net/v1/dotnet-install.sh Failed: n[r].split is not a function. Abortion. Please ensure that you are online.

Error
Failed to download .NET 8.0:
Web Request to https://dot.net/v1/dotnet-install.sh Failed: n[r].split is not a function. Abortion. Please ensure that you are online.

Error
Failed to download .NET 8.0.3~arm64:
Command failed: "/Users/username/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.0.3/dist/install scripts/dotnet-install.sh" -InstallDir "/Users/username/Library/Application Support/Code/User/globalStorage/ms-dotnettools.vscode-dotnet-runtime/.dotnet/8.0.3~arm64" -Version 8.0.3 -Verbose -Runtime dotnet -Architecture arm64
/bin/sh: /Users/username/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.0.3/dist/install scripts/dotnet-install.sh: Permission denied

Error
Failed to download .NET 8.0:
.NET Acquisition Failed: Installation failed: Error: Command failed: "/Users/username/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.0.3/dist/install scripts/dotnet-install.sh" - InstallDir "/Users/username/Library/Application Support/Code/User/globalStorage/ms-dotnettools.vscode-dotnet-runtime/.dotnet/8.0.3~arm64" -Version 8.0.3 -Verbose -Runtime dotnet -Architecture arm64
/bin/sh: /Users/username/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.0.3/dist/install scripts/dotnet-install.sh: Permission denied

The net8 sdk is installed in the /usr/local/share/dotnet/sdk/8.0.203 folder

How do I solve this problem?

### Please attach log file(s) here if you have any.

[...]

PR #1810, opened 2024-05-23
Update Axios Cache Interceptor Bundle for Runtime Error Change

nagilson:
Per the issue in https://github.com/arthurfiorette/tinylibs/issues/301, this is one of our highest facing issues. 
Let's see if this update resolves it. 

This is to solve the issue reported in https://github.com/dotnet/vscode-dotnet-runtime/issues/1734 
`Failed: n[r].split is not a function. Abortion. Please ensure that you are online.`

Along the way I also updated axios to make sure the newest version is utilized, and migrated to the newest version of vsce which is how we package our extension. In doing so, I had to migrate from node 16 to node 18.

### DIFF
--- a/pipeline-templates/install-node.yaml
+++ b/pipeline-templates/install-node.yaml
@@ -1,5 +1,5 @@
 steps:
-- task: NodeTool@0
+- task:  UseNode@1
   inputs:
-    versionSpec: '16.10.0'
+    version: '18.x'
   displayName: '⚛️ Install Node.js'
--- a/vscode-dotnet-runtime-library/src/test/unit/LoggingObserver.test.ts
+++ b/vscode-dotnet-runtime-library/src/test/unit/LoggingObserver.test.ts
@@ -30,7 +30,7 @@ suite('LoggingObserver Unit Tests', () => {
         // Check if the log file content is same as expected content
         fs.readdirSync(tempPath).forEach(file => {
             const logContent = fs.readFileSync(path.join(tempPath, file)).toString();
-            assert.include(logContent, fakeEvent.eventName, 'The log file does not contain the expected content that should be written to it?');
+            assert.include(logContent, fakeEvent.eventName, 'The log file does not contain the expected content that should be written to it');
         });
 
     }).timeout(4000);
--- a/vscode-dotnet-runtime-library/src/test/unit/LoggingObserver.test.ts
+++ b/vscode-dotnet-runtime-library/src/test/unit/LoggingObserver.test.ts
@@ -33,5 +33,5 @@ suite('LoggingObserver Unit Tests', () => {
             assert.include(logContent, fakeEvent.eventName, 'The log file does not contain the expected content that should be written to it?');
         });
 
-    });
+    }).timeout(4000);
 });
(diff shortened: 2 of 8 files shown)


joeloff:
BTW, does the error message in the description contain a typo? Shouldn't it read `Aborting`?

joeloff:
Are the lock files part of the PR or should they be excluded with .gitignore because they're artifacts from building locally?

nagilson:
Yes, that string is old or doesnt come from our repo, nowadays its says aborting ;) but good catch!

https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts#L211

vscode-dotnet-runtime-library/src/test/unit/LoggingObserver.test.ts
joeloff, line 1: nit: The assert message seems to be asking question? 😄

pipeline-templates/install-node.yaml
```diff
@@ -1,5 +1,5 @@
 steps:
-- task: NodeTool@0
+- task:  UseNode@1
   inputs:
-    versionSpec: '16.10.0'
+    version: '18.x'
```
joeloff, line 4: Is this an open version/minimum version? Any risk with installing different versions of 18.x and getting different results in pipelines?

joeloff (approved):

nagilson:
> Are the lock files part of the PR or should they be excluded with .gitignore because they're artifacts from building locally?

That's a good question. The standard convention is that the lock files should get checked in as it allows everyone to have a consistent build. We use yarn and npm and yarn bases its install off the lockfile, it is surprisingly tough to get the yarn file to match the npm file so deleting it or not maintaining them together would lose a lot of work. 

https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

pipeline-templates/install-node.yaml
baronfel: IMO no risk - this is the runtime used to build/package the tool, but not the runtime used by end users when they run our extension. The user-applicable nodejs runtime is controlled entirely by VSCode and kept consistent/updated by them.

nagilson: That's a good question and I agree with Chet. There is minimal risk. 18 is also LTS and pretty old at this point (think .NET 6)

vscode-dotnet-runtime-library/src/test/unit/LoggingObserver.test.ts
```diff
@@ -33,5 +33,5 @@ suite('LoggingObserver Unit Tests', () => {
             assert.include(logContent, fakeEvent.eventName, 'The log file does not contain the expected content that should be written to it?');
```
nagilson, line 33:
```suggestion
            assert.include(logContent, fakeEvent.eventName, 'The log file does not contain the expected content that should be written to it');
```

vscode-dotnet-runtime-library/src/test/unit/LoggingObserver.test.ts
nagilson: Fair enough, I fixed that 😄

PR #1816, opened 2024-05-28
Enable SBOM Generation 📜 

nagilson:
An SBOM is a formal record containing the details and supply chain relationships of various components used in building software.

We should produce such a thing for our software.
To see that this is successful, please see [Internal Build System URL]/_build/results?buildId=2463838&view=artifacts&pathAsName=false&type=publishedArtifacts


1es-azure-pipeline.yml
```diff
@@ -54,6 +54,8 @@ extends:
         name: NetCore1ESPool-Internal
         image: 1es-windows-2022
         os: windows
+      sbom:
+        enabled: false # SBOM requires very special setup, and is done in compliance.yml
```
nagilson, line 58:
```suggestion
        enabled: false # SBOM requires a special setup for node.js projects to ensure all dependencies are included.
```

pipeline-templates/sbom.yaml
```diff
@@ -0,0 +1,25 @@
+parameters:
+  pool: ''
+
+jobs:
+- job: SBOM
+  pool:
+    ${{ if eq(parameters.useOneEngineeringPool, 'true') }}:
+        ${{ parameters.pool }}
+    ${{ else }}:
+        vmImage: ${{ parameters.pool.vmImage }}
```
MiYanni, line 10: The `${{ parameters.pool }}` and vmImage parts are indented an extra 2 spaces.

pipeline-templates/sbom.yaml
```diff
@@ -0,0 +1,25 @@
...
+  displayName: 📝 SBOM Generation
+  dependsOn:
+  - ${{ parameters.pool.os }}_Build
+  - ${{ parameters.pool.os }}_Package
+  steps:
+    - task: ManifestGeneratorTask@0
+      displayName: 📃 SBOM generation
+      inputs:
+        BuildDropPath: $(System.DefaultWorkingDirectory)
```
MiYanni, line 19: nit: It is a stylistic choice, but this task item could be un-indented by 2 spaces (to make it the same alignment style as `dependsOn` and `outputs` arrays).

MiYanni (approved):

pipeline-templates/sbom.yaml
nagilson: Thanks for taking a look :) I think I like this approach since the step is still on the same indentation level, but I see your point

PR #1822, opened 2024-06-03
Improve VS Code SDK Error Diagnosability

nagilson:
We would like to have more concrete data on when an Install is requested, and then whether or not it succeeds.

We can gather this information currently by doing some math and piecing data together for whether the install requested is for a local runtime install, a global sdk install, or a local sdk install. But this is unnecessary when we can follow the same pattern I introduced for diagnosing runtime failures we began using 8 months ago.

In the same vein, we can also make it easier to tease apart whether a failure is one related to SDK Global Acquisition or not by wrapping all of those failures in a single event type. The original event type name must be preserved to allow us to tell which error is what, which is why the class type is forwarded to the new error class.

Unfortunately, the existing pattern in the code is a bit funky. An EventBus is used to listen to events to write to the log, and those Events must be of a certain type for the VSCode Telemetry pattern to report the event. However, we prefer to throw Errors with a callstack. Not all Events need a callstack, and making all Events Events and Errors simultaneously seems to break the API system from VS Code, at least in terms of including a lot of extraneous information in our DB we don't need.

Hence, the simplest/quickest way to do this was to create an Event Error type that gets passed the EventName. This is redundant and ugly, but it does allow us to differentiate between the Event and Error name when we want to do so.


vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
```diff
@@ -69,6 +74,14 @@ export async function callWithErrorHandling<T>(callback: () => T, context: IIssu
             );
         }
 
+        if(acquireContext?.installMode === 'sdk' && acquireContext.acquisitionContext?.installType === 'global')
+        {
+            context.eventStream.post(new DotnetGlobalSDKAcquisitionError(error, (caughtError?.eventType) ?? 'Unknown',
+             GetDotnetInstallInfo(acquireContext.acquisitionContext.version, acquireContext.installMode, 'global', acquireContext.installingArchitecture ??
+                DotnetCoreAcquisitionWorker.defaultArchitecture()
+             )));
```
MiYanni, line 82: nit: 1 space indentation is a bit odd.

MiYanni (approved):
Looks great! Definitely a good idea to have some more specific errors all around. And I like there's some code cleanup in this PR too. :shipit:

vscode-dotnet-runtime-library/src/Utils/ErrorHandler.ts
```diff
@@ -69,6 +74,14 @@ export async function callWithErrorHandling<T>(callback: () => T, context: IIssu
             );
         }
 
+        if(acquireContext?.installMode === 'sdk' && acquireContext.acquisitionContext?.installType === 'global')
+        {
+            context.eventStream.post(new DotnetGlobalSDKAcquisitionError(error, (caughtError?.eventType) ?? 'Unknown',
+             GetDotnetInstallInfo(acquireContext.acquisitionContext.version, acquireContext.installMode, 'global', acquireContext.installingArchitecture ??
```
nagilson, line 80:
```suggestion
               GetDotnetInstallInfo(acquireContext.acquisitionContext.version, acquireContext.installMode, 'global', acquireContext.installingArchitecture ??

```

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -201,8 +204,22 @@ export function activate(context: vscode.ExtensionContext, extensionContext?: IE
...
+            // Warning: Between now and later in this call-stack, the context 'version' is incomplete as it has not been resolved.
+            // Errors between here and the place where it is resolved cannot be routed to one another.
+
+            sdkAcquisitionWorker.setAcquisitionContext(commandContext);
+            telemetryObserver?.setAcquisitionContext(sdkContext, commandContext);
+
+            if(commandContext.version === '' || !commandContext.version)
+            {
+                    throw new EventCancellationError('BadContextualRuntimeVersionError',
+                        `No version was defined to install.`);
```
nagilson, line 220:
```suggestion
                    `No version was defined to install.`);
```

Issue #1827, opened 2024-06-06
[NETE2ESDK][Loc] Installing .NET SDK failed in Simple Chinese OS


NicoleWang001:
### Describe the bug.

Installing .NET SDK failed in Simple Chinese OS

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

Try to install the .NET SDK System-Wide in Simple Chinese Windows OS.

### Expected Behavior

SDK could be installed succeed.

### Exceptions & Errors (if any)

Failed to download .NET 8.0.301~unknown: .NET Acquisition Failed: The requested version
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/d705ad85-7561-41cc-986c-6be1a041d571)

### .NET Install Tool Version

2.0.6, 2.0.5

### Your operating system? Please provide more detail, such as the version and or distro above.

Windows

### VS Code Version

1.90.0

NicoleWang001:
.NET SDK could be installed successfully in English OS.
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/1e99f2c8-33b3-464a-ba6b-b4c968be2750)

nagilson:
this should be fixed pending the new build

NicoleWang001:
Thanks, @nagilson. Checked the new build, this issue was fixed. So close it.
![image](https://github.com/dotnet/vscode-dotnet-runtime/assets/1578239/1f6eac6a-7231-44d8-908f-f2303302dee8)

PR #1830, opened 2024-06-06
Don't rely on english output for registry lookups

nagilson:
⚠️ Problem: We relied on 'ERROR' to determine if an error occurred during a registry lookup. This lookup is to find already existing installs of the .NET SDK on the machine. If the system display language is not english, reg.exe does not return in english. Unfortunately there is no option to standardize output or output language. https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/reg-query

💡 Solution: Read the status code first to check if the query resulted in an error.

❓ Alternative solution: Add registry-js library from GitHub to query registry in a less hacky way. Unfortunately this adds native binaries to our extension and would require us to ship platform specific packages, which is extra work I would prefer not to do.

💥 : Ways this could be improved: 
The `CommandExecutor` only returns a string at the moment. It could be improved to return a result object with stdout, stderr, and status code, so we don't need to run the command twice. This is obviously the better solution in the long term. But it would require updating a lot of code, and I don't think it's a priority to do that right now.

♻️ Testing: I was unable to get reg.exe to output in any language besides english but it does seem to happen. Setting the system locale, preferences, display language, and preferred language did not impact reg.exe after the fact. Will rely on vendors to test on a machine with different OS language.
image

image

Resolves: 
https://github.com/dotnet/vscode-dotnet-runtime/issues/1827
https://github.com/dotnet/vscode-csharp/issues/7158

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
@@ -278,7 +278,7 @@ Please correct your PATH variable or make sure the 'open' utility is installed s
      */
     private extractVersionsOutOfRegistryKeyStrings(registryQueryResult : string) : string[]
     {
-        if(registryQueryResult.includes('ERROR') || registryQueryResult === '')
+        if(registryQueryResult === '')
         {
                 return [];
         }
@@ -346,8 +346,19 @@ Please correct your PATH variable or make sure the 'open' utility is installed s
                 {
                     const registryQueryCommand = path.join(`${process.env.SystemRoot}`, `System32\\reg.exe`);
                     // /reg:32 is added because all keys on 64 bit machines are all put into the WOW node. They won't be on the WOW node on a 32 bit machine.
-                    const fullQuery = `${registryQueryCommand}`;
-                    const installRecordKeysOfXBit = await this.commandRunner.execute(CommandExecutor.makeCommand(registryQueryCommand, [`query`, `${query}`, `\/reg:32`]));
+                    const command = CommandExecutor.makeCommand(registryQueryCommand, [`query`, `${query}`, `\/reg:32`]);
+
+                    let installRecordKeysOfXBit = '';
+                    const oldReturnStatusSetting = this.commandRunner.returnStatus;
+                    this.commandRunner.returnStatus = true;
+                    const registryLookupStatusCode = await this.commandRunner.execute(command);
+                    this.commandRunner.returnStatus = oldReturnStatusSetting;
+
+                    if(registryLookupStatusCode === '0')
+                    {
+                        installRecordKeysOfXBit = await this.commandRunner.execute(command);
+                    }
+
                     const installedSdks = this.extractVersionsOutOfRegistryKeyStrings(installRecordKeysOfXBit);
                     // Append any newly found sdk versions
                     sdks = sdks.concat(installedSdks.filter((item) => sdks.indexOf(item) < 0));
--- a/vscode-dotnet-runtime-library/src/test/mocks/MockObjects.ts
+++ b/vscode-dotnet-runtime-library/src/test/mocks/MockObjects.ts
@@ -325,6 +325,7 @@ export class MockCommandExecutor extends ICommandExecutor
 {
     private trueExecutor : CommandExecutor;
     public fakeReturnValue = '';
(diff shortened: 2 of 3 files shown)


vscode-dotnet-runtime-library/src/test/mocks/MockObjects.ts
```diff
@@ -342,6 +344,7 @@ export class MockCommandExecutor extends ICommandExecutor
     {
         this.attemptedCommand = CommandExecutor.prettifyCommandExecutorCommand(command);
 
+
```
nagilson, line 347:
```suggestion
```

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -346,8 +346,19 @@ Please correct your PATH variable or make sure the 'open' utility is installed s
...
                     // /reg:32 is added because all keys on 64 bit machines are all put into the WOW node. They won't be on the WOW node on a 32 bit machine.
-                    const fullQuery = `${registryQueryCommand}`;
-                    const installRecordKeysOfXBit = await this.commandRunner.execute(CommandExecutor.makeCommand(registryQueryCommand, [`query`, `${query}`, `\/reg:32`]));
+                    const command = CommandExecutor.makeCommand(registryQueryCommand, [`query`, `${query}`, `\/reg:32`]);
+
+                    let installRecordKeysOfXBit = '';
+                    const oldReturnStatusSetting = this.commandRunner.returnStatus;
+                    this.commandRunner.returnStatus = true;
+                    const registryLookupStatusCode = await this.commandRunner.execute(command);
+                    this.commandRunner.returnStatus = oldReturnStatusSetting;
```
nagilson, line 355: Obviously this should be false, but I would prefer to keep this pattern so we can update them all at the same time if/when we want to fix this code.

vscode-dotnet-runtime-library/src/test/mocks/MockObjects.ts
```diff
@@ -342,16 +344,29 @@ export class MockCommandExecutor extends ICommandExecutor
...
         }
-        else if(this.otherCommandsToMock.some(x => x.includes(command.commandRoot)))
+
+        for(let i = 0; i < this.otherCommandPatternsToMock.length; ++i)
+        {
+            const commandPatternToLookFor = this.otherCommandPatternsToMock[i];
+            if(command.commandRoot.includes(commandPatternToLookFor) ||
+                command.commandParts.some((arg) => arg.includes(commandPatternToLookFor)))
+            {
+                if(this.shouldReturnStatus() && this.otherCommandsStatusReturnValues)
```
nagilson, line 359:
```suggestion
                if(this.shouldReturnStatus() && this.otherCommandsStatusReturnValues[i])
```

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
baronfel: So if I understand correctly, what we're doing here is a two-phase lookup - in the first phase were invoking the command specifically to see if it returns successfully (since our current usage/API for executing commands here doesn't give us both exit code + stdin/stderr?). And then, once we are convinced that we can successfully get the reg query result, we invoke the query again in an execution mode where we get the result of the query back?

Forgind:
Is this fixing a theoretical problem or a user-reported problem? I saw you linked two issues, but both included screenshots that said 'error' somewhere, and you said you had difficulty reproducing this for testing. MSBuild looks for Error:, but it isn't a problem because the Error: part isn't localized anyway.

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
nagilson: That is 100% correct

nagilson:
> Is this fixing a theoretical problem or a user-reported problem? I saw you linked two issues, but both included screenshots that said 'error' somewhere, and you said you had difficulty reproducing this for testing. MSBuild looks for Error:, but it isn't a problem because the Error: part isn't localized anyway.

Good question. The German user had 'FEHLER' instead of error and the vendors were able to repro the bug with CN output and also confirm the fix.

baronfel (approved):
LGTM, but it would be good to log your comment about making the API for command results better

PR #1843, opened 2024-06-18
Add Support for ASP.NET Installation

nagilson:
# Intro
To support aspire, we need to be able to download the ASP.NET Runtime and install a local copy of it onto the disk.
We can do this by using the -Runtime flag to the .NET Install Script.

Much of the code was already refactored prior to use a 'DotnetInstallMode' for 'runtime', 'sdk', or 'aspnetcore'. 

# API Changes

This adds to the acquire API via changing the public 'IDotnetAcquireContext' type and giving it the 'mode' option. This option allows API callers such as C# to install aspnetcore runtimes. If no mode is provided, it will be undefined and use 'runtime', which was the old, pre-existing scenario.

Note that for the .NET SDK installation API, it has its own endpoint and the mode is ignored. This is not the greatest pattern but we can't change public interfaces. We can, however, improve the `acquire` endpoint in the future to make the other API endpoints obsolete. That is more work that could go in a different PR from this one.

# Code Changes

This updates the 'VersionResolver' and 'DirectoryProvider' to find aspnetcore versions, adds an endpoint to install aspnet core, and also generalizes events using a 'ModalEventRepublisher' to prevent duplicated code where we would publish the same telemetry event but with a different name for each installation mode. 

E.g. 'AcquisitionStarted' has an 'SDKAcquisitionStarted', 'RuntimeAcquisitionStarted', etc, now the code just needs to publish one and the Republisher will publish the more specific event for upstream telemetry sources to analyze. We do this via the publish-subscriber pattern.

It also adds `~aspnetcore` to the install key for aspnetcore runtimes to differentiate the installation directories. This is important because `'-'` used to be used to parse whether the install record was `'global'` and `'~'` used to be used to determine whether the install had an architecture. (Old records without an architecture did not have such a thing, the record would just be `'7.0.20'`, now it is `'7.0.20~x64'`, or `'7.0.20-global~x64'`. Adding `~aspnetcore` to the end will not break the logic due to when it was added, and because we now store the install information as a class instead of in strings.

Adds 2.0.7 version branding for release as well.


vscode-dotnet-runtime-library/src/Acquisition/DirectoryProviderFactory.ts
```diff
@@ -12,5 +13,8 @@ import { SdkInstallationDirectoryProvider } from './SdkInstallationDirectoryProv
 
 export function directoryProviderFactory(mode: DotnetInstallMode, storagePath: string) : IInstallationDirectoryProvider
 {
-    return mode === 'runtime' ? new RuntimeInstallationDirectoryProvider(storagePath) : new SdkInstallationDirectoryProvider(storagePath);
+    return mode === 'runtime' ? new RuntimeInstallationDirectoryProvider(storagePath) :
+    mode === 'sdk' ? new SdkInstallationDirectoryProvider(storagePath) :
+    mode === 'aspnetcore' ? new ASPNetRuntimeInstallationDirectoryProvider(storagePath)
+    : new RuntimeInstallationDirectoryProvider(storagePath); // default if no mode is provided - should never happen
```
tmeschter, line 19: If it should never happen, should this throw an error or fail in some other way?

vscode-dotnet-runtime-library/src/Acquisition/IDotnetCoreAcquisitionWorker.ts
```diff
@@ -14,9 +14,9 @@ export interface IDotnetCoreAcquisitionWorker
 {
     uninstallAll(eventStream : IEventStream, storagePath : string, extensionState : IExtensionState): void;
 
-    acquireRuntime(context: IAcquisitionWorkerContext, invoker : IAcquisitionInvoker): Promise<IDotnetAcquireResult>;
+    acquireLocalRuntime(context: IAcquisitionWorkerContext, invoker : IAcquisitionInvoker): Promise<IDotnetAcquireResult>;
```
tmeschter, line 17: Should this have an `acquireLocalASPNET` method as well?

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
```diff
@@ -29,12 +31,23 @@ export class EventBasedError extends Error
...
+{
+    abstract readonly mode : DotnetInstallMode;
+    abstract readonly installType : DotnetInstallType;
+}
+
+export class DotnetAcquisitionStarted extends GenericModalEvent
+{
     public readonly eventName = 'DotnetAcquisitionStarted';
     public readonly type = EventType.DotnetAcquisitionStart;
+    public readonly mode;
```
tmeschter, line 44: Should this have a type?

tmeschter (approved):

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
nagilson: That's a good question. The type is inherited from the base class, but I will include the type here because that's what I've done elsewhere!

vscode-dotnet-runtime-library/src/Acquisition/IDotnetCoreAcquisitionWorker.ts
nagilson: I wrote the same comment on my review but discarded it... idk why, good point so Ill add it!

vscode-dotnet-runtime-library/src/Acquisition/DirectoryProviderFactory.ts
nagilson: Hm, that's a good point -- it could! I think it's a matter of preference? In the future if someone adds a new mode and forgets to update this it will fail, which could be good because it will let them know they should update it.

That said, even if they dont update it, it will function properly. Its also pretty clear IMO that this is an area that can be updated when a new mode is added, so I lean towards keeping it. This has also been vendor tested so I am more hesitant to take potential changes like this one, I might do it in another PR.

PR #1854, opened 2024-06-25
Handle No Connection Errors Better

### LINKED ISSUE #2: Add support for Linux ARM and Alpine
Much like https://github.com/OmniSharp/omnisharp-vscode/issues/2771 we will need to do work to properly acquire the "right" bits for ARM scenarios.

We need to adjust the dynamic acquisition logic to get a compatible version of the runtime for either ARM or Alpine. (On first check, Alpine appeared to work when using the .NET Core image – I have not checked ARM, however.)  We briefly [describe the issue and how this can be detected here](https://code.visualstudio.com/api/advanced-topics/remote-extensions#supporting-nonx8664-hosts-or-alpine-linux-containers).

nagilson:
In respone to the issue where we never timeout if there is no internet connection at https://github.com/microsoft/vscode-dotnettools/issues/1028

Axios (web request library we use) has a bug with no connection timeouts. Its timeout handler only works with request timeouts, not no connection timeouts. More at https://stackoverflow.com/questions/36690451/timeout-feature-in-the-axios-library-is-not-working. So yeah, it's going to be stuck forever waiting for a connection. I fixed #2 in the past by adding a custom timeout with a cancellation token. dotnet/vscode-dotnet-runtime@ced82f3#diff-ad11027eb5a8406f68425da0e49b1c8a4aa83394444043e247c10e7bb1976999

It looks like I got rid of that because I forgot about the axios timeout issue. dotnet/vscode-dotnet-runtime@3b1c251

This was initially in 1.8.1 and the failure rate improved a lot with 2.0.0 but that had a huge amount of code change. We need to inspect this change and QA it to make sure it doesnt make things worse. This will be the only change for 2.0.8 because of this.

This code used to be in the codebase for several versions so we have some degree of confidence in it.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
@@ -71,9 +71,18 @@ export class WebRequestWorker
         {
             throw new EventBasedError('AxiosGetFailedWithInvalidURL', `Request to the url ${this.url} failed, as the URL is invalid.`);
         }
+        const timeoutCancelTokenHook = new AbortController();
+        const timeout = setTimeout(() =>
+        {
+            timeoutCancelTokenHook.abort();
+            const formattedError = new Error(`TIMEOUT: The request to ${this.url} timed out at ${this.websiteTimeoutMs} ms. This only occurs if your internet
+ or the url are experiencing connection difficulties; not if the server is being slow to respond. Check your connection, the url, and or increase the timeout value here: https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md#install-script-timeouts`);
+            this.context.eventStream.post(new WebRequestError(new EventBasedError('WebRequestError', formattedError.message, formattedError.stack), null));
+            throw formattedError;
+        }, this.websiteTimeoutMs);
 
-        const response = await this.client.get(url, { ...options });
-
+        const response = await this.client.get(url, { signal: timeoutCancelTokenHook.signal, ...options });
+        clearTimeout(timeout);
         return response;
     }
 


nagilson:
https://github.com/dotnet/vscode-dotnet-runtime/issues/1858 is related and will be the complete fix for the user problem.

richlander:
I get "No connection" a lot (on preview builds). I can share logs or test builds if that would be helpful. I am on Apple M1.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -71,9 +71,18 @@ export class WebRequestWorker
...
         }
-
-        const response = await this.client.get(url, { ...options });
-
+        const timeoutCancelTokenHook = new AbortController();
+        const timeout = setTimeout(() =>
+        {
+            timeoutCancelTokenHook.abort();
+            const formattedError = new Error(`TIMEOUT: The request to ${this.url} timed out at ${this.websiteTimeoutMs} ms. This only occurs if your internet
+ or the url are experiencing connection difficulties; not if the server is being slow to respond. Check your connection, the url, and or increase the timeout value here: https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md#install-script-timeouts`);
```
MiYanni, line 79: I can't remember, does this repo have localization? Just something worth mentioning for an error like this since I'm assuming the user will see this one.

MiYanni (approved):

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
nagilson: That's a good question. It does not, but eventually it will need to.

PR #1855, opened 2024-06-25
Prevent EPERM error hit when trying to open installer.exe

nagilson:
# Changes to Permissions

`fs.chmod` does not set execute permissions on windows https://nodejs.org/api/fs.html#fschmodpath-mode-callback. This means we should also rely on icacls to try to do so.
If we still dont have permission, we can collect more information so we know how to fix it better going forward.

This also makes each installer get its own unique folder. SHA is used since it is an algorithm widely available on multiple OS. https://support.globalsign.com/ssl/ssl-certificates-life-cycle/sha-256-compatibility
If this does not work, we can ask for elevation using the vscode library on windows if permission to run the installer is denied.

# Command Exeuctor Private Interface Changes
Make Command Executor return an object instead of string. This prevents us from having to run commands twice to get the status and stdout.

Caution must be taken to make sure that the assumptions here in what is returned as stdout vs stderr are correct.


vscode-dotnet-runtime-library/src/Acquisition/GenericDistroSDKProvider.ts
```diff
@@ -23,47 +23,39 @@ export class GenericDistroSDKProvider extends IDistroDotnetSDKProvider
...
+            symLinkReadCommand = CommandExecutor.replaceSubstringsInCommands(symLinkReadCommand, this.missingPathKey, commandResult[0].stdout);
+            const resolvedPath = (await this.commandRunner.executeMultipleCommands(symLinkReadCommand))[0].stdout;
             if(resolvedPath)
             {
                 return path.dirname(resolvedPath.trim());
             }
         }
 
-        return commandResult[0] ?? null;
+        return commandResult[0].stdout.trim() ?? null;
```
nagilson, line 58:
```suggestion
        return commandResult[0].stdout ?? null;
```

vscode-dotnet-runtime-library/src/Acquisition/GenericDistroSDKProvider.ts
```diff
@@ -73,8 +65,9 @@ export class GenericDistroSDKProvider extends IDistroDotnetSDKProvider
         command = CommandExecutor.replaceSubstringsInCommands(command, this.missingPackageNameKey, sdkPackage);
         const commandResult = (await this.commandRunner.executeMultipleCommands(command))[0];
 
+        // todo: this seems very wrong
```
nagilson, line 68:
```suggestion
```

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -471,29 +449,25 @@ result: ${commandResult.toString()} had no status or signal.`));
...
-                    return '';
-                }
-                else
-                {
-                    if(commandResult.stdout)
-                    {
-                    this.context?.eventStream.post(new CommandExecutionStdOut(`The command ${fullCommandStringForTelemetryOnly} encountered stdout:
+            })();
+
+            return { status: statusCode, stderr: commandResult.stderr?.toString() ?? '', stdout: commandResult.stdout?.toString() ?? ''}
```
nagilson, line 454: We could probably trim here and not have to trim everywhere else where its needed, but there are some situations where whitespace matters so I am hesitant to do so.

MiYanni (approved):

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -141,19 +141,29 @@ We cannot verify .NET is safe to download at this time. Please try again later.`
...
         }
 
         await this.webWorker.downloadFile(installerUrl, installerPath);
         try
         {
             fs.chmodSync(installerPath, 0o744);
+
+            if(os.platform() === 'win32') // Windows does not have chmod +x ability with nodejs.
+            {
+                const commandRes = await this.commandRunner.execute(CommandExecutor.makeCommand('icacls', ['/grant', `"${installerPath}"`, `"%username%":F`, '/t', '/c']), null, false);
```
joeloff, line 160: Are these parameters ordered? According to https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/icacls#syntax the syntax should be `icacls <filename> [/grant[:r] <sid>:<perm>[...]]`

Also note that because you're not using `grant:r` you will be adding and retaining existing perms, which is different from the explicit 744 that chmod is doing. Also, should the chmod be moved into a separate `if` for non-windows and then explicitly set the ACL.

Can you share the results of this? Run the code, then from Powershell, run `Get-Acl <path-to-filename> | Select-Object -ExpandProperty Access` and just strip out any PII before pasting in it (or ping me offline).

vscode-dotnet-runtime-library/src/Acquisition/DotnetInstall.ts
```diff
@@ -42,18 +43,16 @@ export function IsEquivalentInstallationFile(a: DotnetInstall, b: DotnetInstall)
...
     return a.installKey === b.installKey;
 }
 
 /**
  * @returns A string set representing the installation of either a .NET runtime or .NET SDK.
  */
-export function InstallToStrings(key: DotnetInstall | null) {
-    if (!key) {
-        return { installKey: '', version: '', architecture: '', isGlobal: '', installMode: '' };
-    }
```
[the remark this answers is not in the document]
nagilson: That's a great point. `EditorConfig` is pretty awesome, but the javascript and typescript settings appear to be less standardized than the C# and .NET settings for brackets. Anywho, I added a config file. :)

That is the convention, but I disagree with the convention. I always think about apples SSL bug https://www.codecentric.de/wissens-hub/blog/curly-braces#:~:text=To%20sum%20up%20the%20debacle,To%20say%20the%20least. and that is a large part of my reasoning for my opinion here.

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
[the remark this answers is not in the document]
nagilson: That's a very good question. 

I did some digging, the internet seems to suggest you need administrator rights to run `icalcs`. That makes sense. But I just tried to do it without admin rights on a local user file without administrator rights and it also worked.

This is being run by a sub-process of vscode, which should have the same privileges as vscode. If it is an admin it is not a concern. If it is not admin it should be running under the user account. The file its trying to add +x to is also a user local file, so my understanding is that it should function properly and be permitted.

Unfortunately we dont have any actual customer reports of EPERM issues, just that its showing up in our error data. I think either way though, if this does not work, the primary option we have is just to ask for elevation right away.

vscode-dotnet-runtime-library/src/Acquisition/DotnetInstall.ts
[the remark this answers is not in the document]
MiYanni: Personally, learning to read/write in the conventions of another language is very useful and transferrable. I tried to write in a C#-style in ColdFusion and it didn't work whatsoever. Not just stylistically, but conceptually. But TS is pretty loose on this particular convention. However, any TS I've worked on always had the curly braces at the end of the line, to match the JavaScipt style.

PR #1859, opened 2024-07-03
Timeout Installer.Exe

nagilson:
Currently we will wait forever for the .NET installer to complete. This is not correct behavior, especially when users sometimes ignore the installer prompt and leaving it running, it will block our thread from continuing.

This PR adds the timeout option to nodejs `spawn`. This sends the SIGTERM response. 
https://nodejs.org/api/child_process.html#child_processspawncommand-args-options

This also renames a function to be semantically correct.. this change is in a lot of files, but it does not have an impact on anything. Additionally, some functions that can be used more broadly were moved out of the command executor class.

This PR also adds the editorconfig file to standardize repo style conventions. More work will likely need to be done to apply it to files in another PR.


vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -426,6 +380,12 @@ ${(commandOutputJson as CommandExecutorResult).stderr}`));
 with options ${JSON.stringify(options)}.`));
             const commandResult = proc.spawnSync(command.commandRoot, command.commandParts, options);
 
+            if(os.platform() === 'win32')
+            {
+                proc.spawn('taskkill', ['/pid', commandResult.pid.toString(), '/f', '/t']);
```
nagilson, line 385: this is here to try to kill a timed out installer. will not work without privelleges.

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -411,9 +361,17 @@ ${(commandOutputJson as CommandExecutorResult).stderr}`));
...
+            options.cwd = path.resolve(__dirname);
+        }
+        if(options && !options?.shell)
+        {
+            options.shell = true;
+        }
         if(!options)
         {
-            options = {cwd : path.resolve(__dirname), shell: true};
+            options = {};
```
nagilson, line 374:
```suggestion
            options = {cwd : path.resolve(__dirname), shell: true};
```

nagilson:
LGTM on my own review.

joeloff:
Are you launching the install with UI? Have you considered maybe just running it with /passive. You'll get progress, but not other UI elements

nagilson:
Oh interesting, I just tried it out and /passive will even auto elevate. That's wonderful. I will add that to this, thanks @joeloff!

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -1,516 +1,519 @@
...
+    export const showAcquisitionLog = 'showAcquisitionLog';
+    export const ensureDotnetDependencies = 'ensureDotnetDependencies';
+    export const reportIssue = 'reportIssue';
+}
+
+const commandPrefix = 'dotnet';
+const configPrefix = 'dotnetAcquisitionExtension';
+const displayChannelName = '.NET Install Tool';
+const defaultTimeoutValue = 600;
+const moreInfoUrl = 'https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md';
```
MiYanni, line 94: Should this always link to `main` or some static commit? If you go to that URL right now and then hit `y`, it'll change your address bar to have the commit ID in the URL. It turns into `https://github.com/dotnet/vscode-dotnet-runtime/blob/e68687f05ea3f59dc08437bfaa6c4d07774cf953/Documentation/troubleshooting-runtime.md` for me right now. You should also use an aka.ms link (let me know if you need help creating them).

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
MiYanni: Does `spawn` care about the result of the operation?

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -222,14 +222,14 @@ ${axiosBasedError.cause? `Error Cause: ${axiosBasedError.cause!.message}` : ``}
 Please ensure that you are online.
 
 If you're on a proxy and disable registry access, you must set the proxy in our extension settings. See https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md.`);
```
MiYanni, line 224: Similar comment about the same link in the other file.

vscode-dotnet-runtime-library/src/Acquisition/GlobalInstallerResolver.ts
```diff
@@ -277,7 +277,7 @@ Please report this issue so it can be remedied or investigated.`), getInstallKey
         const fileErr = new DotnetNoInstallerFileExistsError(new EventBasedError('DotnetNoInstallerFileExistsError',
         `The SDK installation files for version ${specificVersion} running on ${desiredRidPackage} couldn't be found.
 Is the version in support? Note that -preview versions or versions with build numbers aren't yet supported.
-Visit https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core for support information.`), getInstallKeyFromContext(this.context));
+Visit https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core for support information.`), getInstallFromContext(this.context));
```
MiYanni, line 280: This link should be culture agnostic. Meaning, `https://dotnet.microsoft.com/platform/support/policy/dotnet-core`

MiYanni (approved):

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
nagilson: It does not.

vscode-dotnet-runtime-library/src/Acquisition/GlobalInstallerResolver.ts
nagilson, line 280:
```suggestion
Visit https://dotnet.microsoft.com/platform/support/policy/dotnet-core for support information.`), getInstallFromContext(this.context));
```

nagilson: That's a good catch, thank you. Fixed it!

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: I think it should default to `main` so it always provides the latest documentation if we ever improve the document.

The aka.ms link idea is a good one :) I will create another issue to do that. I appreciate that you offered to help, I'll see what I can do first! https://github.com/dotnet/vscode-dotnet-runtime/issues/1867

MiYanni: The reason to not link to main is that file path changes (or renames) would make the URL no longer valid. But, if you use an aka.ms link, then main is fine as the forwarding link can be updated without issue if the file is moved/renamed.

Issue #1806, opened 2024-05-22
Apt-Get Lock can be Busy when Executing Linux Install


nagilson:
### Describe the bug.

When installing the Global .NET SDK on the machine, either via C# DevKit's SDK walkthrough 'Install' button or via our command, Linux users who are supported (newer versions of Ubuntu and RHEL) may face the following error:

`Cancelling .NET Install, as command apt-get update returned with status 100.`

This error occurs because the apt-get lock is busy, (or for one of many other reasons related to apt-get failing.) This could happen for any number of reasons, e.g. another install on the system, of which the system may occasionally do automatically.

This is the highest cause of failure on Linux we have and constitutes a decent chunk of our failure rate.
If you want a workaround you can try to install the SDK yourself or push the extension through this step by running `sudo apt-get update` manually and see what the error is, or` sudo apt-get install -y dotnet-sdk-8.0` (or your given version) and try again.

For some other reasons this might happen, see:
https://stackoverflow.com/questions/38002543/apt-get-update-returned-a-non-zero-code-100?rq=4

For us, we should investigate remedying this. The first thing we can do and maybe the only one is wait and spin to see if the apt-get lock opens up if taken.

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

Install SDK on Linux with apt-get lock taken

### Expected Behavior

Should install once apt-get is free

### .NET Install Tool Version

2.0.5

### Your operating system? Please provide more detail, such as the version and or distro above.

N/A

baronfel:
An additional note - any mechanism we create here should be able to work on other platforms as well - the `yum` manager has similar locks, as do Windows MSI installations.

PR #1861, opened 2024-07-03
Update Linux Apt Commands to Handle Lock Timeouts

nagilson:
resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1806#event-13429482290

A lot of users are seeing failure due to the DPKG lock being held by another process. We can add a flag to wait for the lock to be freed. This wont fix everything, as Apt will return 100 whenever there is any failure. https://github.com/Debian/apt/blob/aa56836331870d975c212a5df2f13db9ce3914bf/apt-private/private-cmndline.cc#L602

This means the errors we are seeing are not just due to locking errors. But, perhaps this will help.

This will also add a timeout to our own executions so that they dont keep the lock.

Also this will try to run update twice, to see if that 2nd try will work. It seems a bit silly, but it does actually seem to improve the chance of success when something else is running on the system.

The timeout time is mildly arbitrary.... we can try to improve that in the future to be based off a fraction of the total timeout time, or something like this. But I would like to see the impact of how much this helps without this change.


vscode-dotnet-runtime-library/distro-data/distro-support.json
```diff
@@ -34,15 +43,18 @@
...
                 ]
             },
             {
                 "runUnderSudo": true,
                 "commandRoot": "apt-get",
                 "commandParts": [
+                    "-o",
+                    "DPkg::Lock::Timeout=120",
                     "upgrade",
-                    "-y",
```
nagilson, line 45: To check: does apt-get upgrade not accept -y? need to test tmrw on a vm.

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -77,26 +78,33 @@ export class InstallTrackerSingleton
...
-            finally
+            else
             {
-                await release();
+                await lockfile.lock(lockPath, { retries: { retries: 10, minTimeout: 5, maxTimeout: 10000 } })
+                .then(async (release) =>
+                {
+                    returnResult = await f(...(args));
+                    this.eventStream?.post(new DotnetLockReleasedEvent(`Lock about to be released.`, new Date().toISOString(), lockPath, lockPath));
+                    await release();
```
nagilson, line 101:
```suggestion
                    return release();
```

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -77,26 +78,33 @@ export class InstallTrackerSingleton
...
-            const release = await lockfile.lock(lockPath, { retries: { retries: 10, minTimeout: 5, maxTimeout: 2000 } });
-            try
-            {
-                return await f(...(args));
-            }
-            finally
+            else
             {
-                await release();
+                await lockfile.lock(lockPath, { retries: { retries: 10, minTimeout: 5, maxTimeout: 10000 } })
```
MiYanni, line 96: What units are the min and max timeout in?

MiYanni (approved):

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
nagilson: they are in MS. The library will do an exponential equation where it starts trying again after 5 ms and then takes longer each retry until its at 10000 ms.

vscode-dotnet-runtime-library/distro-data/distro-support.json
nagilson: it does.

PR #1869, opened 2024-07-10
Add AcquireStatus support for ASP.NET runtimes

nagilson:
The C#DevKit team would like this API to support ASP.NET Runtimes as well.


MiYanni (approved):
:shipit:

PR #1874, opened 2024-07-11
Fix Linux Update

nagilson:
This adds a very simple additional test for the acquireStatus API earlier, that I forgot to push while I wasn't feeling well. 

This also fixes an issue with update, the || operator does not work in this context and itll treat it as an optional flag thats invalid. It was hacky to do anyway, so I just got rid of it.

### DIFF
--- a/vscode-dotnet-runtime-extension/CHANGELOG.md
+++ b/vscode-dotnet-runtime-extension/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning].
 
 ## [Unreleased]
 
-## [2.0.9] - 2024-??-??
+## [2.0.9] - 2024-07-12
 
 Fixes a bug where permissions were not granted to run the .NET Installer. (EPERM issue.)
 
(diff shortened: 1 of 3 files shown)


vscode-dotnet-runtime-extension/CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Performance improvements.
 
 Fixes a bug with apt-get lock holding (status 100) on Linux with package management.
 
+Adds acquireStatus support for ASP.NET runtime installs.
+
+Other bug fixes.
```
MiYanni, line 22: nit: Maybe `Minor bug fixes.` would be a bit more descriptive?

MiYanni (approved):

vscode-dotnet-runtime-extension/CHANGELOG.md
nagilson: I think that is a better way of putting it. :)

nagilson: It got auto-merged though so I will fix it next time 🤦

PR #1878, opened 2024-07-15
Add failure handling for file integrity check

nagilson:
This is now what is hitting the EPERM and ENOENT errors. We need to skip the check and see what happens from here, such as allowing to elevate via windows.

This also adds specific handlers for when we fail to download the SDK.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
@@ -184,7 +184,7 @@ We cannot verify our .NET file host at this time. Please try again later or inst
 
     private async userChoosesToContinueWithInvalidHash() : Promise<boolean>
     {
-        const yes = validationPromptConstants.allowOption
+        const yes = validationPromptConstants.allowOption;
         const no = validationPromptConstants.cancelOption;
         const message = validationPromptConstants.noSignatureMessage;
 
--- a/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
@@ -17,6 +17,7 @@ import {
     DotnetAcquisitionAlreadyInstalled,
     DotnetConflictingGlobalWindowsInstallError,
     DotnetFileIntegrityCheckEvent,
+    DotnetFileIntegrityFailureEvent,
     DotnetInstallCancelledByUserError,
     DotnetNoInstallerResponseError,
     DotnetUnexpectedInstallerOSError,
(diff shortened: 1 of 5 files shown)


vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -181,34 +182,57 @@ We cannot verify .NET is safe to download at this time. Please try again later.`
...
 
-    private async installerFileHasValidIntegrity(installerFile : string) : Promise<boolean>
+    private async userChoosesToContinueWithInvalidHash() : Promise<boolean>
     {
-        const realFileHash = await this.file.getFileHash(installerFile);
-        this.acquisitionContext.eventStream.post(new DotnetFileIntegrityCheckEvent(`The hash of the installer file we downloaded is ${realFileHash}`));
-        const expectedFileHash = this.installerHash;
-        this.acquisitionContext.eventStream.post(new DotnetFileIntegrityCheckEvent(`The valid and expected hash of the installer file is ${expectedFileHash}`));
+        const yes = validationPromptConstants.allowOption
+        const no = validationPromptConstants.cancelOption;
```
MiYanni, line 188: nit: I know semicolons are not required, but maybe it should be in the .editorconfig/tslint.json to suggest them for consistency? Here, the `yes` line doesn't have one.

MiYanni (approved):

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
nagilson: Good catch! unfortunately I couldnt find an editorconfig setting, it looks like there might be an eslint one. First, we would need to change from tslint to eslint, so I think that should be done separately

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -181,34 +182,57 @@ We cannot verify .NET is safe to download at this time. Please try again later.`
...
     }
 
-    private async installerFileHasValidIntegrity(installerFile : string) : Promise<boolean>
+    private async userChoosesToContinueWithInvalidHash() : Promise<boolean>
     {
-        const realFileHash = await this.file.getFileHash(installerFile);
-        this.acquisitionContext.eventStream.post(new DotnetFileIntegrityCheckEvent(`The hash of the installer file we downloaded is ${realFileHash}`));
-        const expectedFileHash = this.installerHash;
-        this.acquisitionContext.eventStream.post(new DotnetFileIntegrityCheckEvent(`The valid and expected hash of the installer file is ${expectedFileHash}`));
+        const yes = validationPromptConstants.allowOption
```
nagilson, line 187:
```suggestion
        const yes = validationPromptConstants.allowOption;
```

PR #1880, opened 2024-07-15
Interpret Installer Exit Codes

nagilson:
These will give the user more actionable items, and also increase the chances the post a log so we can help them if something goes wrong and understand the issue better.

Error code interpretation comes from https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--1300-1699- pages, the installer source code, and joeloff (area expert)

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
@@ -78,6 +78,40 @@ export class WinMacGlobalInstaller extends IGlobalInstaller {
         this.webWorker = new WebRequestWorker(context, installerUrl);
     }
 
+    public static InterpretExitCode(code : string) : string
+    {
+        const reportLogMessage = `Please provide your .NET Installer log (note our privacy notice), which can be found at %temp%.
+The file has a name like 'Microsoft_.NET_SDK*.log and should appear in recent files.
+This report should be made at https://github.com/dotnet/vscode-dotnet-runtime/issues.`
+
+        switch(code)
+        {
+            case '1':
+                return `The .NET SDK installer has failed with a generic failure. ${reportLogMessage}`;
+            case '5':
+                return `Insufficient permissions are available to install .NET. Please run the installer as an administrator.`;
+            case '67':
+                return `A network name has not resolved. ${reportLogMessage}`;
+            case '112':
+                return `The disk is full. Please free up space and try again.`;
+            case '255':
+                return `The .NET Installer was terminated by another process unexpectedly. Please try again.`;
+            case '1260':
+                return `The .NET SDK is blocked by group policy. Can you please report this at https://github.com/dotnet/vscode-dotnet-runtime/issues`
+            case '1603':
+                return `A peculiar system error prevented the .NET SDK from installing. ${reportLogMessage}`;
+            case '1618':
+                return `Another installation is in progress. Please wait for the other installation to finish and try again.`;
+            case '000751':
+                return `An unknown error occurred. ${reportLogMessage}`;
+            case '2147500037':
+                return `An unspecified error occurred. ${reportLogMessage}`;
+            case '2147942405':
+                return `Insufficient permissions are available to install .NET. Please try again as an administrator.`;
+        }
+        return '';
+    }
+
     public async installSDK(install : DotnetInstall): Promise<string>
     {
         // Check for conflicting windows installs
--- a/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
@@ -103,7 +103,7 @@ This report should be made at https://github.com/dotnet/vscode-dotnet-runtime/is
             case '1603':
                 return `Fatal error during .NET SDK installation. ${reportLogMessage}`;
             case '1618':
-                return `Another installation is in progress. Please wait for the other installation to finish and try again.`;
+                return `Another installation is already in progress. Complete that installation before proceeding with this install.`;
             case '000751':
                 return `An unknown error occurred, which might be an OS page fault. ${reportLogMessage}`;
             case '2147500037':
--- a/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
@@ -103,7 +103,7 @@ This report should be made at https://github.com/dotnet/vscode-dotnet-runtime/is
             case '1618':
                 return `Another installation is in progress. Please wait for the other installation to finish and try again.`;
             case '000751':
-                return `An unknown error occurred. ${reportLogMessage}`;
+                return `An unknown error occurred, which might be an OS page fault. ${reportLogMessage}`;
             case '2147500037':
                 return `An unspecified error occurred. ${reportLogMessage}`;
             case '2147942405':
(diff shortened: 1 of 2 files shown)


vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -78,6 +78,40 @@ export class WinMacGlobalInstaller extends IGlobalInstaller {
...
+            case '67':
+                return `A network name has not resolved. ${reportLogMessage}`;
+            case '112':
+                return `The disk is full. Please free up space and try again.`;
+            case '255':
+                return `The .NET Installer was terminated by another process unexpectedly. Please try again.`;
+            case '1260':
+                return `The .NET SDK is blocked by group policy. Can you please report this at https://github.com/dotnet/vscode-dotnet-runtime/issues`
+            case '1603':
+                return `A peculiar system error prevented the .NET SDK from installing. ${reportLogMessage}`;
```
joeloff, line 104: For anything that does look like an HRESULT/LSTATUS code I'd suggest using the official description, e.g. for 1603. "Fatal error during installation"

https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--1300-1699-

nagilson, line 104:
```suggestion
                return `Fatal error during .NET SDK installation. ${reportLogMessage}`;
```

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -78,6 +78,40 @@ export class WinMacGlobalInstaller extends IGlobalInstaller {
...
+            case '1618':
+                return `Another installation is in progress. Please wait for the other installation to finish and try again.`;
```
nagilson, line 106:
```suggestion
                return `Another installation is already in progress. Complete that installation before proceeding with this install.`;
```

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -78,6 +78,42 @@ export class WinMacGlobalInstaller extends IGlobalInstaller {
...
+            case '1260':
+                return `The .NET SDK is blocked by group policy. Can you please report this at https://github.com/dotnet/vscode-dotnet-runtime/issues`
+            case '1460':
+                return `The .NET SDK had a timeout error. ${reportLogMessage}`;
+            case '1603':
+                return `Fatal error during .NET SDK installation. ${reportLogMessage}`;
+            case '1618':
+                return `Another installation is already in progress. Complete that installation before proceeding with this install.`;
+            case '000751':
+                return `An unknown error occurred, which might be an OS page fault. ${reportLogMessage}`;
```
nagilson, line 108:
```suggestion
                return `Page fault was satisfied by reading from a secondary storage device. ${reportLogMessage}`;
```

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
nagilson: I didnt know about this website, thank you. :)

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -78,6 +78,42 @@ export class WinMacGlobalInstaller extends IGlobalInstaller {
...
+This report should be made at https://github.com/dotnet/vscode-dotnet-runtime/issues.`
+
+        switch(code)
+        {
+            case '1':
+                return `The .NET SDK installer has failed with a generic failure. ${reportLogMessage}`;
+            case '5':
+                return `Insufficient permissions are available to install .NET. Please run the installer as an administrator.`;
+            case '67':
+                return `A network name has not resolved. ${reportLogMessage}`;
```
nagilson, line 94:
```suggestion
                return `The network name cannot be found. ${reportLogMessage}`;
```

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
nagilson: I updated most of the strings. For a few, I think our string is more actionable and precise to our circumstances, so I kept them.

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -78,6 +78,42 @@ export class WinMacGlobalInstaller extends IGlobalInstaller {
         this.webWorker = new WebRequestWorker(context, installerUrl);
     }
 
+    public static InterpretExitCode(code : string) : string
+    {
+        const reportLogMessage = `Please provide your .NET Installer log (note our privacy notice), which can be found at %temp%.
+The file has a name like 'Microsoft_.NET_SDK*.log and should appear in recent files.
```
smitpatel, line 84: Is `Microsoft_.NET_SDK` is correct?

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -78,6 +78,42 @@ export class WinMacGlobalInstaller extends IGlobalInstaller {
...
+            case '5':
+                return `Insufficient permissions are available to install .NET. Please run the installer as an administrator.`;
+            case '67':
+                return `The network name cannot be found. ${reportLogMessage}`;
+            case '112':
+                return `The disk is full. Please free up space and try again.`;
+            case '255':
+                return `The .NET Installer was terminated by another process unexpectedly. Please try again.`;
+            case '1260':
+                return `The .NET SDK is blocked by group policy. Can you please report this at https://github.com/dotnet/vscode-dotnet-runtime/issues`
```
smitpatel, line 100: Should this be using reportLogMessage string instead?

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -78,6 +78,42 @@ export class WinMacGlobalInstaller extends IGlobalInstaller {
...
+            case '1603':
+                return `Fatal error during .NET SDK installation. ${reportLogMessage}`;
+            case '1618':
+                return `Another installation is already in progress. Complete that installation before proceeding with this install.`;
+            case '000751':
+                return `Page fault was satisfied by reading from a secondary storage device. ${reportLogMessage}`;
+            case '2147500037':
+                return `An unspecified error occurred. ${reportLogMessage}`;
+            case '2147942405':
+                return `Insufficient permissions are available to install .NET. Please try again as an administrator.`;
```
smitpatel, line 112: this & 5 is same. May be just merge or use goto if you want number ordering.

smitpatel (approved):

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
nagilson: Yes! 
image

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
nagilson: That's a good point, though there's nothing we'd gain from the installer log in that case, so I just used this string.

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
nagilson: That's a good idea, if there were more instances of it I would use `goto`, in this case since it's only 1 time I don't feel too inspired to fix it 😄

PR #1881, opened 2024-07-15
Allow elevation through non-linux commands

nagilson:
We use @vscode/sudo-prompt to elevate. This way we don't need to handle the user password.

As discussed here https://github.com/dotnet/vscode-dotnet-runtime/pull/1620 Sudo-prompt does not store elevation permission, so every time it makes a request, it makes a new password request. For linux this is extremely annoying so there is a ton of extra setup I do to prevent this requirement.

For windows and mac, we can just call into the API directly since we dont need to elevate multiple times, it's less annoying.
This codepath was never used before (`runUnderSudo` only ever provided on linux.)

image
The prompt looks like this on Mac. 
image
Windows.

There isn't much we can do to change that, since that's part of vscode's library.

On Windows, sometimes we will fail to run due to denied permissions, we can at least retry by launching the application with more permissions the first time.

For Mac and Windows when uninstalling, for some uninstallation methods, we need to be elevated from the beginning, otherwise the install will immediately fail. Example. https://github.com/dotnet/cli-lab/blob/ffb7b78931370c04f6b2a045cf8e79d299f66286/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs#L41 

This PR will enable future code to elevate. Namely, https://github.com/dotnet/vscode-dotnet-runtime/pull/1897 cannot work without this because it needs to run `sudo rm`.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
@@ -368,7 +368,7 @@ with options ${JSON.stringify(options)}.`));
 
             let commandResult;
 
-            if(os.platform() === 'win32' && command.runUnderSudo)
+            if(command.runUnderSudo)
             {
                 execElevated(fullCommandString, options, (error?: any, execStdout?: any, execStderr?: any) =>
                 {
--- a/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
@@ -370,13 +370,13 @@ with options ${JSON.stringify(options)}.`));
 
             if(os.platform() === 'win32' && command.runUnderSudo)
             {
-                execElevated(fullCommandString, options, (error?: any, stdout?: any, stderr?: any) =>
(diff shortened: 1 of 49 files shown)


vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -125,13 +125,11 @@ Please install the .NET SDK manually by following https://learn.microsoft.com/en
         this.context?.eventStream.post(new CommandExecutionUserAskDialogueEvent(`Prompting user for command ${fullCommandString} under sudo.`));
 
         // The '.' character is not allowed for sudo-prompt so we use 'NET'
```
smitpatel, line 127: Move this comment around line 557

smitpatel (approved):

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
nagilson: Good suggestion, I will do that here: https://github.com/dotnet/vscode-dotnet-runtime/pull/1897

PR #1882, opened 2024-07-16
Remove references to the string 'key' and change them to 'id'

nagilson:
VS Code has a filter for the string 'key' -- this causes our telemetry to be dropped when referencing the term 'install key' as it thinks it may be a secret. https://github.com/microsoft/vscode/blob/48ba01b5896376718aa66a11c60305cb105815c0/src/vs/platform/telemetry/common/telemetryUtils.ts#L336

Id is safe. I talked with the vscode team and there is no good alternative besides turning off all telemetry safety scanning, which I didn't want to do.

Hence, 'key' --> 'id'

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
@@ -52,8 +52,8 @@ export class InstallTrackerSingleton
     protected static instance: InstallTrackerSingleton;
 
     protected inProgressInstalls: Set<InProgressInstall> = new Set<InProgressInstall>();
-    private readonly installingVersionsKey = 'installing';
-    private readonly installedVersionsKey = 'installed';
+    private readonly installingVersionsid = 'installing';
+    private readonly installedVersionsid = 'installed';
 
     protected constructor(protected eventStream : IEventStream, protected extensionState : IExtensionState)
     {
@@ -124,16 +124,16 @@ export class InstallTrackerSingleton
 
     /**
      *
-     * @param key the install key to get a working install promise for.
+     * @param id the install id to get a working install promise for.
      */
-    public getPromise(key : DotnetInstall) : Promise<string> | null
+    public getPromise(id : DotnetInstall) : Promise<string> | null
     {
             this.inProgressInstalls.forEach(x =>
             {
-                const xAsKey = x.dotnetInstall as DotnetInstall;
-                if(IsEquivalentInstallationFile(xAsKey, key))
+                const xAsid = x.dotnetInstall as DotnetInstall;
+                if(IsEquivalentInstallationFile(xAsid, id))
                 {
-                    this.eventStream.post(new DotnetAcquisitionStatusResolved(key, key.version));
+                    this.eventStream.post(new DotnetAcquisitionStatusResolved(id, id.version));
                     return x.installingPromise;
                 }
             })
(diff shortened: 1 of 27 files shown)


vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -124,57 +124,57 @@ export class InstallTrackerSingleton
 
     /**
      *
-     * @param key the install key to get a working install promise for.
+     * @param id the install id to get a working install promise for.
      */
-    public getPromise(key : DotnetInstall) : Promise<string> | null
+    public getPromise(id : DotnetInstall) : Promise<string> | null
```
smitpatel, line 129: same better parameter name

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -226,21 +226,21 @@ ${convertedInstalls.map(x => `${JSON.stringify(x.dotnetInstall)} owned by ${x.in
...
+    protected async removeVersionFromExtensionState(context : IAcquisitionWorkerContext, idStr: string, installIdObj: DotnetInstall)
     {
-        return this.executeWithLock( false, async (key: string, install: DotnetInstall) =>
+        return this.executeWithLock( false, async (id: string, install: DotnetInstall) =>
         {
-            this.eventStream.post(new RemovingVersionFromExtensionState(`Removing ${JSON.stringify(install)} with key ${key} from the state.`));
-            const existingInstalls = await this.getExistingInstalls(key === this.installedVersionsKey, true);
+            this.eventStream.post(new RemovingVersionFromExtensionState(`Removing ${JSON.stringify(install)} with id ${id} from the state.`));
+            const existingInstalls = await this.getExistingInstalls(id === this.installedVersionsid, true);
             const installRecord = existingInstalls.filter(x => IsEquivalentInstallation(x.dotnetInstall, install));
```
smitpatel, line 244: installedVersionsId

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -315,38 +315,38 @@ ${convertedInstalls.map(x => `${JSON.stringify(x.dotnetInstall)} owned by ${x.in
...
     }
 
-    public async checkForUnrecordedLocalSDKSuccessfulInstall(context : IAcquisitionWorkerContext, dotnetInstallDirectory: string, installedInstallKeysList: InstallRecord[]): Promise<InstallRecord[]>
+    public async checkForUnrecordedLocalSDKSuccessfulInstall(context : IAcquisitionWorkerContext, dotnetInstallDirectory: string, installedInstallIdsList: InstallRecord[]): Promise<InstallRecord[]>
     {
-        return this.executeWithLock( false, async (dotnetInstallDir: string, installedInstallKeys: InstallRecord[]) =>
+        return this.executeWithLock( false, async (dotnetInstallDir: string, installedInstallIds: InstallRecord[]) =>
         {
-            let localSDKDirectoryKeyIter = '';
+            let localSDKDirectoryidIter = '';
```
smitpatel, line 328: LocalSDKDirectoryIdIter

smitpatel (approved):
Since this touch telemetry, also check if you need to classify the new property.

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
nagilson: Thanks for your approval, looks like I forgot to do a case-sensitive search on this file. I also improved some of the names in other spots where it looks like I missed it.

You're right in that I will need to go into the GDPR tool and reclassify the events and update some of our queries as well.

smitpatel (approved):

PR #1884, opened 2024-07-16
Try to find the .NET SDK PATH on Mac if it Installs to Unexpected Loc

nagilson:
Our highest failure for .NET SDK Installs on Mac only is that the `InstallationValidator` which checks the hard-coded expected path of Installation does not find .NET in the expected location. Unfortunately, nobody has reported this so we can't get more info from them.

To get to this point in the code of `Validation`, everything has to have succeeded, including the installer only returning and exiting 0, which it would only do upon installation success.

I'm guessing there is some undocumented case where the .NET SDK is installed to a different location, but I couldn't find where.

Here is our path check. https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts#L239

According to the installer source code and our documentation, it should only install to this location or with x64 during emulation.
```
<PkgInstallDirectory>/usr/local/share/dotnet</PkgInstallDirectory>
      <x64EmulationPkgInstallDirectory>/usr/local/share/dotnet/x64</x64EmulationPkgInstallDirectory>
```

The 64 directory however, is only for when you install a 64-bit SDK on an ARM Mac. We wouldn't expect this to happen because we only install the SDK that matches os.platform(), so we should never install an x64 SDK on an ARM machine.

I'm curious to see where this path is and this will also help us track that. It's a bit of a hack, but I think it will have the desired outcome.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
@@ -31,6 +31,7 @@ import {
     SuppressedAcquisitionError,
     EventBasedError,
     EventCancellationError,
+    DotnetInstallationValidated,
 } from '../EventStream/EventStreamEvents';
 
 import { GlobalInstallerResolver } from './GlobalInstallerResolver';
@@ -61,6 +62,7 @@ import { DotnetInstallMode } from './DotnetInstallMode';
 import { IEventStream } from '../EventStream/EventStream';
 import { IExtensionState } from '../IExtensionState';
 import { getInstallKeyCustomArchitecture } from '../Utils/InstallKeyUtilities';
+import { CommandExecutor } from '../Utils/CommandExecutor';
 
 /* tslint:disable:no-any */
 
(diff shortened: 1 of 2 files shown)


vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -404,12 +406,35 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
...
         TelemetryUtilities.setDotnetSDKTelemetryToMatch(context.isExtensionTelemetryInitiallyEnabled, this.extensionContext, context, this.utilityContext);
 
-        context.installationValidator.validateDotnetInstall(install, installedSDKPath, os.platform() !== 'win32');
+        try
+        {
+            context.installationValidator.validateDotnetInstall(install, installedSDKPath, os.platform() !== 'win32');
+        }
+        catch(error : any)
+        {
+            if(install.isGlobal && os.platform() === 'darwin')
```
nagilson, line 420:
```suggestion
            if(os.platform() === 'darwin')
```

smitpatel (approved):

nagilson:
btw, we dont use --list-sdks to verify the SDK because we dont want to trigger the 1st run sentinel if the user hasn't used the SDK yet.

PR #1891, opened 2024-07-18
Separate 1 Lock into 2 Mutexes/Locks for Data Management

nagilson:
'installing' and 'installed' data that is stored in the extension state is preserved between runs. We allow concurrent requests. Sometimes, a request may come in while another request is currently processing. Therefore, we have a lock guard on the install request tracking, so for example, if two identical requests come in at the same time, there is no race condition where the request work gets duplicated.

The promise chain which stores currently in progress installs used to be held behind a lock as well but this doesn't make sense because it's a local data structure, operations on it should be atomic.

We can separate the lock shared for all of these things into 2 separate locks since their operations are discrete from one another to improve performance and stability.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -82,18 +82,3 @@ popd
 & "$(Split-Path $MyInvocation.MyCommand.Path)/mock-webpack.ps1"
 
 Write-Host "Build Succeeded" -ForegroundColor $successColor
-
-
-#################### Allow Local Signing ####################
-
-$InstallNuGetPkgScriptPath = "$PSScriptRoot\msbuild\Install-NuGetPackage.ps1"
-$nugetVerbosity = 'quiet'
-if ($Verbose) { $nugetVerbosity = 'normal' }
-$MicroBuildPackageSource = 'https://pkgs.dev.azure.com/devdiv/_packaging/MicroBuildToolset%40Local/nuget/v3/index.json'
-if ($Signing) {
-    Write-Host "Installing MicroBuild signing plugin" -ForegroundColor $HeaderColor
-    & $InstallNuGetPkgScriptPath MicroBuild.Plugins.Signing -source $MicroBuildPackageSource -Verbosity $nugetVerbosity
-    $EnvVars['SignType'] = "Test"
-}
-
-& "$PSScriptRoot/msbuild/Set-EnvVars.ps1" -Variables $EnvVars -PrependPath $PrependPath | Out-Null
--- a/build.ps1
+++ b/build.ps1
@@ -82,3 +82,18 @@ popd
 & "$(Split-Path $MyInvocation.MyCommand.Path)/mock-webpack.ps1"
 
 Write-Host "Build Succeeded" -ForegroundColor $successColor
+
+
+#################### Allow Local Signing ####################
+
+$InstallNuGetPkgScriptPath = "$PSScriptRoot\msbuild\Install-NuGetPackage.ps1"
+$nugetVerbosity = 'quiet'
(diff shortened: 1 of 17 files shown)


build.ps1
```diff
@@ -81,4 +81,19 @@ popd
...
+$nugetVerbosity = 'quiet'
+if ($Verbose) { $nugetVerbosity = 'normal' }
+$MicroBuildPackageSource = 'https://pkgs.dev.azure.com/devdiv/_packaging/MicroBuildToolset%40Local/nuget/v3/index.json'
+if ($Signing) {
+    Write-Host "Installing MicroBuild signing plugin" -ForegroundColor $HeaderColor
+    & $InstallNuGetPkgScriptPath MicroBuild.Plugins.Signing -source $MicroBuildPackageSource -Verbosity $nugetVerbosity
+    $EnvVars['SignType'] = "Test"
+}
+
+& "$PSScriptRoot/msbuild/Set-EnvVars.ps1" -Variables $EnvVars -PrependPath $PrependPath | Out-Null
```
nagilson, line 99:
```suggestion
```

LLotme (commented):
gh pr checkout 1891

officialmofabs (approved):

officialmofabs (approved):

officialmofabs (approved):

LLotme (commented):
gh pr checkout 1891

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -120,7 +120,7 @@ export class InstallTrackerSingleton
 
     public async clearPromises() : Promise<void>
     {
-        await this.executeWithLock( false, () => {this.inProgressInstalls.clear();});
+        this.inProgressInstalls.clear();
```
smitpatel, line 123: I noticed quite a few functions in this file which are async but have no async code inside it like this one.

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -185,7 +182,8 @@ Installs: ${[...this.inProgressInstalls].map(x => x.dotnetInstall.installId).joi
      */
     public async getExistingInstalls(getAlreadyInstalledVersion : boolean, alreadyHoldingLock = false) : Promise<InstallRecord[]>
     {
-        return this.executeWithLock( alreadyHoldingLock, (getAlreadyInstalledVersions : boolean) =>
+        return this.executeWithLock( alreadyHoldingLock, getAlreadyInstalledVersion ? this.installedVersionsId : this.installingVersionsId,
+            (getAlreadyInstalledVersions : boolean) =>
```
smitpatel, line 186: nit: passing in string data key here would be better as the outer function will calculate it based on bool value.

smitpatel (approved):

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
nagilson: ah yes, good point. these no longer need to be async with this change.

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
nagilson: I think that's a good point but this interface is also used in a public fashion and I don't want outer callers to be aware of the string data key. If you disagree, please let me know.

PR #1894, opened 2024-07-24
Add API to uninstall just one runtime

nagilson:
The existing API surface only allowed uninstallation of ALL runtimes.
This `uninstall` API currently only works with local installs but it will be expanded to allow global ones.
We can enforce more strict checking in the api caller request since it's not going to break anything.

`force` was added as a later option for when we allow users to force override and uninstall a version even if other extensions still depend on the installation.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/extension.ts
+++ b/vscode-dotnet-runtime-extension/src/extension.ts
@@ -53,6 +53,7 @@ import {
     UserManualInstallVersionChosen,
     UserManualInstallRequested,
     UserManualInstallSuccess,
+    InvalidUninstallRequest,
     UserManualInstallFailure,
     DotnetInstall,
     EventCancellationError,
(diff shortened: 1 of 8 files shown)


vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -336,6 +338,52 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
...
 
+    /**
+     * @returns 0 on success. Error string if not.
+     */
+    const dotnetUninstallRegistration = vscode.commands.registerCommand(`${commandPrefix}.${commandKeys.uninstall}`, async (commandContext: IDotnetAcquireContext | undefined) =>
+    {
+        return uninstall(commandContext);
+    });
+
+    async function uninstall(commandContext: IDotnetAcquireContext | undefined) : Promise<string>
```
smitpatel, line 349: nit: potentially moving all command processor to separate file may improve readability of extension.

smitpatel (approved):

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: Thanks for your review :) I think this is a good suggestion but probably worth it's own PR, so I won't do this here.

PR #1897, opened 2024-07-30
Allow uninstallation of global .NET Installations

nagilson:
# Global Uninstall

Allows uninstall of global .NET SDKs.

For windows, use /uninstall flag on the installer.
For mac, use sudo rm rf as /uninstall flag DNE on mac
For linux, use apt-get remove. (pre-existing code with pre-existing tests)

# Other Code Changes

* These are needed to make uninstall work successfully

Sets a `DEBIANT_FRONTEND` environment variable temporarily in the shell script process to prevent apt-get remove from trying to grab stdin which may be busy.

Adds an `exit.txt` variable to tell the `master sudo` process to die. Sometimes this process will hold onto the lock after install/uninstall and then this causes problems. Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1904

Adds `-rf` to the final process trap exit to prevent failure due to file DNE if it's already been cleaned up before the process dies.

Adds pre-existing images which used to be ignored for some reason.

Uses VS Code Icon Pack to be compliant with iconography restrictions and also keep icons updated per API instead of trying to invoke win-32 API to get the UAC icon.

Increase DPKG lock timeout to 180 seconds as it can sometimes take longer.

Adds -y to some commands which may be necessary with DEBIAN_FRONTEND.

# UI 
Adds a UI to uninstall a local or global .NET installation from the user.

Base command:
image

Picking an installation:
image

Has warning menu if you try to uninstall a .NET version which will break an extension which relies on that installation.
image

# Uninstall Experience

Windows, non-elevated:
image

This output window appears. If not-elevated, UAC prompt will appear, no further intervention is required.

image
(this text is old, it now just says uninstalled)

Mac: 
Password prompt will appear to elevate to uninstall.
image

Linux:
image

If no installs exist:
Context menu will show no installs.
image


vscode-dotnet-runtime-extension/package.json
```diff
@@ -51,6 +51,11 @@
 				"title": "Install the .NET SDK System-Wide.",
 				"category": ".NET Install Tool",
 				"enablement": "!dotnetAcquisitionExtension.isGlobalSDKUnsupported"
+			},
+            {
+				"command": "dotnet.uninstallPublic",
+				"title": "Uninstall .NET.",
```
smitpatel, line 57: Commands in command palette don't end with period. Something needs to be fixed in all of above.
Also titles are not getting localized?

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -75,11 +79,13 @@ namespace configKeys {
     export const existingSharedPath = 'sharedExistingDotnetPath'
     export const proxyUrl = 'proxyUrl';
 }
+
 namespace commandKeys {
     export const acquire = 'acquire';
     export const acquireGlobalSDK = 'acquireGlobalSDK';
     export const acquireStatus = 'acquireStatus';
     export const uninstall = 'uninstall';
+    export const uninstallPublic = 'uninstallPublic'
```
smitpatel, line 88: what does public refer to?

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -338,6 +344,70 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
...
+        {
+            if(x.dotnetInstall.installMode === y.dotnetInstall.installMode)
+            {
+                return x.dotnetInstall.version.localeCompare(y.dotnetInstall.version);
+            }
+            return x.dotnetInstall.installMode.localeCompare(y.dotnetInstall.installMode);
+        })?.map(install =>
+        {
+            return {
+                label : `.NET ${(install.dotnetInstall.installMode).toUpperCase()} ${install.dotnetInstall.version}`,
```
smitpatel, line 362: Since installMode has only 3 values, use .NET styling rather than upper case.

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -338,6 +344,70 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
...
+                description : `${install.dotnetInstall.architecture ?? ''} | ${install.dotnetInstall.isGlobal ? 'machine-wide' : 'vscode-local' }`,
+                detail : `Used by ${install.installingExtensions.join(', ')}`,
+                iconPath : install.dotnetInstall.isGlobal ? new vscode.ThemeIcon('shield') : new vscode.ThemeIcon('trash'),
+                internalId : install.dotnetInstall.installId
+            }
+        });
+        const chosenVersion = await vscode.window.showQuickPick(menuItems, { placeHolder: 'Select a version to uninstall.' });
```
smitpatel, line 369: If there are no versions installed, show a notification when command is invoked rather than empty list.

vscode-dotnet-runtime-extension/package.json
nagilson: Nothing in this extension is localized atm. This is planned for the future 😱

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: public meaning users can use it and it shows up in the command palette

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: I think this is a valid behavior but I'm not sure I prefer this behavior. Is this how other uis in devkit work? I'm curious to hear what other's opinions are.

vscode-dotnet-runtime-extension/src/extension.ts
smitpatel: So what is opposite of it? Are there installations done by extension which user cannot see?

vscode-dotnet-runtime-extension/src/extension.ts
smitpatel: And empty quick pick is not intuitive to user that if the generating list for quick pick failed to work correctly by extension or there are no items to display. You can choose to show it whichever way you want but you should notify user that there are no installations.

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: There are about 10 other commands / apis here that are non 'public', by this it means that only other extensions calling our extension apis can use them. They are not labeled with 'private'

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: Thank you, I think this is a good idea.

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: That's a good point. I'll follow this suggested change. Thanks for looking at this PR!

smitpatel (approved):

Issue #1858, opened 2024-07-01
Add API option to check version matches less strictly


nagilson:
### Is your feature request related to a problem? Please describe.

When making an API call and a generic version is given such as 8.0 we must resolve this using the server to a more specific version. We then check our existing installs to see if that version is available.

### A clear and concise description of what you want to happen. Include any alternative solutions you've considered.

This would be part of `IDotnetAcquireContext` as the option boolean, which is undefined/false by default, as `allowAlternativeVersions`.

### Describe the solution you'd like

If the user is offline but had installed .NET successfully before then we should not fail. The best way to do this would be to add an optional flag to our API (to prevent breaking others who may rely on this expectation of only accepting the latest .NET install) that would skip the online check for versions and just accept any version as a match that has the same major.minor.

nagilson:
See https://github.com/microsoft/vscode-dotnettools/issues/1028

PR #1898, opened 2024-07-30
Don't fail if the user is offline

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1866
https://github.com/dotnet/vscode-dotnet-runtime/issues/1858 (does not add api, we decide to do this automatically)

The `VersionResolver` class makes a web-request before downloading .NET to find the newest version of .NET to keep users up to date. The problem is, this resolution of which version of .NET to download occurs before when we check the pre-existing installations to see if we need to do an installation at all, because we need to resolve a version string such as '8.0' to '8.0.2'. 

If the user had downloaded the .NET before hand then went on an airplane, which many have, it would not check if that install exists because it'd first see if there is a newer version to download then fail. 

This adds a check where if we are offline we check the preexisting installs first to prevent this failure from occurring.

# Previous Behavior:

It would wait for the timeout silently and then fail.
See users reporting this in https://github.com/microsoft/vscode-dotnettools/issues/1028 which this resolves. 

# New Behavior:

image

For the first 2 requests, I request '7.0' which was installed previously. It uses the old existing install and remarks to update when possible. 

For the next request, I request '8.0' which was not installed previously. It has a new warning message instead of hanging and will wait..

When connecting back to the internet, it works again.


vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -176,13 +179,28 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
                 return existingPath;
             }
 
+            if(!(await WebRequestWorker.isOnline(timeoutValue ?? defaultTimeoutValue, globalEventStream)))
+            {
+                workerContext.acquisitionContext.architecture = workerContext.acquisitionContext.architecture ?? DotnetCoreAcquisitionWorker.defaultArchitecture();
```
smitpatel, line 184: Can use `??=`

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -231,20 +249,29 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
...
             }
 
-            const globalInstallerResolver = new GlobalInstallerResolver(sdkContext, commandContext.version);
+            if(!(await WebRequestWorker.isOnline(timeoutValue ?? defaultTimeoutValue, globalEventStream)))
+            {
+                const existingPathOffline = await worker.getSimilarExistingInstall(workerContext);
+                if(existingPathOffline?.dotnetPath)
+                {
+                    return Promise.resolve(existingPathOffline);
+                }
```
smitpatel, line 258: This does not log message like the other?

vscode-dotnet-runtime-library/src/Acquisition/VersionUtilities.ts
```diff
@@ -0,0 +1,165 @@
...
+        const event = new DotnetFeatureBandDoesNotExistError(new EventCancellationError('DotnetFeatureBandDoesNotExistError',
+            `${invalidFeatureBandErrorString}${fullySpecifiedVersion}.`),
+            getInstallFromContext(context));
+        eventStream.post(event);
+        throw event.error;
+    }
+    return patch
+}
+
+
```
smitpatel, line 92: extra line break

smitpatel (approved):
Please split the function to handle command in different files. It is making harder to review changes in extension.ts file as it is hard to find which function is getting modified.

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: omg thank you I forgot that operator existed 🤯

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: yikes it also doesnt set the architecture lemme refactor it out into a function

nagilson: good catch thank you!!

nagilson:
Thanks for your review, you caught some important issues. For the refactor, I've made a separate issue to track this so it can be done in another PR: https://github.com/dotnet/vscode-dotnet-runtime/issues/1917

PR #1914, opened 2024-08-13
Fix parsing `-preview` versions in the registry when installed via Visual Studio

nagilson:
This issue is pointed out to me by @webreidi and became visible today.

Internal Visual Studio build 17.11.04 shipped with a preview `SDK 8,0.400-preview.0`.
This does not impact customers but impacts internal people who try to install using our tool.
Our logic worked for preview SDKs installed by users but not the ones installed by Visual Studio.

When opening up the registry to check if we needed to do this installation, we would see the version as `8.0.400-preview.0` and try to cast `400-preview` to a number. This would fail and we would not install the SDK.

We can fix this in several ways. One way is to improve some logic around parsing the preview text, but that text is not necessary to handle. So instead, we can just trim out the -preview text and cause it to work.

Before 
image

After
image

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/VersionResolver.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/VersionResolver.ts
@@ -255,7 +255,7 @@ export class VersionResolver implements IVersionResolver {
      */
     private getPatchVersionString(fullySpecifiedVersion : string) : string
     {
-        const patch : string | undefined = fullySpecifiedVersion.split('.')?.at(2)?.substring(1).split('-')?.at(0);
+        const patch : string | undefined = fullySpecifiedVersion.split('.')?.at(2)?.substring(1)?.split('-')?.at(0);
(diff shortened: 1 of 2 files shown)


vscode-dotnet-runtime-library/src/Acquisition/VersionResolver.ts
```diff
@@ -255,7 +255,7 @@ export class VersionResolver implements IVersionResolver {
      */
     private getPatchVersionString(fullySpecifiedVersion : string) : string
     {
-        const patch : string | undefined = fullySpecifiedVersion.split('.')?.at(2)?.substring(1);
+        const patch : string | undefined = fullySpecifiedVersion.split('.')?.at(2)?.substring(1).split('-')?.at(0);
```
nagilson, line 258:
```suggestion
        const patch : string | undefined = fullySpecifiedVersion.split('.')?.at(2)?.substring(1)?.split('-')?.at(0);
```

smitpatel (approved):

Issue #1916, opened 2024-08-15
git commit Author identity unknown


Jinying-Tang:
### Describe the bug.

When I change the code in vscode and click the submit button on the git source code management page, it prompts to ensure that a username and email are configured in git. The log shows:
2024-08-09 00:53:36.781 [info] > git -c user.useConfigOnly=true commit --quiet --allow-empty-message --file - [129ms]
2024-08-09 00:53:36.781 [info] Author identity unknown
*** Please tell me who you are.

However, I can see that I have already set up a username and email using git config user.name and git config user.email in the terminal, and changes can be committed through git commit prompt in the terminal.

### Please attach log file(s) here if you have any.

2024-08-15 09:21:41.177 [info] > git -c user.useConfigOnly=true commit --quiet --allow-empty-message --file - [137ms]
2024-08-15 09:21:41.177 [info] Author identity unknown

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: no email was given and auto-detection is disabled
2024-08-15 09:21:41.284 [info] > git config --get-all user.name [102ms]
2024-08-15 09:21:41.295 [info] [Git][getRemotes] No remotes found in the git config file
2024-08-15 09:21:41.404 [info] > git config --get commit.template [114ms]
2024-08-15 09:21:41.414 [info] > git for-each-ref --format=%(refname)%00%(upstream:short)%00%(objectname)%00%(upstream:track)%00%(upstream:remotename)%00%(upstream:remoteref) --ignore-case refs/heads/develop refs/remotes/develop [119ms]
2024-08-15 09:21:41.556 [info] > git status -z -uall [137ms]
2024-08-15 09:21:41.670 [info] > git for-each-ref --format=%(refname)%00%(upstream:short)%00%(objectname)%00%(upstream:track)%00%(upstream:remotename)%00%(upstream:remoteref) --ignore-case refs/heads/develop refs/remotes/develop [110ms]
2024-08-15 09:21:41.865 [info] > git config --local branch.develop.vscode-merge-base [190ms]
2024-08-15 09:21:41.865 [warning] [Git][config] git config failed: Failed to execute git
2024-08-15 09:21:41.977 [info] > git reflog develop --grep-reflog=branch: Created from *. [106ms]
[...]

Issue #1795, opened 2024-05-07
gitignore not working after enabling the extension


dpalmetzvbg:
### Describe the bug.

After enabling the extension the global gitignore-file, which worked properly before, does not work anymore. There are many untracked files, that should not be tracked and are listed in the global gitignore-file. Also the python-language server shows some errors in the logfile.
git itsself does not show any errors in the logfiles. After disabling the extension the gitignorefile works properly again.

I am using a .env-File for my python environment.

### Please attach log file(s) here if you have any.

Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

open vscode
open the project folder
enable the extension
open the git-panel

### Expected Behavior

The gitignore-file should work properly and there should not be additional untracked files.

### .NET Install Tool Version

2.0.3

### Your operating system? Please provide more detail, such as the version and or distro above.

Windows

### VS Code Version

1.89.0

dpalmetzvbg:
I have fixed the python-error, but the gitignore problem still exists.

dpalmetzvbg:
It also seems to effect git in some other ways. When activated the extension I no longer can push commits to my remote repository, I get an error: permission denied (publickey). When disabling the extension, everything works fine again.

dpalmetzvbg:
I found the reason why this happens: I am in an centrally managed enterprise-environment, and my Home-Folder is, e.g., J:\ and not C:\Users\<myuser>.
So my .ssh, .gitconfig and .gitignore reside on J:\:
I have no environment-variable HOME set.
When enabling this extension and opening a shell for a pod, an environment-variable HOME is created, which points to C:\Users\<myuser>
This seems to cause the problem. When I move my files there, everything works fine. It also does not help to set an environment-variable "HOME" in windows and point it to J:\. The extension still creates its own HOME-Variable that points to C:\Users\<bkor>.
As I pointed out, I am in a centrally controlled enterprise-environemnt, and the configuration, where the ssh-files and git-configuration files are stored is not up to me. And always copying files to my user-folder is not a possible solution.

BrianL-STCU:
It looks like this may be the same issue as #1400, #1686, and #1707, which has been going on for most of a year. There are extensions that rely on this one that I just can't use if I want to also use the git integration.

[...]

Issue #1707, opened 2024-03-14
Version 2.0.2 caused that global config of VSC git client is ignored


josch83:
### Describe the bug.

It was a struggle to narrow down the cause. Suddenly the git client stopped working because the user information and other configurations such as SSL were no longer known. A check of the global git configuration in the terminal showed that the configuration was still there. Deactivating the VSC extensions one after the other showed that the ".NET Install Tool" was the cause. Rolling back to version 2.0.1 fixed my problems.

I have no idea how this could happen with this extension, which has no obvious relation to Git.

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

I did not check it on other environments, such I don't know if its reproducible.
1. Clone a git repo in a terminal. In my case one with an invalid SSL certificate. Ensure that user name an email are configured in global config. Disable SSL validation globally if applicable (and undo it after the testing, of course).
2. Open repo in VSC with extension installed (version 2.0.2)
3. Try to fetch (SSL validation error) or commit changes (missing user email)

### Expected Behavior

This extension should have no impact to git.

### .NET Install Tool Version

2.0.2

### Your operating system? Please provide more detail, such as the version and or distro above.

Windows

### VS Code Version

1.85.1 (user-setup)

josch83:
@meadowsjared reported a similar behaviour in isse #1686 : https://github.com/dotnet/vscode-dotnet-runtime/issues/1686#issuecomment-1979165653

Issue #1686, opened 2024-03-01
The .NET Install Tool causes that HOME environment variable does not work


Patrick-Ze:
Type: <b>Bug</b>

The issue occurs on the Windows platform:

1. Set the user environment variable HOME to a value different from the default user directory and restart VSCode. GPG key is also stored in this directory.
2. Install the .NET Install Tool extension `v2.0.2`.
3. Use the includeif function in .gitconfig, and reference the HOME directory with `~`, for example:
```gitconfig
[includeIf "hasconfig:remote.*.url:git@gitlab.mydomain.ai:*/**"]
	path = ~/gitconfig/mydomain.conf
```
4. Configure the GPG key in `mydomain.conf`.
```ini
[user]
	name = First Last
	email = first.last@mydomain.ai
	signingkey = AABBCCDDEEFF
[commit]
	gpgsign = true
```
6. Attempt to commit a change using the VSCode Source Control GUI.
7. Encounter an error:
```log
gpg: directory '/c/Users/first.last/.gnupg' created
gpg: keybox '/c/Users/first.last/.gnupg/pubring.kbx' created
gpg: skipped "First Last <first.last@mydomain.ai>": No secret key
gpg: signing failed: No secret key
error: gpg failed to sign the data
fatal: failed to write commit object
```
9. Note: if I commit in the integrated terminal instead of source control panel, everything works.
10. Disable this extension, then the problem is gone.

Extension version: 2.0.2
VS Code version: Code 1.87.0 (019f4d1419fbc8219a181fab7892ebccf7ee29a2, 2024-02-27T23:41:44.469Z)
OS version: Windows_NT x64 10.0.19045
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz (12 x 2592)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled|
|Load (avg)|undefined|
[...]

PR #1923, opened 2024-08-16
Don't set HOME environment variable

nagilson:
On Activation our Extension sets the HOME variable causing Git to no longer function.

Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1795 
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1400
Resolves  https://github.com/dotnet/vscode-dotnet-runtime/issues/1686 
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1707
Resolves  https://github.com/dotnet/vscode-dotnet-runtime/issues/1916

This import statement in our code on `WebRequestWorker.ts` sets HOME.

```js
const get_proxy_settings_1 = require("get-proxy-settings");
```

How do I know? I used the SysInternals Process Explorer to read the environment for every vscode sub process and went through the decompiled JS line by line in a binary search to see which line of code causes HOME to change 🥱 

image

HOME is no longer changed.

This is definitely one of the weirdest bugs Ive had to solve.... seriously. We checked the source code this library and they dont edit HOME either, probably their 1 of dependencies does. But it doesn't really matter.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
@@ -5,7 +5,18 @@
 import Axios, { AxiosError, isAxiosError } from 'axios';
 import axiosRetry from 'axios-retry';
 import { HttpsProxyAgent } from 'https-proxy-agent';
+
+process.env.VSCODE_DOTNET_INSTALL_TOOL_ORIGINAL_HOME = process.env.HOME
+// IMPORTING THIS LIBRARY SETS 'HOME' VARIABLE
+// Causing Git to BREAK!
 import { getProxySettings } from 'get-proxy-settings';
+// NODE JS CASTS UNDEFINED ENV VAR TO STRING 'undefined'
+process.env.HOME = process.env.VSCODE_DOTNET_INSTALL_TOOL_ORIGINAL_HOME === 'undefined' ? '' : process.env.VSCODE_DOTNET_INSTALL_TOOL_ORIGINAL_HOME
+if(process.env.HOME === '')
+{
+    delete process.env.HOME
+}
+
 import { AxiosCacheInstance, buildMemoryStorage, setupCache } from 'axios-cache-interceptor';
 import * as dns from 'dns';
 import * as fs from 'fs';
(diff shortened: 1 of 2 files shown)


vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -5,7 +5,18 @@
...
+process.env.VSCODE_DOTNET_INSTALL_TOOL_ORIGINAL_HOME = process.env.HOME
+// IMPORTING THIS LIBRARY SETS 'HOME' VARIABLE
+// Causing Git to BREAK!
 import { getProxySettings } from 'get-proxy-settings';
+// NODE JS CASTS UNDEFINED ENV VAR TO STRING 'undefined'
+process.env.HOME = process.env.VSCODE_DOTNET_INSTALL_TOOL_ORIGINAL_HOME === 'undefined' ? '' : process.env.VSCODE_DOTNET_INSTALL_TOOL_ORIGINAL_HOME
+if(process.env.HOME === '')
+{
+    delete process.env.HOME
+}
```
smitpatel, line 18: Would simplified form work?
```suggestion
if (process.env.VSCODE_DOTNET_INSTALL_TOOL_ORIGINAL_HOME === 'undefined')
{
    delete process.env.HOME
}
else
{
    process.env.HOME = process.env.VSCODE_DOTNET_INSTALL_TOOL_ORIGINAL_HOME
}
```

nagilson: yes, I was just being lazy :)

smitpatel (approved):

pbl-pw:
I searched the node_modules directory and found that npm-conf was writing the HOME environment variable.

dpalmetzvbg:
Thanks for investing your time and fixing it. Seems it was a difficult one... :-)

nagilson:
Thanks for reporting it :) and @pbl-pw for helping as well with the detailed report. It was more confusing/weird and time-consuming than it was difficult

nagilson:
There has been an issue on their repo since 2018 https://github.com/kevva/npm-conf/issues/13 💀  Yes it is npm-conf

PR #1930, opened 2024-08-20
Handle .NET Failed: Undefined

nagilson:
Rejecting a promise with a rejected promise as a return value from a function does not reject with the object the inner promise rejected with. Instead it rejects with an undefined value.

This means if the SDK installer fails or we fail to elevate, that counts as .NET Install failed: undefined. This fixes this. It also adds a check for the string error as this seems to happen sometimes if the password prompt is rejected.

How to test this? Set a break point in win mac global installer on the execute install and edit the command to point to a non existent file. It will then fail and you can see it works now.


nagilson:
Merging this so this can get out, some reviewers are OOF atm and I will be OOF soon.

PR #1932, opened 2024-08-20
Add PATH for Arm emulation x64 on Mac

nagilson:
Sometimes the PATH is not found even though the installer succeeds. This adds a check to see if it might be the other mac path that the installer can use. I believe this is the only other path it can use. https://github.com/dotnet/sdk/blob/a620dca0942ca7b9bcd7b6dac2b847e5b46afe7e/src/Installer/redist-installer/targets/GeneratePKG.targets#L17

I could not get my hands on a mac m1 that worked despite trial and tribulation (except for like buying one myself for this one bug) so this will add a check to see if it installed in this emulation path.

Theoretically it shouldnt be possible to have an architecture mismatch but I dont know what values os.platform and arch will return if its on an emulated architecture.


vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -349,11 +349,25 @@ If you were waiting for the install to succeed, please extend the timeout settin
...
+        {
+            if(fs.existsSync(standardHostPath) || !fs.existsSync(arm64EmulationHostPath))
+            {
+                return standardHostPath;
+            }
+            else
+            {
+                return arm64EmulationHostPath;
+            }
+        }
```
MiYanni, line 370: Isn't this just:
```
        if(!macPathShouldExist || fs.existsSync(standardHostPath) || !fs.existsSync(arm64EmulationHostPath))
        {
            return standardHostPath;
        }
        else
        {
            return arm64EmulationHostPath;
        }
```

And if that is the case, this would be simpler:
```suggestion
        return !macPathShouldExist ||
        fs.existsSync(standardHostPath) ||
        !fs.existsSync(arm64EmulationHostPath)) ?
            standardHostPath : 
            arm64EmulationHostPath;
```

Or you can make it one long line or whatever format you like.

MiYanni (approved):

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
nagilson: You're right, this is much better 😉  Thank you. Sometimes  I dont think about simplifying code in this way.

nagilson:
Waiting to see if I can get a mac to do this better or not.

nagilson:
Otherwise, this PR is ready to go.

nagilson:
Getting a mac has still taken a while. I am going to merge this

PR #1942, opened 2024-09-04
Leverage Managed Identity Publish

nagilson:
# Adds automated release

First, we need to make the pipeline publish to the marketplace again, like it used to here. We were told to remove this but later told that this was incorrect. This is the old code: https://github.com/sfoslund/vscode-dotnetcore/blob/ec7631ea6a24f4726ba546bd4e1827936c099e22/azure-pipelines.yml
But it doesn't quite work like it used to anymore.

# Adds Managed Identity 

Per requirements, we should switch to managed identity rather than marketplace tokens or individual uploads.
This implements said change. Here is an example of the C# Extension adding the same shared MI: https://github.com/dotnet/vscode-csharp/pull/7495/files_

Please see the internal branch for the newest run which includes this change working. It does not include an example of publishing to prod, which I will not test until we have a new release.

# Adds new manual validation step to release

image

# Changes icons for builds

This is so windows runs first because it is the agent that is used to then package and publish


pipeline-templates/publish.yaml
```diff
@@ -0,0 +1,35 @@
...
+  pool: ''
+  SignType: ''
+  version: ''
+
+jobs:
+  # Note: This step requires approval, allows for manual smoke testing the artifact before release
+  - deployment: Deploy
+    pool:
+      vmImage: ${{ parameters.pool.vmImage }}
+    displayName: '✨ Deploy to Release Environment'
```
nagilson, line 11: We should condition this to not happen on the public pipeline

pipeline-templates/publish.yaml
```diff
@@ -0,0 +1,60 @@
...
+  useOneEngineeringPool: ''
+
+jobs:
+  - job: waitForValidation
+    displayName: ☁️ Wait for release approval
+    pool: server
+    timeoutInMinutes: 4320 # job times out in 3 days
+    steps:
+    - task: ManualValidation@0
+      timeoutInMinutes: 4320
```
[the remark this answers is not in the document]
nagilson: Yes it is!

joeloff (approved):

nagilson:
Merging because I need to update the repo to the new names of the builds to actually pass the check and to do that I need to merge it

pipeline-templates/package-vsix.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
     name: GetVersion
     displayName: '❓ Get Version'
     workingDirectory: $(dir-name)
-  - ${{ if eq('${{ parameters.useOneEngineeringPool }}', 'true') }}:
+  - ${{ if eq(parameters.useOneEngineeringPool, 'true') }}:
```
MiYanni, line 45: I'm assuming this didn't work correctly before? AFAIK, you can't nest `${{ }}` inside `${{ }}`. It doesn't recursively evaluate like that.

nagilson: that is correct 🆗

PR #1944, opened 2024-09-09
Migrate to ESLint

nagilson:
# Summary 

TSLint was deprecated when this project had no maintainer. 
We waited a while to migrate because there was no real reason to. 

I followed https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration to help migrate.

Nowadays TSLint is failing a lot in our pipeline and having sporadic behavior. This might be due to upgrading to node 20. Or maybe due to the .editorconfig. I figured now was the time to migrate.

This introduced about 900 new linting errors. I fixed some of them. For files with no code churn, I disabled the linter for those files. If/when we touch those files again, we can fix those issues then. Some of them are very time consuming and don't make sense to test unless we are editing the pre-existing behavior because it's already a 'known-good.'

# Linting Issues 

One of the errors I did add a disable on a per line basis was unsafe member access. Unfortunately this only checks against calling `.` on the `any` type, not whether accessing `.foo` is safe to call (e.g. `x?.bar`  is OK, but not `x.bar` when `x` is `any` and could this be `undefined` or `null`.)

But I didnt want to disable it project wide because without the ? it is actually a serious problem. But in JS anything can be thrown (not just exception) so we want to catch it all. But that makes eslint very mad. 

So I resorted to commenting on it everywhere in the code which is ugly, but there doesn't seem to be a better option for now, unless we want to only catch errors, but some libraries will throw stuff up that's not an error because yay javascript
https://stackoverflow.com/questions/66462386/how-to-handle-the-typescript-eslint-no-unsafe-member-access-rule-on-catcherr

# Why ES Lint 8

There is a thing called ES Lint 9. That seems to not really support migrating from an eslintrc.js file. And that is the file type that's created when you want to convert from a tslint project to lint typescript. Even their tooling says dont do this yet.


pipeline-templates/lint.yaml
```diff
@@ -2,17 +2,17 @@ parameters:
...
         vmImage: ${{ parameters.pool.vmImage }}
-    displayName: '🧶 TSLint'
+    displayName: '🧶 ESLint'
     steps:
     - template: install-node.yaml
+    - script: build.cmd
     - bash: |
-        npm install --cache /temp/empty-cache
-        npm install tslint --reg https://registry.npmjs.org/ --verbose
+        npm install eslint@^8.57.0 @typescript-eslint/parser@^8.0.0 @typescript-eslint/eslint-plugin@^8.0.0 --reg https://registry.npmjs.org/ --save-dev --verbose
```
joeloff, line 16: How often would these versions need to change?

nagilson: The `^` will allow them to change automatically, up to the latest major. 8 is the latest major that supports the things we need in typescript, (9 being the true latest) so this should be good for a while. Eventually it will need to be changed to 9, but at this point it will require a lot of work to adjust to breaking changes like this PR did.

joeloff (approved):

PR #1946, opened 2024-09-10
Update `existingDotnetPath` to only take affect if it works.

nagilson:
# Approach

To determine whether the path setting meets an API request:

Try to run `dotnetPathSetting --list-runtimes` on the path. If it is able to return a runtime that matches the same `major.minor` and `type` (`aspnetcore` vs regular runtime vs `windowsdesktop`), then the path is used. If nothing is found, try running `dotnetPathSetting --list-sdks`. If there is an sdk found with the same `major.minor` as the acquire request, use the path. Else, dont use the path.

# Changes 

The `existingDotnetPath` and `sharedExistingDotnetPaths` setting will experience a breaking change.
This setting is used by many users to set the .NET path. That path is returned by our extension when someone makes an API request for a version of .NET.

The problem is, this check does not see whether the path satisfies the version of .NET that the API request was asking for.

Users set the setting for 2 main reasons:
1. They didnt want us to download a .NET Runtime for them.
This often comes from a misunderstanding. The runtime they have is used if it meets the requirements DevKit needs to run under. If it does not, a runtime is downloaded. They think it's the runtime for their project, but it's not. DevKit etc cannot run without this runtime, so if they have installed devkit, they don't really have another option here.

2. The runtime installation did not work for them.
Since the early days of devkit, success rates have gotten much better. We are more confident in the ability to install. In addition, for these people if the setting made things work, we'd expect their setting to still be used because it should in theory have the same runtime.

We add a setting to opt out of this behavior.

# Warning

We also add a warning if the setting is ignored because it does not match.

Due to the above confusion, this may upset some percentage of users.
However, I think this change will make things work for many users who don't understand and could not get things working.

And it is probably better to warn them if their setting is no longer being used. If their setting gets ignored after this change and they were previously in a working state, then it's likely to have been an unstable state where DevKit was running on some sort of runtime... but not really what it was built on top of.

This would appear like so.

image

image

# Messaging

Within the extension settings:

Tries to make the messaging for what this setting does more clear:
image


vscode-dotnet-runtime-library/src/Acquisition/ExistingPathResolver.ts
```diff
@@ -36,21 +67,125 @@ export class ExistingPathResolver
...
-                    return { dotnetPath: existingPaths.sharedExistingPath}
+                    return existingPaths.sharedExistingPath;
                 }
                 else
                 {
                     windowDisplayWorker.showWarningMessage(
                         `Ignoring existing .NET paths defined in settings.json because the setting is only set for other extensions, and not for ${extensionId}`,
                         () => { /* No callback */ },
                     );
+                    return null;
```
nagilson, line 83: I think this should still be undefined

vscode-dotnet-runtime-library/src/Acquisition/ExistingPathResolver.ts
```diff
@@ -36,21 +67,125 @@ export class ExistingPathResolver
...
+    {
+        return this.workerContext.extensionState.get<boolean>('dotnetAcquisitionExtension.allowInvalidPaths') ?? false;
+    }
+
+    private disableWarning() : boolean
+    {
+        return this.workerContext.extensionState.get<boolean>('dotnetAcquisitionExtension.disableExistingPathWarning') ?? false;
+    }
+
+    private async existingPathMatchesAPIRequestCondition(existingPath : string, apiRequest : IDotnetAcquireContext) : Promise<boolean>
```
nagilson, line 101: this name is too long

nagilson:
Adding @baronfel to review text / messaging

baronfel:
Made a couple small tweaks to the messaging, mostly to make the call to action at the end use a heirarchy I think is more likely to result in success for the user:
* VSCode's built-in SDK command
* the .NET download page
* last-resort manual PATH munging

I also tweaked some of the formatting:
* remove the use of backticks in favor of `'` in the non-Markdown description
* added the `markdownDescription` and kept the use of both hyperlinks and backticks in this version

nagilson:
Thank you, I didn't know about the markdownDescription property!

vscode-dotnet-runtime-library/src/Acquisition/ExistingPathResolver.ts
```diff
@@ -36,21 +66,121 @@ export class ExistingPathResolver
...
+        const availableRuntimes = await this.getRuntimes(existingPath);
+        const requestedMajorMinor = versionUtils.getMajorMinor(apiRequest.version, this.workerContext.eventStream, this.workerContext);
+
+        if(availableRuntimes.some((runtime) =>
+        {
+            return runtime.mode === apiRequest.mode && versionUtils.getMajorMinor(runtime.version, this.workerContext.eventStream, this.workerContext) === requestedMajorMinor;
+        }))
+        {
+            return true;
+        }
```
MiYanni, line 107: nit: This formatting seems a little odd to me. I think I'd expect the first set of `{}` to be intended and not directly under the `if` width. That way, it doesn't look like it is the body of the if-statement. Same thing for `if(availableSDKs.some` a few lines below this.

vscode-dotnet-runtime-library/src/Acquisition/ExistingPathResolver.ts
```diff
@@ -36,21 +66,121 @@ export class ExistingPathResolver
...
+                        mode: 'sdk',
+                        version: parts[0],
+                        directory: sdk.split(' ').slice(1).join(' ').slice(1, -1) // need to remove the brackets from the path [path]
+                    } as IDotnetListInfo;
+                }).filter(x => x !== null) as IDotnetListInfo[];
+
+                return runtimeInfos;
+            });
+
+            return sdkInfo;
```
MiYanni, line 152: nit: Extra tab indent.

vscode-dotnet-runtime-library/src/Acquisition/ExistingPathResolver.ts
```diff
@@ -36,21 +66,121 @@ export class ExistingPathResolver
...
+                    {
+                        return null;
+                    }
+                    const parts = sdk.split(' ', 2); // account for spaces in PATH, no space should appear before then and luckily path is last
+                    return {
+                        mode: 'sdk',
+                        version: parts[0],
+                        directory: sdk.split(' ').slice(1).join(' ').slice(1, -1) // need to remove the brackets from the path [path]
+                    } as IDotnetListInfo;
+                }).filter(x => x !== null) as IDotnetListInfo[];
```
MiYanni, line 147: nit: This logic might be simpler as a RegEx, but I don't know the output being parsed. Just an idea.

vscode-dotnet-runtime-library/src/Acquisition/ExistingPathResolver.ts
```diff
@@ -36,21 +66,121 @@ export class ExistingPathResolver
...
+                    return null;
+                }
+                const parts = runtime.split(' ', 3); // account for spaces in PATH, no space should appear before then and luckily path is last
+                return {
+                    mode: parts[0] === aspnetCoreString ? 'aspnetcore' : parts[0] === runtimeString ? 'runtime' : 'sdk', // sdk is a placeholder for windows desktop, will never match since this is for runtime search only
+                    version: parts[1],
+                    directory: runtime.split(' ').slice(2).join(' ').slice(1, -1) // account for spaces in PATH, no space should appear before then and luckily path is last.
+                    // the 2nd slice needs to remove the brackets from the path [path]
+                } as IDotnetListInfo;
+            }).filter(x => x !== null) as IDotnetListInfo[];
```
MiYanni, line 179: nit: Same as above where it might be easier as RegEx. The logic looks similar too, so potential for deduplication.

MiYanni (approved):

vscode-dotnet-runtime-library/src/Acquisition/ExistingPathResolver.ts
nagilson: I can see the logic there, it would probably be shorter. But I am a bit opinionated on RegEx.  I would much rather debug code like this than a RegEx expression. I do appreciate this suggestion, though, and thanks for reviewing :)

vscode-dotnet-runtime-library/src/Acquisition/ExistingPathResolver.ts
nagilson: That's a good point, I will fix this and the other indent issue here: https://github.com/dotnet/vscode-dotnet-runtime/pull/1954

I've refactored the code out into it's own file without any changes so it will be easier to review. Since this file will be deleted, it doesn't make sense to fix it in this PR.

PR #1947, opened 2024-09-11
Improve ARM Mac Path Check

nagilson:
Just as https://github.com/dotnet/vscode-dotnet-runtime/pull/1932 automerged I got some help getting a Mac. 

Here is the conclusion:
VS Code runs on an emulated version of node which will return x64 or use x86 emulation for ARM devices.
os.arch() returns the architecture of the node binary, not the system architecture, so it will not report arm on an arm device.

This is why I've added the additional check below using `uname` which will report the correct architecture that's not being emulated on.

`uname` is a bit costly so we don't need to use it everywhere. Btw, if this command fails because someone removed `uname` from their system, it will just use the existing fallback logic, which is why I've kept it in place. (Believe me, people do some weird system alterations like that...)

Final note:

The .NET SDK can be not under emulation when node is. That is an important caveat, even though all installs this extension makes will be under emulation since node will be, that does not mean existing installs from other sources will be. We can deal with this by checking if the path exists, if we'd expect it to exist when looking for existing installs.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
@@ -208,8 +208,9 @@ This report should be made at https://github.com/dotnet/vscode-dotnet-runtime/is
         }
         else
         {
-            const command = CommandExecutor.makeCommand(`rm`, [`-rf`, `${path.join(path.dirname(this.getMacPath()), 'sdk', install.version)}`, `&&`,
-`rm`, `-rf`, `${path.join(path.dirname(this.getMacPath()), 'sdk-manifests', install.version)}`], true);
+            const macPath = await this.getMacPath();
+            const command = CommandExecutor.makeCommand(`rm`, [`-rf`, `${path.join(path.dirname(macPath), 'sdk', install.version)}`, `&&`,
+`rm`, `-rf`, `${path.join(path.dirname(macPath), 'sdk-manifests', install.version)}`], true);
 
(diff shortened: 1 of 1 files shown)


nagilson:
Actually, while node being on emulation does mean sdks installed by our extension will be using this PATH, this does not mean that other sdks will be installed using the x64 emulation. I don't really like this change. We'd have to condition it nonetheless based on the same check that was already there, so it does not add much value.

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -352,10 +354,19 @@ If you were waiting for the install to succeed, please extend the timeout settin
...
     }
 
-    private getMacPath(macPathShouldExist = true) : string
+    private async getMacPath(macPathShouldExist = true) : Promise<string>
     {
         const standardHostPath = path.resolve(`/usr/local/share/dotnet/dotnet`);
         const arm64EmulationHostPath = path.resolve(`/usr/local/share/dotnet/x64/dotnet`);
+
+        const findTrueArchCommand = CommandExecutor.makeCommand(`uname`, [`-p`]);
+        if((os.arch() === 'x64' || os.arch() === 'ia32') && (await this.commandRunner.execute(findTrueArchCommand, null, false)).stdout.toLowerCase().includes('arm'))
```
nagilson, line 363:
```suggestion
        if((os.arch() === 'x64' || os.arch() === 'ia32') && (await this.commandRunner.execute(findTrueArchCommand, null, false)).stdout.toLowerCase().includes('arm') && (fs.existsSync(arm64EmulationHostPath) || !macPathShouldExist))
```

MiYanni (approved):

PR #1954, opened 2024-09-18
Add new API `dotnet.findPath` to find the .NET host to use if one already exists that meets a requirement.

nagilson:
Helps resolve:
https://github.com/microsoft/vscode-dotnettools/issues/1466
https://github.com/microsoft/vscode-dotnettools/issues/1243
https://github.com/microsoft/vscode-dotnettools/issues/1094

(CDK and C# will need to adopt this API.)

Adds the command `dotnet.findPath` to return the path to a dotnet executable or dotnet file to run that is set on the system. This will only be returned if it meets the requirement. This API is meant to search for machine install or runtimes set via environment variables such as PATH or DOTNET_ROOT.

This API takes in a `IDotnetFindPathContext` which has an Acquire Context. This allows us to edit this API without having to change the acquire API, which uses the `IDotnetAcquireContext`. This `has-a` approach is very common in `.js`.

Callers will add the 'requirement' which allows a .NET version that is 'greater_than_or_equal', 'equal', or 'less_than_or_equal' for `major.minor` versions that they request.

Here is the algorithm / control flow for the PATH lookup that we decided on with partners. Hopefully this helps explain the code behavior and what it should do.

This will also fix the issue where the vscode path setting `dotnet.existingDotnetPath` is used by the extensions which will call into this API as they previously had their own path lookup logic which ignored our setting.

image

* The architecture search is not implemented any longer because it is not possible to do it correctly with the existing dotnet host information. We will need to talk to them about this.

I've tested this with various installs, such as snap installs. These installs have a 'polymorphic executable', which means the realpath of which dotnet is actually the snap executable, so the realpath doesn't work. However, there are installs such as the microsoft packages install which installs a symlink path, and that path doesn't work until the realpath is called. This is why we have various path lookup mechanisms, as various installs use different mechanisms.


vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
```diff
@@ -0,0 +1,150 @@
...
+    {
+        return '';
+        /* The host architecture can be inaccurate. Imagine a local runtime install with no host. There is no way to tell the architecture of that runtime,
+        The Host will not print its architecture in dotnet info.
+        Return '' for now to pass all arch checks.
+
+        Need to get an issue from the runtime team. See https://github.com/dotnet/sdk/issues/33697 and https://github.com/dotnet/runtime/issues/98735*/
+    }
+
+    public async getSDKs(existingPath : string) : Promise<IDotnetListInfo[]>
```
nagilson, line 77: A lot of this code is moved over from another file (for reviewers)

sample/package.json
```diff
@@ -7,7 +7,7 @@
 		"url": "https://github.com/dotnet/vscode-dotnet-runtime.git"
 	},
 	"license": "MIT",
-	"version": "0.0.1",
+	"version": "0.0.9",
```
nagilson, line 10: (This is an internal testing extension only so the version doesn't matter much. We could try to keep it concurrent with the extension version.)

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -432,6 +443,95 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
...
+            throw new EventCancellationError('BadContextualFindPathError', `The find path request was missing required information: a mode, version, architecture, and requestingExtensionId.`);
+        }
+
+        globalEventStream.post(new DotnetFindPathLookupSetting(`Looking up vscode setting.`));
+        const workerContext = getAcquisitionWorkerContext(commandContext.acquireContext.mode, commandContext.acquireContext);
+        const existingPath = await resolveExistingPathIfExists(existingPathConfigWorker, commandContext.acquireContext, workerContext, utilContext, commandContext.versionSpecRequirement);
+
+        if(existingPath)
+        {
+            globalEventStream.post(new DotnetFindPathSettingFound(`Found vscode setting.`));
```
joeloff, line 473: Q: If this is coming from a setting, is this just an abitrary value. Do you need to verify that the path actually exists or is that responsibility deferred to something else?

vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -185,6 +201,45 @@ suite('DotnetCoreAcquisitionExtension End to End', function()
     assert.isFalse(fs.existsSync(result!.dotnetPath), 'the dotnet path result does not exist after uninstalling from all owners');
   }
 
+  function includesPathWithLikelyDotnet(pathToCheck : string) : boolean
+  {
+    const lowerPath = pathToCheck.toLowerCase();
+    return lowerPath.includes('dotnet') || lowerPath.includes('program') || lowerPath.includes('share') || lowerPath.includes('bin') || lowerPath.includes('snap') || lowerPath.includes('homebrew');
```
joeloff, line 207: So this would consider something like `C:\ProgramData` as potentially pointing to a valid location that contains the `dotnet` host?

vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
```diff
@@ -0,0 +1,150 @@
...
+     * The .NET Host will only list versions of the runtime and sdk that match its architecture.
+     * Thus, any runtime or sdk that it prints out will be the same architecture as the host.
+     *
+     * @remarks Will return '' if the architecture cannot be determined for some peculiar reason (e.g. dotnet --info is broken or changed).
+     */
+    // eslint-disable-next-line @typescript-eslint/require-await
+    private async getHostArchitecture(hostPath : string) : Promise<string>
+    {
+        return '';
+        /* The host architecture can be inaccurate. Imagine a local runtime install with no host. There is no way to tell the architecture of that runtime,
```
joeloff, line 70: the host is always included with the runtime, doesn't that make the installation incomplete and therefore invalid?

vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
```diff
@@ -0,0 +1,211 @@
...
+    undefined,
+    false))?.stdout}
+
+Bin Bash Path: ${os.platform() !== 'win32' ? (await this.executor?.execute(CommandExecutor.makeCommand('env', ['bash']), {shell : '/bin/bash'}, false))?.stdout : 'N/A'}
+`
+        ));
+
+        let windowsWhereCommand = '';
+        let unixWhichCommand = '';
+        if(os.platform() === 'win32')
```
joeloff, line 113: only one of these two conditions will be true. Could this be simplified as

```
let command = '';

if (os.platform() === 'win32')
{
    command = ...
}
else
{
    command = ...
}

const finderCommand = command; 
```

joeloff:
You could probably read the machine type information from the host (COFF or ELF depending on the OS). For the runtimes themselves, reading the COFF header of the DLLs should be sufficient.

I wouldn't use the .version file because not everyone writes the same information. Microsoft.NETCore.App doesn't write architecture information, but the SDK does, so it's not reliable

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: Good question, we do verify the path in the function/classes that get the path setting. That was added in the last version.

vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
nagilson: Yes -- this is for the test and is meant to remove any possible dotnets from the path. It is not production code. This is to prevent dotnet installs on the CI machines from interfering with the results of the tests.

vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
nagilson: That's true, this comment (the one in the code) is wrong. However, the host can be there but not return an architecture if it's just a local runtime and a host. I should update this comment.

vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
nagilson: Thank you, this is a good point.

nagilson:
The architecture suggestion is a good one -- we should add this in the next release along with improvements to parsing the PATH with chinese characters due to a nodejs bug.

MiYanni (approved):

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -432,6 +443,95 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
...
+        const dotnetOnRealPATH = await finder.findRealPathEnvironmentSetting();
+        for(const dotnetPath of dotnetsOnPATH ?? [])
+        {
+            const validatedRealPATH = await getPathIfValid(dotnetPath, validator, commandContext);
+            if(validatedRealPATH)
+            {
+                outputChannel.show(true);
+                return validatedRealPATH;
+            }
+        }
```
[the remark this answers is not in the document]
nagilson: Ooo youre right, that's a very subtle bug. :) This is why redundant code is bad but it didn't feel like enough to make its own function. Will fix. Thank you.

vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -227,6 +282,59 @@ suite('DotnetCoreAcquisitionExtension End to End', function()
...
+        // They also have an x64 HOST. We can't set DOTNET_MULTILEVEL_LOOKUP to 0 because it will break the ability to find the host on --info
+        // As our runtime installs have no host. So the architecture will read as x64 even though it's not.
+        //
+        // This is not fixable until the runtime team releases a better way to get the architecture of a particular dotnet installation.
+        await findPathWithRequirementAndInstall('3.1', 'runtime', os.arch() == 'arm64' ? 'x64' : os.arch(), 'greater_than_or_equal', false,
+            {version : '3.1', mode : 'runtime', architecture : 'arm64', requestingExtensionId : requestingExtensionId}
+        );
+    }
+  }).timeout(standardTimeoutTime);
+  */
```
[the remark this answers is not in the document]
nagilson: Yes. This test will work and should be re-added when architecture check is added.

Issue #1958, opened 2024-09-24
Error: Please provide instrumentation key


digama0:
### Describe the bug.

I'm not sure why this extension is installed as I don't use .NET, but it reports an error to the console on startup:

```
Error: Please provide instrumentation key
	at ue (/home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:382196)
	at /home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:305992
	at Object.m [as use] (/home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:297462)
	at /home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:296806
	at s.<computed> [as watch] (/home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:296817)
	at e.<computed> [as initialize] (/home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:305780)
	at e.n [as initialize] (/home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:374092)
	at i (/home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:372019)
	at /home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:372089
	at L (/home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:375051)
	at new e (/home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:371471)
	at t.appInsightsClientFactory (/home/mario/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.6/dist/extension.js:2:393101) {stack: 'Error: Please provide instrumentation key
	at…net-runtime-2.1.6/dist/extension.js:2:393101)', message: 'Please provide instrumentation key'}
```

[...]

PR #1968, opened 2024-09-30
Migrate to connection strings

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1958

The old application insights key was created by @LakshanF, when we migrated to the new vscode-extension-telemetry service, their API had a breaking change to require a connection string instead of an insights key. https://github.com/dotnet/vscode-dotnet-runtime/pull/1948

The connection key can be public and is hard coded. Our existing key has been in our open source, source code for many years. Here is what their guidance says: https://www.npmjs.com/package/@vscode/extension-telemetry

> Follow guide to set up Application Insights in Azure and get your connection string. Don't worry about hardcoding it, it is not sensitive.

Also confirmed this with the devdiv data team who manages this application id within azure. I can't post further information about it here, but they should be a good contact in the future.

Also confirmed when packing the VSIX nothing shows up in the debug console about the key anymore : 
image


vscode-dotnet-runtime-extension/package.json
```diff
@@ -11,7 +11,7 @@
 	"author": "Microsoft Corporation",
 	"displayName": ".NET Install Tool",
 	"description": "This extension installs and manages different versions of the .NET SDK and Runtime.",
-	"appInsightsKey": "02dc18e0-7494-43b2-b2a3-18ada5fcb522",
+	"connectionString": "InstrumentationKey=02dc18e0-7494-43b2-b2a3-18ada5fcb522;IngestionEndpoint=https://westus2-0.in.applicationinsights.azure.com/;LiveEndpoint=https://westus2.livediagnostics.monitor.azure.com/;ApplicationId=e8e56970-a18a-4101-b7d1-1c5dd7c29eeb",
```
baronfel, line 14: One thing that is slightly concerning is that all telemetry calls will be going to the `westus` Azure region despite our users being global. Can you check with them if there is a set of urls we should be using based on user locale instead?

nagilson: That's a good question. The devdiv data team shared that this is intentional as there is only one endpoint. I discussed this further and apparently DotnetCLI and AzureCLI use 1 endpoint in western US as well and they state they havent had any scalability issues so far with this methodology. I did emphasize the regional issues we've solved in the past as being a point of concern.

nagilson:
Closing as this was added in another PR.

Issue #1935, opened 2024-08-22
Do not stealing focus from VS Code editor


stewartadam:
### Describe the bug.

Various extensions depend on the .NET install tool to grab the .NET SDK. As a result, on each initialization of the editor it steals focus from the user to display the Output tab -- for example:

```
ms-azuretools.vscode-bicep: Trying to install .NET 8.0.8~arm64 but it already exists. No downloads or changes were made.
```

See also:
- https://github.com/microsoft/vscode/issues/105270
- https://github.com/microsoft/vscode/issues/203478
- https://github.com/microsoft/vscode/pull/205225

### Please attach log file(s) here if you have any.

```
8/22/2024, 10:10:21 AM DotnetWSLCheckEvent
Checking if system is WSL. OS: darwin

8/22/2024, 10:10:50 AM DotnetAcquisitionRequested
8.0
bcf1754f51e047c9496e95489f0796759eadf89b340c12bde65beaf959b7d7d8

8/22/2024, 10:10:50 AM DotnetRuntimeAcquisitionRequested
8.0
bcf1754f51e047c9496e95489f0796759eadf89b340c12bde65beaf959b7d7d8
runtime

8/22/2024, 10:10:50 AM WebRequestSent
https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json

8/22/2024, 10:10:50 AM DotnetVersionResolutionCompleted
8.0
8.0.8

8/22/2024, 10:10:50 AM DotnetInstallIdCreatedEvent
The requested version 8.0.8 is now marked under the install: {"installId":"8.0.8~arm64","version":"8.0.8","architecture":"arm64","isGlobal":false,"installMode":"runtime"}.

8/22/2024, 10:10:50 AM DotnetLockAttemptingAcquireEvent
Lock Acquisition request to begin.
2024-08-22T17:10:50.472Z
/Users/stewartadam/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.5/dist/installing.lock
/Users/stewartadam/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.5/dist/installing.lock

8/22/2024, 10:10:50 AM DotnetLockAttemptingAcquireEvent
Lock Acquisition request to begin.
2024-08-22T17:10:50.476Z
/Users/stewartadam/.vscode/extensions/ms-dotnettools.vscode-dotnet-runtime-2.1.5/dist/installed.lock
[...]

PR #1976, opened 2024-10-03
Permit Installing .NET via a package.json file

nagilson:
# Intro 

Extensions that rely on .NET often have a slow startup time due to our existing API, which needs to be invoked in typescript code.
This is per a request from the Copilot for Azure extension.

# Change 

For extensions that want the .NET runtime to be installed as soon as possible, they can now also make an API request in their package.json. https://github.com/dotnet/vscode-dotnet-runtime/issues/1935

They should add a [IDotnetAcquireContext](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/IDotnetAcquireContext.ts) object in a section titled 'x-dotnet-acquire' to do so. They don't need to include the `requestingExtensionId`, since this is already in their `package.json`,

This should go at the root of a `package.json`, and not in the `contributes` section.
```
"x-dotnet-acquire": {
    "version": "8.0",
    "mode": "aspnetcore"
}
```

We run the install on startup and also whenever a new extension is installed via the onDidChange API. https://code.visualstudio.com/api/references/vscode-api#Extension&lt;T&gt;

# Testing 

You can easily test this by modifying the SDK extension and adding this, then packging it with `vsce`, installing it, and running the Runtime Extension under the Debug tab. If you do so, you will see that:

If you install a new extension, the request is made. From my testing, it also doesn't grab focus away from the terminal which is great, I figured that would be a huge pain point. Something to still watch out for as it needs to be improved: https://github.com/dotnet/vscode-dotnet-runtime/issues/1935

(Before installing extension)
image
(On install)
image

If you had already installed the extension, the request is also made.
(On Opening VS Code)
image

# Tests

It's a bit hard to test this functionality comprehensively, so while there's a very basic test, I've added a page about it in our vendor testing as well.


Documentation/commands.md
```diff
@@ -13,3 +17,34 @@ This command surfaces an output channel to the user which provides status messag
...
+
+## dotnet.findPath
+
+You can execute this command to return a string to a dotnet runtime path. Pass it a [IDotnetFindPathContext](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/IDotnetFindPathContext.ts) object. The major.minor will be respected based on the requirement you give (greater_than_or_equal will return a dotnet on the PATH that is >= the version requested.)
+
+This returns undefined if no matches are found.
+
+## dotnet.uninstall
+
+You can execute this command to deference / uninstall .NET, either the SDK or runtime, as long as its managed by this extension. Pass it a [IDotnetFindPathContext](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/IDotnetFindPathContext.ts) object suggesting which install you want to remove.
```
nagilson, line 29:
```suggestion
You can execute this command to dereference / uninstall .NET, either the SDK or runtime, as long as it's managed by this extension. Pass it a [IDotnetFindPathContext](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/IDotnetFindPathContext.ts) object suggesting which install you want to remove.
```

vscode-dotnet-runtime-extension/README.md
```diff
@@ -2,7 +2,7 @@
...
 
-This extension provides a unified way for other extensions like the [C#] and [C# Dev Kit] extensions to install local versions of the .NET Runtime, and machine-wide versions of the .NET SDK. Those extensions tell the .NET Install Tool when they would like a .NET SDK to be on the machine, and we install one for them if there's not already one that matches the SDK they need to run properly. In the future, this tool may support allowing users to call the API via VS Code to install .NET and pick a version of the SDK to install themselves.
+This extension provides a unified way for other extensions like the [C#] and [C# Dev Kit] extensions to install local versions of the .NET Runtime, and machine-wide versions of the .NET SDK. Those extensions tell the .NET Install Tool when they would like a .NET SDK to be on the machine, and we install one for them if there's not already one that matches the SDK they need to run properly. Users can also install the .NET SDK to install themselves by reading below.
```
nagilson, line 5:
```suggestion
This extension provides a unified way for other extensions like the [C#] and [C# Dev Kit] extensions to install local versions of the .NET Runtime, and machine-wide versions of the .NET SDK. Those extensions tell the .NET Install Tool when they would like a .NET SDK to be on the machine, and we install one for them if there's not already one that matches the SDK they need to run properly. Users can also install the .NET SDK themselves by reading below.

```

jacdavis:
shipit

bwateratmsft (approved):

nagilson:
Thanks for your review. We still need to get the branding for the versions PR approved and a few other minor PRs for the release, so I expect it will take a few days.

PR #1979, opened 2024-10-10
Exclude Microbuild from scanner

nagilson:
Microbuild is failing credscan following TSA upload. It has several issues and the pipeline is publishing artifacts to the root artifacts staging directory and CredScan is considering everything under that path as part of the pipeline output. MicroBuild is being installed there, so it is incorrectly being brought into the scan.

This repo does not own microbuild, we only do it for signing.

See that cred scan no longer shows up in internal
image


nagilson:
It's going to take some time for this to show up in the cred scanner as resolved because it will need to go into main and stop triggering the existing branches.

joeloff (approved):

Issue #1572, opened 2023-11-21
[NETE2ESDK] [MAC]The failure message should say the installation was cancelled rather than expected dotnet folder/usr/local/share/dotnet/dotnet does not exist if cancelling .NET SDK installation.


v-xiaofchen:
# Describe the bug

The failure message should say the installation was cancelled rather than expected dotnet folder/usr/local/share/dotnet/dotnet does not exist if cancelling .NET SDK installation.

## Other Info:

**Reproduction Steps**
1. On Mac, run the sample command ‘Install .NET SDK Globally via .NET Install Tool (Former Runtime Extension)‘.
2. Use the version it provided, which is 7.0.103. Hit enter.
3. On the .NET Installer, click close before it can continue.
4. Observer the error info

**Expected behavior**
 a failure message saying the installation was cancelled.

**Exceptions & Errors (if any)**
.NET Acquisition Failed: Validation of .dotnet installation for version 7.0.103 failed: Expected dotnet folder/usr/local/share/dotnet/dotnet does not exist.
![Mac1](https://github.com/dotnet/vscode-dotnet-runtime/assets/99376012/1f64db14-8dac-41fa-91f0-7934944527a0)

**Further technical details**
OS: Mac Ventura X64(13.6.1)
VSCode: 1.84.2
Extensions: dotnet-install-tool-2.0.0

Note:
1.There is also a warning pop up in VS Code about telemetry collection in the bottom right corner.
This pop up will either say
 
The .NET tools collect usage data in order to help us improve your experience. It is collected by Microsoft and shared with the community. You can opt-out of telemetry by setting the DOTNET_CLI_TELEMETRY_OPTOUT environment variable to '1' or 'true' using your favorite shell.
Read more about .NET CLI Tools telemetry: [https://aka.ms/dotnet-cli-telemetry`](https://aka.ms/dotnet-cli-telemetry%60),

2.This issue is only repro on Mac.

nagilson:
It seems I was mistaken since there even isnt really a custom message for this on windows, it currently just exits with 1 on the installer.

nagilson:
Mac does not have an exit code for cancels and you can  only cancel by hitting X. There does not seem to be a good way for us to catch this at the moment.

v-xiaofchen:
on Extensions: dotnet-install-tool-2.0.9, It actually fail like bellow:
[...]

PR #1993, opened 2024-10-16
Consider when the user Cancels the SDK Installer on Mac

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1572 

# Background 

When installing via Mac, our biggest failure right now is 'Validation of Installation' failed. This is because we use `open` which returns 0 even if someone closes the installer, cancels the installation, or fails the password prompt in the .NET installer, so we assume it has succeeded and then check to make sure .NET was installed and fail because it was not
 
We could
A - Assume that someone cancelled if the install isnt there and open on the installer returned 0
B - Move to `sudo installer -pkg Path -target LocalSystem`
 
Unfortunately to use installer you need to use sudo first so we need to use this joyous password prompt every single time if we do that:

image

I consulted our PM to decide how to handle this and he suggested:

 > we could pop up a question notification? "hey, it looks like SDK installation failed - would you like to retry the installation? Y/N/Shut up and go away forever"

I started the implementation of this but it is very ugly and the telemetry logic/counting of where we are in the install gets jank and it would take a lot of effort. In addition, we would want a 'yes' and 'no' button with distinct categories to solve what we are trying to solve -- yes, I want to retry and I had cancelled it; yes, I want to retry because it failed; no with either or the above.

So what I decided to do:

# What's in the PR

When the Mac installation for SDKs fails but the installer returned 0, we add a message asking the user if they had cancelled the installation request, and if so and they want the SDK to be installed, to please try again.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/IInstallationValidator.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/IInstallationValidator.ts
@@ -3,11 +3,13 @@
 *  The .NET Foundation licenses this file to you under the MIT license.
 *--------------------------------------------------------------------------------------------*/
 
+import { IDotnetAcquireContext, IVSCodeExtensionContext } from '..';
 import { IEventStream } from '../EventStream/EventStream';
 import { DotnetInstall } from './DotnetInstall';
 
 export abstract class IInstallationValidator {
     constructor(protected readonly eventStream: IEventStream) {}
 
-    public abstract validateDotnetInstall(install: DotnetInstall, dotnetPath: string, isDotnetFolder? : boolean, failOnErr? : boolean): void;
+    public abstract validateDotnetInstall(install: DotnetInstall, dotnetPath: string, isDotnetFolder? : boolean, failOnErr? : boolean,
+        sdkInstallationToRetry? : IDotnetAcquireContext, vscodeContext? : IVSCodeExtensionContext): void;
 }
--- a/vscode-dotnet-runtime-library/src/Acquisition/IInstallationValidator.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/IInstallationValidator.ts
@@ -10,6 +10,5 @@ import { DotnetInstall } from './DotnetInstall';
 export abstract class IInstallationValidator {
     constructor(protected readonly eventStream: IEventStream) {}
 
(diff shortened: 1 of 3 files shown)


vscode-dotnet-runtime-library/src/Acquisition/IInstallationValidator.ts
```diff
@@ -3,6 +3,7 @@
 *  The .NET Foundation licenses this file to you under the MIT license.
 *--------------------------------------------------------------------------------------------*/
 
+import { IDotnetAcquireContext, IVSCodeExtensionContext } from '..';
```
smitpatel, line 6: Is this needed?

vscode-dotnet-runtime-library/src/Acquisition/InstallationValidator.ts
```diff
@@ -34,8 +35,16 @@ export class InstallationValidator extends IInstallationValidator {
...
-            `${dotnetValidationFailed} The dotnet folder is empty "${dotnetPath}"`, install, dotnetPath);
+            try
+            {
+                this.assertOrThrowError(failOnErr, fs.readdirSync(folder).length !== 0,
+                `${dotnetValidationFailed} The dotnet folder is empty "${dotnetPath}"`, install, dotnetPath);
+            }
+            catch(error : any) // fs.readdirsync throws ENOENT so we need to recall the function
+            {
+                this.assertOrThrowError(failOnErr, false,
+                `${dotnetValidationFailed} The dotnet file dne "${dotnetPath}"`, install, dotnetPath);
```
smitpatel, line 46: dne?

smitpatel (approved):

vscode-dotnet-runtime-library/src/Acquisition/InstallationValidator.ts
nagilson: does not exist

nagilson: I should be more clear with these messages, good point

vscode-dotnet-runtime-library/src/Acquisition/IInstallationValidator.ts
nagilson: This is no longer needed, thank you, I will remove this in a upcoming change

PR #1999, opened 2024-10-23
Always Return `IDotnetAcquireResult` or `undefined` instead of sometimes

nagilson:
Per https://github.com/dotnet/vscode-csharp/pull/7684#discussion_r1811675432
The current findPath API can return a `IDotnetAcquireResult` (which is just an object with `dotnetPath : string`) if the existingDotnetPath is returned, but otherwise it will return a string. This is a mistake. I would like to break it as quickly as possible before it has more widespread adoption.

 The class gives us more flexibility to add more information in the future besides the path string without breaking people.

I also added return type annotations to the functions so this does not happen again.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/extension.ts
+++ b/vscode-dotnet-runtime-extension/src/extension.ts
@@ -450,7 +450,8 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
      * The architecture should be of the node format ('x64', 'x86', 'arm64', etc.)
      *
      * @returns the path to the dotnet executable as an IDotnetAcquireResult (or undefined), if one can be found. This should be the true path to the executable. undefined if none can be found.
-     * // Before version 2.2.2, the result could be a string, undefined, or an IDotnetAcquireResult. This was changed to be more consistent with the rest of the APIs.
+     * Before version 2.2.2, the result could be a string, undefined, or an IDotnetAcquireResult. This was changed to be more consistent with the rest of the APIs.
+
      *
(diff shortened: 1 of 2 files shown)


vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -448,14 +449,15 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
      * We wrap an AcquisitionContext which must include the version, requestingExtensionId, architecture of .NET desired, and mode.
      * The architecture should be of the node format ('x64', 'x86', 'arm64', etc.)
      *
-     * @returns the path to the dotnet executable, if one can be found. This should be the true path to the executable. undefined if none can be found.
+     * @returns the path to the dotnet executable as an IDotnetAcquireResult (or undefined), if one can be found. This should be the true path to the executable. undefined if none can be found.
+     * // Before version 2.2.2, the result could be a string, undefined, or an IDotnetAcquireResult. This was changed to be more consistent with the rest of the APIs.
```
joeloff, line 453: Is the double `//` necessary? Aren't you already inside a `/* */` comment block?

nagilson, line 453:
```suggestion
     * Before version 2.2.2, the result could be a string, undefined, or an IDotnetAcquireResult. This was changed to be more consistent with the rest of the APIs.

```

nagilson: No, I blame co pilot

joeloff (approved):

PR #2000, opened 2024-10-23
Fix .NET Version Parsing for FindPath API

nagilson:
The old comparison used string comparison which is fine for versions of dotnet until dotnet 10. '10.0' < '9.0'. :joy:

This fixes that. I suppose it's a bit clunky and we could dedupe / share the compare logic but it's not being used elsewhere in this same form, so I don't plan to do that right now. There are also places that depend on the output versions being strings, so that's a problem to change that now too. It probably should have been a class from the beginning but unfortunately that just isn't how it was initially done, which is understandable because the scope at that time was much more limited.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
+++ b/vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
@@ -243,7 +243,7 @@ suite('DotnetCoreAcquisitionExtension End to End', function()
     if(shouldFind)
     {
         assert.exists(result.dotnetPath, 'find path command returned a result');
-        assert.equal(result.dotnetPath, installPath, 'The path returned by findPath is correct');
+        assert.equal(result.dotnetPath.toLowerCase(), installPath.toLowerCase(), 'The path returned by findPath is correct');
     }
     else
     {
@@ -319,7 +319,7 @@ suite('DotnetCoreAcquisitionExtension End to End', function()
 
   test('Find dotnet PATH Command Met Version Condition', async () => {
     // Install 8.0, look for 3.1 with accepting dotnet less than or eq to 8
-    await findPathWithRequirementAndInstall('8.0', 'runtime', os.arch(), 'less_than_or_equal', true,
+    await findPathWithRequirementAndInstall('8.0', 'runtime', os.arch(), 'greater_than_or_equal', true,
         {version : '3.1', mode : 'runtime', architecture : os.arch(), requestingExtensionId : requestingExtensionId}
     );
   }).timeout(standardTimeoutTime);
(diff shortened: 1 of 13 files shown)


vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -316,13 +316,27 @@ suite('DotnetCoreAcquisitionExtension End to End', function()
     }
   }).timeout(standardTimeoutTime);
 
-  test('Find dotnet PATH Command Unmet Version Condition', async () => {
-    // Install 3.1, look for 8.0 which is not less than or equal to 3.1
-    await findPathWithRequirementAndInstall('8.0', 'runtime', os.arch(), 'less_than_or_equal', false,
+  test('Find dotnet PATH Command Met Version Condition', async () => {
+    // Install 8.0, look for 3.1 with accepting dotnet less than or eq to 8
```
nagilson, line 320:
```suggestion
    // Install 8.0, look for 3.1 with accepting dotnet gr than or eq to 3.1

```

vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
```diff
@@ -136,22 +136,38 @@ Please set the PATH to a dotnet host that matches the architecture ${requirement
         return os.platform() === 'win32' ? (await this.executor!.tryFindWorkingCommand([CommandExecutor.makeCommand('chcp', ['65001'])])) !== null : false;
     }
 
-    private stringVersionMeetsRequirement(foundVersion : string, requiredVersion : string, requirement : DotnetVersionSpecRequirement) : boolean
+    private stringVersionMeetsRequirement(availableVersion : string, requestedVersion : string, requirement : DotnetVersionSpecRequirement) : boolean
     {
```
smitpatel, line 140: can you use semver for comparison?

nagilson: That's a good suggestion, `semver` is a good option. I hadn't considered that, thank you. I wish we were using semver in the code a lot earlier in many places. 

Going forward I will use that. I do want to get this merged quickly though, so I'm not sure about making that change now considering this is already written and tested 🤔

smitpatel (approved):

vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
smitpatel: Can you file an issue to use semver rather than parsing versions ourselves?

nagilson: Yes, good idea :eyes: https://github.com/dotnet/vscode-dotnet-runtime/issues/2003

Issue #2003, opened 2024-10-25
Use Semver Library


nagilson:
Instead of manually coding all of our version parsing, we should probably migrate to the semver library.
https://github.com/dotnet/vscode-dotnet-runtime/pull/2000#discussion_r1817205415 Please see this for more information.

agarwalishita:
Can you please assign this issue on my name?
Since it's a first contribution, I may connect with you for more information.

nagilson:
Absolutely, I would be happy to :) This will be replacing some of the logic in the VersionUtilities and the code that references them, as well as the logic mentioned in the PR above.

nagilson:
I need to make an urgent change regarding this, so for this task I am going to pick it up now.

nagilson:
This isn't really a simple fix. We need to support '9.0' and '9' as versions but those are not valid semver versions

PR #2008, opened 2024-11-04
Consider Runtime Patch Versions in PATH API

nagilson:
The C# extension will now fail if the runtime is not 8.0.10 or higher on mac so we need to support patch version lookup.

~~Sadly we cannot migrate to semver as described in https://github.com/dotnet/vscode-dotnet-runtime/issues/2003 because semver rejects strings such as 8.0 which are allowable in many parts of our code and APIs.~~ (Thank you to @baronfel for seeing this comment and telling me about semver.coerce. Except, I dont really want to rewrite the code again now after writing it once already, gah)

This also adds code for SDK lookup but no test since that's harder to test, like remarked in the code.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
+++ b/vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
@@ -379,6 +379,14 @@ suite('DotnetCoreAcquisitionExtension End to End', function()
     }
   }).timeout(standardTimeoutTime);
 
+  test('Find dotnet PATH Command Unmet Runtime Patch Condition', async () => {
+    // Install 8.0.{LATEST, which will be < 99}, look for 8.0.99 with accepting dotnet gr than or eq to 8.0.99
+    // No tests for SDK since that's harder to replicate with a global install and different machine states
+    await findPathWithRequirementAndInstall('8.0', 'runtime', os.arch(), 'greater_than_or_equal', true,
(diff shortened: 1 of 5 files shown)


MiYanni (approved):

vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -379,6 +379,14 @@ suite('DotnetCoreAcquisitionExtension End to End', function()
     }
   }).timeout(standardTimeoutTime);
 
+  test('Find dotnet PATH Command Unmet Runtime Patch Condition', async () => {
+    // Install 8.0.{LATEST, which will be < 99}, look for 8.0.99 with accepting dotnet gr than or eq to 8.0.99
+    // No tests for SDK since that's harder to replicate with a global install and different machine states
+    await findPathWithRequirementAndInstall('8.0', 'runtime', os.arch(), 'greater_than_or_equal', true,
```
nagilson, line 385: This may actually need to be false

MiYanni: I did look at this line specifically during my review because the comment text confused me a bit. The boolean goes to `shouldFind`, right?

nagilson: Ya, thanks for looking at that. That's right. There are so many parameters going into this test helper function, the way I wrote it makes it kind of mind bending to tell what its doing 😵

nagilson:
Thanks Yanni for your review 😄

PR #2009, opened 2024-11-05
Converted architecture and requirement input to lowercase 

agarwalishita:
Description

If the value of the .NET runtime architecture or the condition to search for a requirement contains uppercase letters, it results in discovery of undefined .NET path 

Fix

Both inputs are converted to lowercase before searching the .net path
Now, extension working as expected


nagilson (approved):
This looks good to me, thank you 😄

PR #2010, opened 2024-11-05
Improve SDK lookup in PATH API

nagilson:
# Summary 

The vscode setting we have is meant for runtime lookup but it currently gets used under sdk mode. When looking for an sdk, we also dont need to scan --list-runtimes since it has no sdk.

We also add code here to check the path string manually if which or where does not exist, which happens more often than I'd prefer with customers.

The rest of the logic is relatively the same. I looked again at the code across our codebases. 

# Misc 

2 things of note compared to their code:
workingWithTestSignedSdk is not respected. This is a C# devkit property used when finding an sdk on mac to support adhoc signed sdks. Their logic can handle this separately and is already isolated so we don't need to couple this logic.

This doesnt use workspace folders first folder for cwd, just uses __dirname via the command executor class, unlike C# DevKIt. but it shouldnt matter because it uses the full path, not just 'dotnet' which is what they use to call into the host.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
@@ -147,7 +147,7 @@ Bin Bash Path: ${os.platform() !== 'win32' ? (await this.executor?.execute(Comma
                     const resolvedDotnetPath = path.resolve(pathOnPATH, dotnetExecutable);
                     if (existsSync(resolvedDotnetPath))
                     {
-                        this.workerContext.eventStream.post(new DotnetFindPathLookupPATH(`Looking up .NET on the path by processing PATH string. resolved: ${resolvedDotnetPath}.`);
+                        this.workerContext.eventStream.post(new DotnetFindPathLookupPATH(`Looking up .NET on the path by processing PATH string. resolved: ${resolvedDotnetPath}.`));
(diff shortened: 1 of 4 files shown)


vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
```diff
@@ -129,12 +129,36 @@ Bin Bash Path: ${os.platform() !== 'win32' ? (await this.executor?.execute(Comma
 
         const findCommand = CommandExecutor.makeCommand(pathLocatorCommand, ['dotnet']);
         const dotnetsOnPATH = (await this.executor?.execute(findCommand, options))?.stdout.split('\n').map(x => x.trim()).filter(x => x !== '');
-        if(dotnetsOnPATH)
+        if(dotnetsOnPATH && dotnetsOnPATH.length > 0)
```
MiYanni, line 132: nit: Just a random Q. Why is `PATH` in all caps here and in other constants below?

MiYanni (approved):

vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
nagilson: It's the PATH environment variable which is generally in caps :) Its easier to tell I mean the env var if I use PATH instead of Path which could be other paths. At least imo. Thanks for your review.

PR #2011, opened 2024-11-07
Add API Option to Reject Preview Runtimes & SDKs

nagilson:
C# DevKit etc do not want to run on preview runtimes or SDKs in certain scenarios.

This adds to the `IDotnetFindPathContext` an optional flag `rejectPreviews` (because previous API calls need to still work, and also this can be turned off by default) that, when the `findPath` command is used, dotnet installations with previews are not accepted. Preview basically meaning... rc, preview, alpha, rtm, etc.

Also, this adds the lodash library (confirmed MIT license and 41 mil downloads per week) to make deep clones of test objects in the tests. This is the most convenient way to make a copy and not grab the internal references of an object besides JSON serializing and deserializing the object, but that is more prone to error and slowness.

### DIFF
--- a/vscode-dotnet-runtime-library/package.json
+++ b/vscode-dotnet-runtime-library/package.json
@@ -25,6 +25,7 @@
 	},
 	"devDependencies": {
 		"@types/chai": "4.2.22",
+		"@types/lodash": "^4.17.13",
 		"@types/proper-lockfile": "^4.1.2",
 		"glob": "^7.2.0"
 	},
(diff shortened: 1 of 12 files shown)


vscode-dotnet-runtime-library/package.json
```diff
@@ -25,6 +25,7 @@
 	},
 	"devDependencies": {
 		"@types/chai": "4.2.22",
+		"@types/lodash": "^4.17.13",
```
MiYanni, line 28: I'm absolutely not against using new packages and things, but does C# DevKit have any restrictions for new packages? Just wondering from a security standpoint.

MiYanni (approved):

vscode-dotnet-runtime-library/package.json
nagilson: Thanks for taking a look :) There shouldn't be any restrictions besides license / legal / security, which are handled internally by our tooling.

PR #2044, opened 2024-11-14
OS emojis idea

MiYanni:
## Summary
I saw that your build use emojis to identify things. I was talking to Jacques and mentioned it, and he made a joke that you could use the actual emojis that associate with the OS's.  So, this just changes the different OS jobs to use the 🪟 for Windows, 🐧 for Linux, and 🍎 for macOS.

### DIFF
--- a/1pr-azure-pipeline.yml
+++ b/1pr-azure-pipeline.yml
@@ -30,15 +30,15 @@ parameters:
   - name: NetCore-Public
     vmImage: windows-latest
     os: windows
-    emoji: ⚪
+    emoji: 🪟
   - name: NetCore-Public
     vmImage: ubuntu-latest
     os: linux
-    emoji: 🟣
+    emoji: 🐧
   - name: NetCore-Public
     vmImage: macOS-latest
     os: macOS
-    emoji: ⚫
+    emoji: 🍎
 
 stages:
 - stage: o # o is just used so it looks like a bullet point in the output of devops


nagilson:
I certainly prefer the look of these :) But part of the reason I chose the ones I did was that I wanted them to be in alphabetical order, such that it would chose Windows first and execute that leg in the pipeline first. Because getting an agent takes a long time and Windows is the leg which then gets used to produce a build. 

But actually it seems it still really loves to run the mac os one first. Do you happen to have any idea as to how to prioritize these?

nagilson:
Also, I guess this just changes the 1pr pipeline, but you might wanna change the 1es pipeline. or maybe that's why you did it as described above ( keep 1es ordering priority queue), though it seems to not be how it behaves?

MiYanni:
@nagilson Jobs always run in parallel. So, there is never an order for execution. If you're looking in Azure DevOps at the pipeline running, the order in which they appear is the order they are declared in the YAML file. The order for the GitHub Checks... I believe is alphabetical, but don't quote me on that. So, I think it really depends on exactly what you want to accomplish.

nagilson:
Thanks @MiYanni :) 

Specifically, this is the order of which one gets executed first. The cloud build agents are slow and sometimes we wait 10-30+ minutes to get an agent per job. I want the windows job to request an agent first and start first so that it is the first one to get an agent. Sometimes the mac one gets an agent (even now with alphabetical priority on Win), then the windows one is queued after and then it takes an extra 30 minutes to get, then the next job that waits on the windows job has to wait again for another agent and it becomes very slow.

nagilson:
Im gonna go in soon so maybe we can talk about this in the non digital realm

nagilson:
Also anyway like, the alphabetical thing doesnt seem to always work or work at all so we can take this change 😄

MiYanni:
@nagilson There's no priority on the jobs at all. They're entirely parallel. The windows and linux jobs use the same agent pool, but macOS doesn't. The reason macOS is almost always going to acquire an agent first is because is isn't a self-hosted pool. Meaning, macOS is using `Azure Pipelines`, which are the general agents given to any Azure Pipelines user. We only use this because our org (dnceng/dnceng-public) doesn't have a macOS agent pool. For windows and linux, we use `NetCore1ESPool-Internal` and `NetCore-Public` respectively. You can look at those pools [here](https://dnceng.visualstudio.com/internal/_settings/agentqueues?queueId=394&view=jobs) and [here](https://dev.azure.com/dnceng-public/public/_settings/agentqueues?queueId=13&view=jobs). But I don't know the actual agent acquisition mechanics beyond what the logs show. I don't know if there are a certain number of windows agents versus linux agents, or if there is just a generic pool of agents that get imaged every acquisition to that OS.

nagilson (approved):

PR #2048, opened 2024-11-15
Add Hostfxr Discovery Logic in place of Bad PATH from VS Code

nagilson:
# What This Does 🤘 

This improves the detection of .NET installs to rely on the installation records containing paths to the dotnet host instead of just the PATH as a backup option for locating dotnet.

What I will describe below is documented here:
See https://github.com/dotnet/runtime/issues/109974
See https://github.com/dotnet/designs/blob/main/accepted/2021/install-location-per-architecture.md

Thank you to @elinor-fung for help understanding these docs.
We've had confirmation from internal folks who were blocked that this fixed the issues for them and I did lots of manual testing. 

## Windows 🪟 

For Windows, the registry includes records written by various installers. Anyone who makes an installer for .NET is supposed to follow the logic which is documented above.

Within `HKEY_LOCAL_MACHINE\\SOFTWARE\\dotnet\\Setup\\InstalledVersions\\${architecture}\\`:
The `sharedhost` registry key stores the path of dotnet that gets put on the PATH. As we want to mimic the terminal behavior as much as possible, we first check this location. For non native installs: sharedhost nodes may exist, but there will be no path key, so we can't use it in this case.

The `InstallLocation` key stores other paths to the host that might not be on the PATH, such as an x64 host on an arm64 machine. The InstallLocation is always under the 32 bit reg node for our purposes, and the sharedhost is always on the native node. You can't install arm64 SDK or Runtime on x64, but you can do vice versa and they follow the same methodology and they follow the same rules.

## Unix 🐧 

For Linux and Mac, installers are supposed to write to `/etc/dotnet/install_location_${requestedArchitecture}` a file which contains only the path. I confirmed various mac installers do that. The ubuntu installer via feeds does this at least in `--classic` mode. The snap installer does as well but does so incorrectly, as it writes /usr/lib/dotnet/ to it, which is not actually where dotnet gets installed via snap. We might wanna alert them to that. Our existing code seems to do a good job with snap, vscode has not given us incorrect information here to my knowledge, so I am not too concerned.

Prior to .NET 6, installers wrote to `/etc/dotnet/install_location`. We use this as a backup and so does .NET 6+.

Thank you to @richlander for coming up with this idea with me.

# Context ❓ 

For years now, the `C#` extension, `Unity` Extension, `MAUI`, etc, and now, `C# DevKit`, have been plagued by issues trying to find the .NET Installation on the PATH environment variable.

Our team did a thorough analysis as to the methods in which .NET is installed across many operating systems and architectures, how to check the PATH, spawning a node.js process with different shells, resolving `realpaths`, `symlinks`, and polymorphic executables (`snap`) etc., and released an API for PATH lookup.

However, we still are facing issues with the PATH. Upon final conclusion, it seems no matter what we call, the PATH information we get as an extension calling into node's `child_process` library, or `process.env`, or any other method we could think of, the environment information is not complete. There is also the `vscode.environmentVariableCollection`, which we do leverage, but my understanding is that this is only for editing the terminal / getting workspace preferences. Using the process explorer, it seems that there are some subprocesses in VS Code that have access to more environment variables than others (presumably electron main vs subprocesses), but I believe the whatever runs the extension host is not one of them.

I asked the VS Code team if we all been taking the incorrect approach here to try `process.env`, `vscode.environmentVariableCollection`, `child_process`, etc. They confirmed our approach is correct.

# When does this cause problems? 🔍 

Here is one such example [[BUG] Cannot find .NET SDK installation from PATH environment (macOS) · Issue #1471 · microsoft/vscode-dotnettools (github.com)](https://github.com/microsoft/vscode-dotnettools/issues/1471). 
The most recent example is from an internal member. There are like 20 to 60+ more. I'm not going to link them all. Many of those are because the PATH API has not been adopted in general, not just a problem with the PATH API, because it is not being called by either extension in RTM prod yet.

In a terminal with bin/bash, his PATH looks like:

```
PATH=/opt/anaconda3/bin:/opt/anaconda3/condabin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/usr/local/share/dotnet:~/.dotnet/tools

```

When we call from our extension, using node: `child_process.spawnSync('env', [], { env: process.env, shell: '/bin/bash')`

```
PATH=/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin
```

Note that the environment pretends that it doesn't have dotnet or anaconda installed, amongst others. There is no alternative correct source of truth to get the actual env, from within vscode. This statement is not coming from just me, it is also coming from the VS Code team.

# How VS Code Gets Env 💥  🐛 

In VS Code, there is a file called [shellEnv.ts](https://github.com/microsoft/vscode/blob/28030eb6be98ab8d92e3883d83bfafeeece17cd3/src/vs/platform/shell/node/shellEnv.ts). I discused this file with them as well. After further discussion with the VS Code team, the VS Code bootstrapping code changes whether vscode is launched in dock mode or from the terminal. In the dock, this`shellEnv.ts` logic kicks in to try to correct and find the environment. It appears there is a problem with that code's ability to determine the environment. It is challenging to get a repro environment, and their logs are not sufficient to diagnose the problem without building from source on a customer's machine, which I am not going to pester someone to do. 

I may drive an effort for a fix in the vscode team, but this approach seemed like the fastest way to get to a resolution.

# Interested Individuals 📲 
cc @arunchndr @dibarbet @adrianvmsft @lifenglu @webreidi @baronfel 
This is not a blocker to the PATH API but a further improvement for when VS Code gives incorrect information. No action is required from those who are adapting the API.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/extension.ts
+++ b/vscode-dotnet-runtime-extension/src/extension.ts
@@ -75,6 +75,7 @@ import {
     DotnetFindPathLookupSetting,
     DotnetFindPathMetCondition,
     DotnetFindPathCommandInvoked,
+    DotnetFindPathNoPathMetCondition,
     JsonInstaller,
 } from 'vscode-dotnet-runtime-library';
 import { dotnetCoreAcquisitionExtensionId } from './DotnetCoreAcquisitionId';
@@ -512,7 +513,27 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
             return { dotnetPath: validatedRoot };
         }
 
+        const dotnetOnHostfxrRecord = await finder.findHostInstallPaths(commandContext.acquireContext.architecture);
+        for(const dotnetPath of dotnetOnHostfxrRecord ?? [])
+        {
+            const validatedHostfxr = await getPathIfValid(dotnetPath, validator, commandContext);
+            if(validatedHostfxr)
+            {
+                loggingObserver.dispose();
+                return { dotnetPath: validatedHostfxr };
+            }
+        }
+
         loggingObserver.dispose();
+        globalEventStream.post(new DotnetFindPathNoPathMetCondition(`Could not find a single host path that met the conditions.
+existingPath : ${existingPath?.dotnetPath}
+onPath : ${JSON.stringify(dotnetsOnPATH)}
+onRealPath : ${JSON.stringify(dotnetsOnRealPATH)}
+onRoot : ${dotnetOnROOT}
+onHostfxrRecord : ${JSON.stringify(dotnetOnHostfxrRecord)}
+
+Requirement:
+${JSON.stringify(commandContext)}`));
         return undefined;
     });
 
--- a/vscode-dotnet-runtime-extension/src/extension.ts
+++ b/vscode-dotnet-runtime-extension/src/extension.ts
@@ -517,7 +517,7 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
         for(const dotnetPath of dotnetOnHostfxrRecord ?? [])
         {
             const validatedHostfxr = await getPathIfValid(dotnetPath, validator, commandContext);
-            if(validatedHostfxr)
+            if(validatedHostfxr && process.env.DOTNET_INSTALL_TOOL_SKIP_HOSTFXR !== 'true')
             {
                 loggingObserver.dispose();
                 return { dotnetPath: validatedHostfxr };
--- a/vscode-dotnet-runtime-library/src/Acquisition/RegistryReader.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/RegistryReader.ts
@@ -32,6 +32,7 @@ export class RegistryReader extends IRegistryReader
     public async getHostLocation(architecture : string) : Promise<string | undefined>
     {
         // InstallLocation vs sharedhost:
+
         // The sharedhost registry key stores the path of dotnet that gets put on the PATH.
         // The InstallLocation stores other paths to the host that might not be on the PATH, such as an x64 host on an arm64 machine.
         // The InstallLocation is always under the 32 bit reg node for our purposes, and the sharedhost is always on the native node.
--- a/vscode-dotnet-runtime-library/src/Acquisition/RegistryReader.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/RegistryReader.ts
@@ -28,6 +28,7 @@ export class RegistryReader extends IRegistryReader
 
     /**
      * @remarks architecture values accepted: x64, arm64. x86, arm32 possible on linux but 32 bit vscode is not supported.
+     * @returns the folder containing the host
      */
     public async getHostLocation(architecture : string) : Promise<string | undefined>
     {
@@ -41,6 +42,7 @@ export class RegistryReader extends IRegistryReader
         // For non native installs: sharedhost nodes may exist, but there will be no path key
 
         // See https://github.com/dotnet/runtime/issues/109974
+        // Docs: https://github.com/dotnet/designs/blob/main/accepted/2021/install-location-per-architecture.md
         const queryForPATHLocationNotViaPATH = `HKEY_LOCAL_MACHINE\\SOFTWARE\\dotnet\\Setup\\InstalledVersions\\${architecture}\\sharedhost`;
         const queryForInstallLocationNotPATH = `HKEY_LOCAL_MACHINE\\SOFTWARE\\dotnet\\Setup\\InstalledVersions\\${architecture}\\InstallLocation`;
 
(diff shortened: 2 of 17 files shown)


vscode-dotnet-runtime-library/src/Acquisition/RegistryReader.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     *
+     * @returns an array containing fully specified / specific versions of all globally installed sdks on the machine in windows for 32 and 64 bit sdks.
+     */
+    public async getGlobalSdkVersionsInstalledOnMachine() : Promise<Array<string>>
+    {
+        let sdks: string[] = [];
+
+
```
MiYanni, line 88: nit: blank line

vscode-dotnet-runtime-library/src/Acquisition/RegistryReader.ts
```diff
@@ -0,0 +1,172 @@
...
+                    const registryLookup = await this.queryRegistry(query, true);
+                    if(registryLookup === null)
+                    {
+                        return [];
+                    }
+
+                    let installRecordKeysOfXBit = '';
+                    if(registryLookup.status === '0')
+                    {
+
```
MiYanni, line 108: nit: blank line

vscode-dotnet-runtime-library/src/Acquisition/RegistryReader.ts
```diff
@@ -0,0 +1,172 @@
...
+        if (os.platform() === 'win32')
+        {
+            const sdkInstallRecords64Bit = 'HKEY_LOCAL_MACHINE\\SOFTWARE\\dotnet\\Setup\\InstalledVersions\\x64\\sdk';
+            const sdkInstallRecords32Bit = sdkInstallRecords64Bit.replace('x64', 'x86');
+            const sdkInstallRecordsArm64 = sdkInstallRecords64Bit.replace('x64', 'arm64');
+
+            const queries = [sdkInstallRecords32Bit, sdkInstallRecords64Bit, sdkInstallRecordsArm64];
+            for ( const query of queries )
+            {
+                    // /reg:32 is added because all keys on 64 bit machines are all put into the WOW node. They won't be on the WOW node on a 32 bit machine.
```
MiYanni, line 98: nit: This section has extra spaces in the indentation.

vscode-dotnet-runtime-library/src/Acquisition/RegistryReader.ts
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     *
+     * @param registryQueryResult the raw output of a registry query converted into a string
+     * @returns
+     */
+    private extractVersionsOutOfRegistryKeyStrings(registryQueryResult : string) : string[]
+    {
+        if(registryQueryResult === '')
+        {
+                return [];
```
MiYanni, line 157: nit: indentation

MiYanni (approved):
Changes look good to me. Thanks for the detailed PR description. The only thought I had was about the design, which doesn't need to be changed for this PR. But my idea was that instead of just having a utility-style DotnetPathFinder, you'd instead make platform-specific implementation classes that know how to resolve the path for those platforms. Basically, turning the imperative `os.platform() === 'win32'` logic into a `WindowsPathResolver` or some such, and make ones for the other OS's. I believe this design is used for `LinuxVersionResolver` and thought it might fit for PATH too.

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -512,7 +513,27 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
...
         loggingObserver.dispose();
+        globalEventStream.post(new DotnetFindPathNoPathMetCondition(`Could not find a single host path that met the conditions.
+existingPath : ${existingPath?.dotnetPath}
+onPath : ${JSON.stringify(dotnetsOnPATH)}
+onRealPath : ${JSON.stringify(dotnetsOnRealPATH)}
+onRoot : ${dotnetOnROOT}
+onHostfxrRecord : ${JSON.stringify(dotnetOnHostfxrRecord)}
+
+Requirement:
+${JSON.stringify(commandContext)}`));
```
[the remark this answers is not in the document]
nagilson: This is for the log for the most part. It does go into telemetry but we shouldnt collect PII so most/all of this is gonna get redacted in telemetry.

vscode-dotnet-runtime-library/src/Acquisition/RegistryReader.ts
nagilson: These nits are good to have, thanks for taking a look :) 
I'm going to follow up with them in an issue here: https://github.com/dotnet/vscode-dotnet-runtime/issues/2082 -- the pipeline appears to be broken and will be a bit of a blocker to deal with that in this PR now, so I've isolated these fixes.

nagilson:
> Changes look good to me. Thanks for the detailed PR description. The only thought I had was about the design, which doesn't need to be changed for this PR. But my idea was that instead of just having a utility-style DotnetPathFinder, you'd instead make platform-specific implementation classes that know how to resolve the path for those platforms. Basically, turning the imperative `os.platform() === 'win32'` logic into a `WindowsPathResolver` or some such, and make ones for the other OS's. I believe this design is used for `LinuxVersionResolver` and thought it might fit for PATH too.

I like the idea of having this be isolated on a per OS basis. A lot of the logic is somewhat shared so it might be possible to do this with an abstract class so some methods are shared. I will add this idea in a new issue too.

PR #2051, opened 2024-11-19
Add Localization

nagilson:
Test adding localization to the extension using the new L10N Method devised by VS Code.

Has some bad Japanese which is done via google translate. Proof of concept.

Need to verify strings with args and that setting the language in vscode actually impacts the language of the tool.

Some errors need to get put into the log in english and put into the user window (error handler) not in english. But they have embedded args which are part of the string that need to go into the log and are not available past the time the error is thrown, We will likely need to create our own simple localizableString class with string and the args array that the log will parse and then l10n can parse for the user facing output.

We can do the same and handle event strings in the output channel observer, etc.

Does not include @vscode/l10n @vscode/l10n-dev in the package jsons as that seems to be a - for vscode internal, and b - a global npm tool invoke-able via npx.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
+++ b/vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
@@ -574,7 +574,7 @@ suite('DotnetCoreAcquisitionExtension End to End', function()
     }
     else
     {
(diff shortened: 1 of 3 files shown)


MsTxBunny (changes requested):
ok

MsTxBunny (changes requested):
YES

MsTxBunny (changes requested):
YES

PR #2077, opened 2025-01-04
Make sure we use the correct bash interpreter

eduardolabo:
Had some trouble running the build script with zsh, fixed it by explicitly using bash.


eduardolabo:
@dotnet-policy-service agree

nagilson (approved):
Thank you for this change.
The Linux failure is due to 9.0 being missing on the CI machine, maybe its version of Linux is out of date. I will have to fix this at the repo root.

PR #2083, opened 2025-01-07
Add Timer for Requests to Support Business CDN Evaluation

nagilson:
`Edgio`, the company running CDNs that power the .NET ecosystem, has declared bankruptcy. We are conducing an analysis of CDNs across the ecosystem because of this. Providers such as `Verizon` have also held our content. Timeouts are a significant issue and while much better than before still represent a concern for us, particularly in Asia.

This PR enables data collection for these web requests. 
Related: https://github.com/dotnet/core/issues/9669

This utilizes `Interceptors` in `Axios` to time the requests. When the request is made, data is added that includes the start time of the request. The end time is then added in the response, which is by a function hook that activates as soon as the response from the server concludes. This helps us avoid possible CPU overhead time that would make the calculations incorrect ( of course it's not infallible, it's going to take some memory access time and time to read the time.) In my testing, the header-side timer was .0002 seconds faster, but that is on a fast CPU - admittedly the code was in debug mode, so that may have also made it significantly slower.  Thank you to @baronfel for suggesting the better header-oriented approach.

The real concern, is that Axios implicitly, and silently JSON-ifies objects that are of type `application/json` in memory in the `transformResponse` section before the full response is returned, which this approach avoids. This is an expensive CPU cost. This requires moving the transformation to be deferred until after the timer, regardless of the timer method.

More Testing:
Header Response:
`118.34 ms`
CPU Timer:
`120.75 ms`

```
1/8/2025, 2:12:40 PM WebRequestTime
Timer for request:
49.42
true
https://download.visualstudio.microsoft.com/download/pr/14ccbee3-e812-4068-af47-1631444310d1/3b8da657b99d28f1ae754294c9a8f426/dotnet-sdk-5.0.408-win-x64.exe
--
1/8/2025, 2:12:45 PM WebRequestTime
Timer for request:
38.54
true
https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json
--
1/8/2025, 2:13:04 PM WebRequestTime
Timer for request:
1340.98
true
https://dot.net/v1/dotnet-install.ps1
```

This data seems sufficient, for now. Our main goal is to look at network location and time as well as CDN perf at large.

### DIFF
--- a/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
@@ -778,7 +778,8 @@ export abstract class WebRequestTimer extends DotnetCustomMessageEvent {
             Message: this.eventMessage,
             DurationMs : this.durationMs,
             Finished : this.finished,
-            Url : this.url,
+            Url : this.url.replace(/\//g, '.'), // urls get redacted as paths, they are not PII able since they are shared common urls. replaceAll may not exist with certain compilers, use regex
+            // see: https://github.com/microsoft/vscode/blob/a26fe3e4666aae75fdbfaacf7be153a07bdd12e8/src/vs/platform/telemetry/common/telemetryUtils.ts#L295C20-L295C105
             Status : this.status
         };
     }
--- a/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
@@ -770,15 +770,16 @@ export class DuplicateInstallDetected extends DotnetCustomMessageEvent {
 export abstract class WebRequestTimer extends DotnetCustomMessageEvent {
 
     constructor(public readonly eventMessage: string, public readonly durationMs: string,
-        public readonly success: string, public readonly url: string
+        public readonly finished: string, public readonly url: string, public readonly status : string
     ) { super(eventMessage); }
 
     public getProperties() {
         return {
             Message: this.eventMessage,
             DurationMs : this.durationMs,
-            Success : this.success,
-            Url : this.url
+            Finished : this.finished,
+            Url : this.url,
+            Status : this.status
         };
     }
 }
--- a/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
@@ -793,6 +793,11 @@ export class WebRequestCachedTime extends WebRequestTimer
     public readonly eventName = 'WebRequestCachedTime';
 }
 
+export class WebRequestTimeUnknown extends WebRequestTimer
+{
+    public readonly eventName = 'WebRequestTimeUnknown';
+}
+
 export class EmptyDirectoryToWipe extends DotnetCustomMessageEvent {
     public readonly eventName = 'EmptyDirectoryToWipe';
 }
--- a/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
@@ -767,6 +767,23 @@ export class DuplicateInstallDetected extends DotnetCustomMessageEvent {
     public readonly eventName = 'DuplicateInstallDetected';
 }
 
+export class WebRequestTime extends DotnetCustomMessageEvent {
+    public readonly eventName = 'WebRequestTime';
+
+    constructor(public readonly eventMessage: string, public readonly durationMs: string,
+        public readonly success: string, public readonly url: string
+    ) { super(eventMessage); }
+
+    public getProperties() {
+        return {
+            Message: this.eventMessage,
+            DurationMs : this.durationMs,
+            Success : this.success,
+            Url : this.url
+        };
+    }
+}
+
 export class EmptyDirectoryToWipe extends DotnetCustomMessageEvent {
     public readonly eventName = 'EmptyDirectoryToWipe';
 }
--- a/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
@@ -767,8 +767,7 @@ export class DuplicateInstallDetected extends DotnetCustomMessageEvent {
     public readonly eventName = 'DuplicateInstallDetected';
 }
 
-export class WebRequestTime extends DotnetCustomMessageEvent {
-    public readonly eventName = 'WebRequestTime';
+export abstract class WebRequestTimer extends DotnetCustomMessageEvent {
 
     constructor(public readonly eventMessage: string, public readonly durationMs: string,
         public readonly success: string, public readonly url: string
@@ -782,6 +781,16 @@ export class WebRequestTime extends DotnetCustomMessageEvent {
             Url : this.url
         };
     }
+}
+
+export class WebRequestTime extends WebRequestTimer
+{
+    public readonly eventName = 'WebRequestTime';
+}
+
+export class WebRequestCachedTime extends WebRequestTimer
+{
+    public readonly eventName = 'WebRequestCachedTime';
 }
 
 export class EmptyDirectoryToWipe extends DotnetCustomMessageEvent {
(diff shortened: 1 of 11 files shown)


vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
```diff
@@ -767,6 +767,37 @@ export class DuplicateInstallDetected extends DotnetCustomMessageEvent {
     public readonly eventName = 'DuplicateInstallDetected';
 }
 
+export abstract class WebRequestTimer extends DotnetCustomMessageEvent {
+
+    constructor(public readonly eventMessage: string, public readonly durationMs: string,
+        public readonly success: string, public readonly url: string
+    ) { super(eventMessage); }
```
Forgind, line 774: Why do you not need to set anything from the other parameters?

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -73,12 +77,50 @@ export class WebRequestWorker
...
+            // res.request headers is undefined. Need to access it via the member. Could not find any documentation or other variable access to do this.
+
+            // _header has all of the headers, separated by \r\n, with kv pair Name: Content
+            // Split it into arrays of headers, trim the whitespace and empty headers, then filter for the startTime header only, then grab the only result (startTime header array) and its value.
+            // Could be done more efficiently.
+            let startTime;
+            try
+            {
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+                startTime = BigInt(res.request._header?.split('\r\n')?.filter((headerLine : string) => headerLine !== '')?.map(
```
Forgind, line 101: Does this need to work for other line endings ('\n')? Would there ever be a header that starts with whitespace or a bom or something?

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -112,8 +155,35 @@ export class WebRequestWorker
...
         }, this.websiteTimeoutMs);
 
+        // Make the web request
         const response = await this.client.get(url, { signal: timeoutCancelTokenHook.signal, ...options });
+        // Timeout for Web Request -> Not Timer. (Don't want to introduce more CPU time into timer)
         clearTimeout(timeout);
+
+        // CDN Timer
+        // Standard timeout time in NS : 60,000,000,000 is < than std max_safe_int_size: 9,007,199,254,740,991
+        const timerPrecision = 2; // decimal places for timer result
```
Forgind, line 165: Two decimal places for the time in nanoseconds? That's very precise

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -73,12 +77,50 @@ export class WebRequestWorker
...
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+                    (value : any) => value.trim())
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+                ).filter((headerAndValue : any) => headerAndValue[0] === 'startTime')[0][1]);
+            }
+            catch(e : any)
+            {
+                // protect against being intercepted and having a weird header type
+            }
+            res.headers.startTime = startTime;
```
Forgind, line 113: Where is this actually used? Is it just added to a database somewhere?

Forgind (commented):
I don't know enough to approve, but I left a few questions 🙂

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
richlander: millisecond level precision is good enough.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -73,12 +77,50 @@ export class WebRequestWorker
...
+        // Register this so it happens before the cache https://axios-cache-interceptor.js.org/guide/interceptors#explanation
+        uncachedAxiosClient.interceptors.request.use( req => {
+            req.headers.startTime = process.hrtime.bigint();
+            return req;
+        });
+
+        // Record when the server responds:
+        // Register this so it happens after the cache -- this means we need to check if the result is cached before reporting perf data!
+        // ^ the request should not be processed if it is in the cache, it seems nonsensical to check this before the cache is hit even though this is possible
+        uncachedAxiosClient.interceptors.response.use(res =>
```
baronfel, line 90: IIRC this will only handle succeeded requests (the 'resolve' delegate) - I think you'd also need to specify a delegate for the 'faulted' promise delegate if we needed to trace failed responses.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
```diff
@@ -73,12 +77,50 @@ export class WebRequestWorker
...
-            this.client = setupCache(uncachedAxiosClient,
-                {
-                    storage: buildMemoryStorage(),
-                    ttl: this.cacheTimeToLive
-                }
-            );
+        // Record when the web requests begin:
+        // Register this so it happens before the cache https://axios-cache-interceptor.js.org/guide/interceptors#explanation
+        uncachedAxiosClient.interceptors.request.use( req => {
+            req.headers.startTime = process.hrtime.bigint();
```
baronfel, line 83: I'm skittish about setting a header - this will be sent to a server. You should be able to modify the request object directly and set a field on it to retrieve later. e.g.:

```
// in the request interceptor
req.timings.start = process.hrtime.bigint();
....
// in the response interceptor
let end = process.hrtime.bigint();
let delta = end - res.config.timings.start;
```

If this works we don't leak any other data to any server.

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
nagilson: Good question! Those will get auto-populated as members of the class as required by the constructor. The only other one goes into the parent / super constructor the same way.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
nagilson: That's a great point. That could cause an issue. I wouldnt expect it but its possible in theory. Im going to try Chet's approach without the headers.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
nagilson: The final precision is for ms, so I think this should be good.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
nagilson: This is added in the web request and fetched later. But I think we dont want to keep that.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
nagilson: Fantastic observation, thank you for catching that.

I tried to do this the traditional way before headers, but Axios seems to have broken the ability to modify the object directly https://github.com/axios/axios/issues/5573. And their documentation kinda sucks, or maybe I didnt search it well enough. Now it looks like you can have a lot of wrapper methods and interfaces or you can change the data or config and store it there. So ill try that.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
nagilson: That's a great point. We have our own timeout abort controller, but I'm not sure if that will handle responses that get rejected with something like 401/404. So, I should write a handler for that.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
nagilson: Ok maybe Im silly. Idk. Maybe this was all very JS-centric and its just always been a typescript thing where you have to make a bunch of wrapper types and functions to make the compiler not mad. I'm just going to make it a bunch of any's.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorker.ts
nagilson: This is a great of example of why its worth reading everything in the medium article even if some of it is boring and just explains the code verbatim. Thank you!

nagilson (commented):
I think this is ready again and feedback is resolved :fire:

baronfel (approved):

Issue #2082, opened 2025-01-06
Fix whitespace nitpicks introduced in #2048


nagilson:
So that we can merge https://github.com/dotnet/vscode-dotnet-runtime/pull/2048, I'm writing this issue to resolve the nitpicks here. We'd like to merge this PR now for release, but the pipeline needs to be remedied as well. To stick to the release cycle and avoid fixing the pipeline in that PR, I've created an issue here to followup on the nits.

nagilson:
Also https://github.com/dotnet/vscode-dotnet-runtime/pull/2048#issuecomment-2573950230 see this for a refactoring suggestion.

PR #2088, opened 2025-01-13
Improve Perf by Conditioning Taskkill + Fix whitespace and other code style no's

nagilson:
Please see, this resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2082 for code clean up

I'll use this as an opportunity to introduce a needed perf improvement per a request from another team as well since we are in the business of improving the code.

taskkill is currently called all of the time. I added that task kill only to kill timed out installer processes (https://github.com/dotnet/vscode-dotnet-runtime/pull/1859/commits/43f7494777f3d757134801f1f268735c9dd334a6#r1669397622), but the code seems to run at times when I wouldn't expect it to. It likely doesn't need to be run at all, I can put a condition on it.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
@@ -438,7 +438,8 @@ with options ${JSON.stringify(options)}.`));
 
             const commandResult : proc.SpawnSyncReturns<string> = proc.spawnSync(command.commandRoot, command.commandParts, options);
 
-            if(os.platform() === 'win32' && options.runtimeExtensionKillTask === true)
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            if(os.platform() === 'win32' && options?.runtimeExtensionKillTask === true)
             {
                 proc.spawn('taskkill', ['/pid', commandResult.pid.toString(), '/f', '/t']);
             }
--- a/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
@@ -438,7 +438,7 @@ with options ${JSON.stringify(options)}.`));
 
             const commandResult : proc.SpawnSyncReturns<string> = proc.spawnSync(command.commandRoot, command.commandParts, options);
 
-            if(os.platform() === 'win32')
+            if(os.platform() === 'win32' && options.runtimeExtensionKillTask === true)
             {
                 proc.spawn('taskkill', ['/pid', commandResult.pid.toString(), '/f', '/t']);
             }
--- a/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
@@ -9,7 +9,8 @@ import * as os from 'os';
 import open = require('open');
 import path = require('path');
 
-import {
+import
+{
     EventCancellationError,
     CommandExecutionEvent,
     CommandExecutionNoStatusCodeWarning,
@@ -41,7 +42,7 @@ import {
     EventBasedError,
     TriedToExitMasterSudoProcess
 } from '../EventStream/EventStreamEvents';
-import {exec as execElevated} from '@vscode/sudo-prompt';
+import { exec as execElevated } from '@vscode/sudo-prompt';
 import * as lockfile from 'proper-lockfile';
 import { CommandExecutorCommand } from './CommandExecutorCommand';
 import { getInstallFromContext } from './InstallIdUtilities';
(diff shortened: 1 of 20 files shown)


vscode-dotnet-runtime-library/src/Utils/FileUtilities.ts
```diff
@@ -12,172 +12,176 @@ import * as crypto from 'crypto';
...
+    FileToWipe,
+    SuppressedAcquisitionError
 } from '../EventStream/EventStreamEvents';
 
 export class FileUtilities extends IFileUtilities
 {
-   public async writeFileOntoDisk(scriptContent: string, filePath: string, alreadyHoldingLock = false, eventStream? : IEventStream)
-   {
-       eventStream?.post(new DotnetFileWriteRequestEvent(`Request to write`, new Date().toISOString(), filePath));
+public async writeFileOntoDisk(scriptContent: string, filePath: string, alreadyHoldingLock = false, eventStream? : IEventStream)
```
MiYanni, line 30: ?? It is in the class but you un-indented it (and everything else inside the class).

richlander: "FormatOnSave"?

vscode-dotnet-runtime-library/src/Utils/FileUtilities.ts
```diff
@@ -12,172 +12,176 @@ import * as crypto from 'crypto';
...
-               eventStream?.post(new SuppressedAcquisitionError(error, `Failed to delete ${f} when marked for deletion.`));
-           }
-       });
-   }
+        // Use rimraf to delete all of the items in a directory without the directory itself.
+        fs.readdirSync(directoryToWipe).forEach(f =>
+        {
+            try
+            {
+                eventStream?.post(new FileToWipe(`The file ${f} may be deleted.`))
```
Forgind, line 120: Since we know two lines down whether f is going to be deleted, can we just wait until we know for sure?

nagilson: This is for the log to show whether it detected the file or not. If it doesnt try to delete the file it wont show up in the log

vscode-dotnet-runtime-library/src/Utils/FileUtilities.ts
nagilson: Good point, this file has always been weirdly indented. I had format on save enabled but it looks like it got disabled at some point. Thank you.

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -438,7 +438,8 @@ with options ${JSON.stringify(options)}.`));
 
             const commandResult : proc.SpawnSyncReturns<string> = proc.spawnSync(command.commandRoot, command.commandParts, options);
 
-            if(os.platform() === 'win32')
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            if(os.platform() === 'win32' && options?.runtimeExtensionKillTask === true)
```
joeloff, line 442: Are you killing the bundle process? That may not necessarily trigger a proper rollback because the individual installers are running under the installer service. Also, some installers may have rollback boundaries. just calling that out as you may end up in a bad state with partially installed pieces.

nagilson: I thought I was but NO. 

I added something to send the SIGTERM response on a timeout of 10 minutes by default for when users didnt interact with the prompt. We use passive and quiet so that doesnt happen much. However, the installer taking more than 10 minutes, being unresponsive, and our code assuming it's failed, does. Then we try to retry and fail because another installer is still in use by the old process. That's why I added that.

Turns out it seems to never have worked as intended. The PID we get here is the CMD PID. The installer is spawned by the CMD and there doesnt seem to be a super easy way to get the PID without spawning some shell to run it and getting that shell's PID which is whole other thing.

Im just going to remove it.

MiYanni (approved):
Whitespace looks good now. :shipit: 

Do keep in mind that the style (seems) to be identical to standard C# style, but applied to TypeScript, which is completely fine to do (it is your repo). However, it would just be very strange for someone who develops in TypeScript to read this style. Looking at my buddy's code who has been writing TypeScript as his primary language for a decade looks like this: 

image

So, like curly braces on the same line, and 2-space indents and stuff like that.

nagilson:
> Whitespace looks good now. :shipit:
> 
> Do keep in mind that the style (seems) to be identical to standard C# style, but applied to TypeScript, which is completely fine to do (it is your repo). However, it would just be very strange for someone who develops in TypeScript to read this style. Looking at my buddy's code who has been writing TypeScript as his primary language for a decade looks like this:
> 
> image
> 
> So, like curly braces on the same line, and 2-space indents and stuff like that.

Thanks. You're right in that the style I chose is a bit unconventional. I don't like the { on same line style. I always think back to the famous disastrous apple bug that occurred due to that coding style. Its somewhat subjective and since I wrote most of the code I kinda just picked what I wanted.. whether that's valid or not is up for debate

Issue #2095, opened 2025-01-22
VSCode .net installer extension requires user to disable powershell scripting excetion policy in advance


jacdavis:
### Describe the bug.

Powershell disables script execution by default. The .net installer vscode extension is implemented as a powershell extension. That means it cannot work unless the user manually disables ps execution policy in advance. For users without powershell experience, this isn't a great experience. 

REQUEST: execute the script with execution policy by-passed. (powershell -ExcecutionPolicy Bypass squid.ps1). Admin is still required to install, but this will get rid of the confusing error message unless powershell is disabled by group policy

### Please attach log file(s) here if you have any.

Powershell disables script execution by default. The .net installer vscode extension is implemented as a powershell extension. That means it cannot work unless the user manually disables ps execution policy in advance. For users without powershell experience, this isn't a great experience. 

REQUEST: execute the script with execution policy by-passed. (powershell -ExcecutionPolicy Bypass squid.ps1). Admin is still required to install, but this will get rid of the confusing error message unless powershell is disabled by group policy

### Reproduction Steps

Powershell disables script execution by default. The .net installer vscode extension is implemented as a powershell extension. That means it cannot work unless the user manually disables ps execution policy in advance. For users without powershell experience, this isn't a great experience. 

REQUEST: execute the script with execution policy by-passed. (powershell -ExcecutionPolicy Bypass squid.ps1). Admin is still required to install, but this will get rid of the confusing error message unless powershell is disabled by group policy

### Expected Behavior

Powershell disables script execution by default. The .net installer vscode extension is implemented as a powershell extension. That means it cannot work unless the user manually disables ps execution policy in advance. For users without powershell experience, this isn't a great experience. 

REQUEST: execute the script with execution policy by-passed. (powershell -ExcecutionPolicy Bypass squid.ps1). Admin is still required to install, but this will get rid of the confusing error message unless powershell is disabled by group policy

### Exceptions & Errors (if any)

Powershell disables script execution by default. The .net installer vscode extension is implemented as a powershell extension. That means it cannot work unless the user manually disables ps execution policy in advance. For users without powershell experience, this isn't a great experience. 

REQUEST: execute the script with execution policy by-passed. (powershell -ExcecutionPolicy Bypass squid.ps1). Admin is still required to install, but this will get rid of the confusing error message unless powershell is disabled by group policy

### .NET Install Tool Version

v0.0.0

### Your operating system? Please provide more detail, such as the version and or distro above.

N/A

[...]

PR #2096, opened 2025-01-23
Fix Powershell Execution Policy & Language Mode Logic

nagilson:
# Info

For https://github.com/dotnet/vscode-dotnet-runtime/issues/2095 

See my comment at the end for context on what this all about.

* Fixes the error which said to change your execution policy, when it was actually about language mode. See below.
* Adds bypass mode when calling the install script instead of restricted. This helps with a limited number of scenarios when default execution policy restrictions are applied by SysAdmins to the User Scope, but not when restrictions are applied MachineWide via group policy as Disabled Execution in the UI, which shows up as restricted.
* Separates the error into two -- so that a clear message is given when the execution policy is set and causing a failure, vs when the language mode is set to cause failure.
* Helps us understand the 'Validation failed' error because we learned from this the directory is missing and the install result can return 0 even if the execution policy restricts the installation and it did not install correctly. This error is actually about execution policy.
* Please see below for a more thorough investigation and testing results.

# Claim: Machine Policy is defaulted to not allow script execution. 
I check this by running on a new Windows VM (not sandbox since that has its own caveats.)
`Get-ExecutionPolicy -List`

VM:
image

That didn't seem to be the case. This is azure VM. When I used 'Default' as the value it was indeed 'Restricted.'

# Claim: You can run unrestricted but even on a signed file if its downloaded from the web like the install script, it would timeout asking for a prompt for the user to continue without bypass:

I made sure the policy was set again machine wide:
- Edit Group Policy MachineWide to Disable
image

`Set-ExecutionPolicy -ExecutionPolicy Default -Scope LocalMachine`
image

I made sure to run this on a non elevated vscode. I started VS Code after the policy.

VM:
image

It actually failed with folder DNE. 
no timeout.
# Claim: If you set it to bypass, it will work correctly.

- VM: 
- Failed with 
```

Error : (DotnetCommandFailed)
Failed to download .NET 8.0.12~x64:
.NET Acquisition Failed: Validation of .dotnet installation for version {"installId":"8.0.12~x64","version":"8.0.12","architecture":"x64","isGlobal":false,"installMode":"runtime"} failed: Expected installation folder c:\Users\noahgilson004\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.vscode-dotnet-runtime\.dotnet\8.0.12~x64 does not exist., Error: Validation of .dotnet installation for version {"installId":"8.0.12~x64","version":"8.0.12","architecture":"x64","isGlobal":false,"installMode":"runtime"} failed: Expected installation folder c:\Users\noahgilson004\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.vscode-dotnet-runtime\.dotnet\8.0.12~x64 does not exist.
	at c.assertOrThrowError (c:\Users\noahgilson004\.vscode\extensions\ms-dotnettools.vscode-dotnet-runtime-2.2.6\dist\extension.js:2:80422)
	at c.validateDotnetInstall (c:\Users\noahgilson004\.vscode\extensions\ms-dotnettools.vscode-dotnet-runtime-2.2.6\dist\extension.js:2:79912)
	at y.<anonymous> (c:\Users\noahgilson004\.vscode\extensions\ms-dotnettools.vscode-dotnet-runtime-2.2.6\dist\extension.js:2:18291)
	at Generator.next (<anonymous>)
	at s (c:\Users\noahgilson004\.vscode\extensions\ms-dotnettools.vscode-dotnet-runtime-2.2.6\dist\extension.js:2:11504)
	at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

```

The same is true if system is default.
Inspection of the log shows that trying to run the script fails.

```
STDERR:  & : File C:\Users\noahgilson004\.vscode\extensions\ms-dotnettools.vscode-dotnet-runtime-2.2.6\dist\install 
scripts\dotnet-install.ps1 cannot be loaded because running scripts is disabled on this system. For more information, 
see about_Execution_Policies at https:/go.microsoft.com/fwlink/?LinkID=135170.
At line:1 char:137
+ ... ]::Tls12; & 'c:\Users\noahgilson004\.vscode\extensions\ms-dotnettools ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : SecurityError: (:) [], PSSecurityException
    + FullyQualifiedErrorId : UnauthorizedAccess

```

However, if the group policy is not set and it is not applied machine wide, then bypass actually makes it work, when it used to fail.

image

image

I will also run this through various vendor testing scenarios to confirm.

# Context

There are language modes, which constrain powershell language features, and then execution policies, which constrain running scripts in general. I have had to set flags to enable running scripts by default from a powershell terminal, but there is no documentation indicating anything but that the language mode default is full language, or unrestricted. I also tried a sandbox and it was Full by default. 

'FullLanguage is the default language mode for default sessions on all versions of Windows.'
 https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_language_modes?view=powershell-7.4&viewFallbackFrom=powershell-7.3#:~:text=The%20FullLanguage%20mode%20permits%20all%20language%20elements%20in%20the%20session.%20FullLanguage%20is%20the%20default%20language%20mode%20for%20default%20sessions%20on%20all%20versions%20of%20Windows.

What causes our error to trigger is not `Get-ExecutionPolicy` but `ExecutionContext.SessionState.LanguageMode`.

We should try to set the execution policy to `bypass` instead of `restricted` in our code as it prompts a dialogue for permission by default on unrestricted which could be the cause of unknown failures/timeouts, which is an awesome insight from @jacdavis. Even if the script is signed. 
Sadly this option does not seem to exist to bypass language mode like how you can for 1 script via execution policy.

The error that spawned this issue triggers by checking the language mode:
```
Your machine policy ConstrainedLanguage disables PowerShell language features that may be needed to install .NET. Read more at: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_language_modes?view=powershell-7.3. If you cannot safely and confidently change the execution policy, try setting a custom existingDotnetPath following our instructions here: https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md.
```
There is code I wrote that just fails it if ExecutionContext.SessionState.LanguageMode is set to ConstrainedLanguage or NoLanguage.
https://github.com/dotnet/vscode-dotnet-runtime/blob/ef5299b38ed9f5e18bfd56d8f54d0b28961630fd/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts#L195

I dont think those 2 to 4 daily people you see with that error are going to go away. The code is failing on them because it sees their language mode is restricted which is not by default and is set by a SysAdmin. If the install script team knows if we can run the script in a way with a constrained language mode, then we could do that, but that sounds impossible.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
@@ -61,7 +61,7 @@ You will need to restart VS Code after these changes. If PowerShell is still not
         {
             try
             {
-                let windowsFullCommand = `powershell.exe -NoProfile -NonInteractive -NoLogo -ExecutionPolicy bypass -Command "& { [Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12; & $ExecutionContext.SessionState.LanguageMode = "ConstrainedLanguage"; ${installCommand} }"`;
+                let windowsFullCommand = `powershell.exe -NoProfile -NonInteractive -NoLogo -ExecutionPolicy bypass -Command "& { [Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12; & ${installCommand} }"`;
                 let powershellReference = 'powershell.exe';
                 if (winOS)
                 {
--- a/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
@@ -84,7 +84,7 @@ You will need to restart VS Code after these changes. If PowerShell is still not
                         if (this.looksLikeBadExecutionPolicyError(stderr))
                         {
                             const badPolicyError = new EventBasedError('PowershellBadExecutionPolicy', `Your powershell execution policy does not allow script execution, so we can't automate the installation.
-Please read more at https:/go.microsoft.com/fwlink/?LinkID=135170`);
+Please read more at https://go.microsoft.com/fwlink/?LinkID=135170`);
                             this.eventStream.post(new PowershellBadExecutionPolicy(badPolicyError, install));
                             reject(badPolicyError);
                         }
--- a/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
@@ -90,7 +90,7 @@ Please read more at https:/go.microsoft.com/fwlink/?LinkID=135170`);
                         }
                         if ((this.looksLikeBadLanguageModeError(stderr) || error?.code === 1) && this.badLanguageModeSet(powershellReference))
                         {
-                            const badModeError = new EventBasedError('PowershellBadLanguageMode', `Your Language Mode disables PowerShell language features needed to install .NET. Read more at: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_language_modes?view=powershell-7.3.
+                            const badModeError = new EventBasedError('PowershellBadLanguageMode', `Your Language Mode disables PowerShell language features needed to install .NET. Read more at: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_language_modes.
 If you cannot change this flag, try setting a custom existingDotnetPath via the instructions here: https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md.`);
                             this.eventStream.post(new PowershellBadLanguageMode(badModeError, install));
                             reject(badModeError);
--- a/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
@@ -19,6 +19,7 @@ import
     DotnetOfflineFailure,
     EventBasedError,
     PowershellBadExecutionPolicy,
+    PowershellBadLanguageMode,
 } from '../EventStream/EventStreamEvents';
 
 import { timeoutConstants } from '../Utils/ErrorHandler'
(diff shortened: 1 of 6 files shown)


vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -163,52 +169,52 @@ You will need to restart VS Code after these changes. If PowerShell is still not
...
+            if (!command)
             {
                 knownError = true;
                 const err = Error(this.noPowershellError);
                 error = err;
             }
 
             // Check Execution Policy
-            const execPolicyOutput = cp.spawnSync(command!.commandRoot, [`-command`, `$ExecutionContext.SessionState.LanguageMode`], {cwd : path.resolve(__dirname), shell: true});
+            const execPolicyOutput = cp.spawnSync(command!.commandRoot, [`-command`, `$ExecutionContext.SessionState.LanguageMode`], { cwd: path.resolve(__dirname), shell: true });
```
JanProvaznik, line 199: I'd rename this variable, to prevent the confusion that executionpolicy and languagemode are different concepts.

JanProvaznik (approved):

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
nagilson: That's a great point, thank you for that suggestion.

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -120,7 +141,41 @@ You will need to restart VS Code after these changes. If PowerShell is still not
...
+        */
+
+        // Unexpected Token may also appear
+
+        return stderr.includes('FullyQualifiedErrorId') || stderr.includes('unexpectedToken')
+    }
+
+    private badLanguageModeSet(powershellReference: string): boolean
+    {
+        const languageModeOutput = cp.spawnSync(powershellReference, [`-command`, `$ExecutionContext.SessionState.LanguageMode`], { cwd: path.resolve(__dirname), shell: true });
```
nagilson, line 172: This should be os locked logic or in a try catch

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -56,60 +61,76 @@ You will need to restart VS Code after these changes. If PowerShell is still not
...
                             this.eventStream.post(new DotnetAcquisitionScriptOutput(install, `STDERR: ${TelemetryUtilities.HashAllPaths(stderr)}`));
-                    }
-                    if (error)
-                    {
-                        if (!(await WebRequestWorker.isOnline(installContext.timeoutSeconds, this.eventStream)))
+                        }
+                        if (this.looksLikeBadExecutionPolicyError(stderr))
+                        {
+                            const badPolicyError = new EventBasedError('PowershellBadExecutionPolicy', `Your powershell execution policy does not allow script execution, so we can't automate the installation.
+Please read more at https:/go.microsoft.com/fwlink/?LinkID=135170`);
```
MiYanni, line 87: Uh oh. The URL is malformed. It should be:
```suggestion
Please read more at https://go.microsoft.com/fwlink/?LinkID=135170`);
```

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -56,60 +61,76 @@ You will need to restart VS Code after these changes. If PowerShell is still not
...
+                            this.eventStream.post(new PowershellBadExecutionPolicy(badPolicyError, install));
+                            reject(badPolicyError);
+                        }
+                        if ((this.looksLikeBadLanguageModeError(stderr) || error?.code === 1) && this.badLanguageModeSet(powershellReference))
+                        {
+                            const badModeError = new EventBasedError('PowershellBadLanguageMode', `Your Language Mode disables PowerShell language features needed to install .NET. Read more at: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_language_modes?view=powershell-7.3.
```
MiYanni, line 93: Url doesn't need specific PowerShell version in it.
```suggestion
                            const badModeError = new EventBasedError('PowershellBadLanguageMode', `Your Language Mode disables PowerShell language features needed to install .NET. Read more at: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_language_modes.
```

MiYanni (approved):
Couple of URL changes suggested but everything else looks good.

nagilson:
Thanks very much :) 
Good catch! My browser is smart enough to fix the url but not all of them might be.

PR #2102, opened 2025-01-27
Exclude Bad Artifacts Dir for Component Gov

nagilson:
Tells 1ES to ignore the .vscode-test content for CG.

image

Handlebars is causing CG alerts because of a vulnerability, and it is only added via vscode-test which is the test framework that loads an instance of vscode to run extension features under to test. We are not in control of this. After contact with the vscode team,  the .vscode test package.json describes the built-in handlebars extension, not the handlebars npm package that has the vulnerability, so it should be excluded as it is not a real usage of it.

CG is auto injected in each task so I had to add it to each job.

Previous iteration:

~~Once a manual, configured CG task runs, the injected one still runs without configuration, but then it skips submitting the results because it's already submitted a result. Hopefully this will override the 1ES auto injected task because I can't find a way to configure that one. I've also asked them if that's possible but haven't heard back yet.~~

~~First Task:~~
```
Submitting 701 registrations.
Submitted 701 registrations.
Took 0.9606174999999999 seconds to submit the registrations.
```

~~Injected Task:~~
```
Submitting 701 registrations.
Snapshot already exists, skipping registration.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,4 +49,3 @@ vscode-dotnet-runtime-library/install scripts/dotnet-install.*
 vscode-dotnet-runtime-extension/LICENSE.txt
 vscode-dotnet-sdk-extension/LICENSE.txt
 .vscode/launch.json
-rimraf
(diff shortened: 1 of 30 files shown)


nagilson:
Linux Failure resolved in https://github.com/dotnet/vscode-dotnet-runtime/pull/2096/files

MiYanni (approved):

PR #2108, opened 2025-01-29
Switch to long URL for install script

richlander:
We should use the long URL. Turns out that the short URL requires a trip to origin.


nagilson (approved):

nagilson:
Thank you for this!

PR #2110, opened 2025-01-30
Introduce Caching to the Command Executor

nagilson:
# Motivation

Every time we run a command we spawn a new child process, which is expensive. Commands may differ a lot based on the CWD, shell, and other factors. However, with many extensions relying on us, C# and C# DevKit for example, often make us call 'dotnet_executable_path --info' with the same settings. This is a lot of overhead which can be cached. That's one of many benefits that was made in us owning and sharing this code in the ecosystem in the first place.

We can't cache it for too long because if someone edits the system, the result of dotnet --info may change. Edits to the PATH won't get picked up because all process spawns from vscode will be children of the main process, so they won't get those updates. The host output can change though as long as the host was on the PATH.

Most of the calls to info etc happen at the launch of vscode which is also where perf is more vital.

# Changes

We add a setting to disable the cache or make it extended longer. This pattern allows caching in the future.

We rely on node-cache which is a reputable library and MIT license. Adds this and updates some dependencies.

For everything that we want to be cached, we send the TTL for that command to the executor via options.dotnetInstallToolCacheTtlMs. That is separated into a different commit. Certain types of commands get an assigned number to make it easier to adjust them on the fly. The numbers are moderately arbitrary and subject to change (e.g. I assume most calls to dotnet --info happen in the first 5 seconds, and it will take about 5 seconds for that system state to change where a new version is installed.)

# Testing & Experimentation

The easiest way to test this is to open our extension log and see that things get cached and that changes depending on the ttl setting.


vscode-dotnet-runtime-extension/package.json
```diff
@@ -102,6 +102,11 @@
 				"dotnetAcquisitionExtension.allowInvalidPaths": {
 					"type": "boolean",
 					"description": "If you'd like to continue using a .NET path that is not meant to be used for an extension and may cause instability (please read above about the existingDotnetPath setting) then set this to true and restart."
+				},
+                "dotnetAcquisitionExtension.cacheTimeToLiveMultiplier": {
```
nagilson, line 106: This doesnt appear to be visible to the user.

nagilson: Apparently if the json is tab indented but space intended on this line then it does not show up in the ui 🤷

vscode-dotnet-runtime-extension/package.json
```diff
@@ -102,6 +102,11 @@
 				"dotnetAcquisitionExtension.allowInvalidPaths": {
 					"type": "boolean",
 					"description": "If you'd like to continue using a .NET path that is not meant to be used for an extension and may cause instability (please read above about the existingDotnetPath setting) then set this to true and restart."
+				},
+				"dotnetAcquisitionExtension.cacheTimeToLiveMultiplier": {
+					"type": "number",
+					"default": 1,
+					"description": "To improve performance, the results of checking .NET Installations may be cached. If you're facing issues with an install not being detected, try setting this to 0.5, or 0; or increasing the number to improve performance. Restart to change."
```
nagilson, line 109: We should write a test to make sure this doesn't fail if its set to 0, and that this works.

vscode-dotnet-runtime-library/src/Utils/FileUtilities.ts
```diff
@@ -247,31 +251,32 @@ export class FileUtilities extends IFileUtilities
...
+    public async isElevated(context: IAcquisitionWorkerContext, utilContext: IUtilityContext): Promise<boolean>
     {
+        const executor = new CommandExecutor(context, utilContext);
         if (os.platform() !== 'win32')
         {
             try
             {
-                const commandResult = proc.spawnSync('id', ['-u']);
-                return commandResult.status === 0;
+                const commandResult = await executor.execute(CommandExecutor.makeCommand('id', ['-u']), { dotnetInstallToolCacheTtlMs: SYSTEM_INFORMATION_CACHE_DURATION_MS });
```
nagilson, line 261: This could be SYS_CMD_SEARCH_CACHE_DURATION_MS, but this is gonna be wiped when vscode restarts anyways which would be required to elevate it. I guess that's one bad caveat if someone upgrades the cache to be persistent in the future.

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -379,35 +380,55 @@ ${(commandOutputJson as CommandExecutorResult).stderr}.`),
...
-        // Remove this when https://github.com/typescript-eslint/typescript-eslint/issues/2728 is done
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-        if (options && !options?.shell)
-        {
-            // Remove this when https://github.com/typescript-eslint/typescript-eslint/issues/2728 is done
+            if (!options?.cwd)
+            {
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+                options.cwd = path.resolve(__dirname);
+            }
```
MiYanni, line 399: nit Q: would this work?
```suggestion
            options.cwd ??= path.resolve(__dirname);
```

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -379,35 +380,55 @@ ${(commandOutputJson as CommandExecutorResult).stderr}.`),
...
+
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-            options.shell = true;
+            if (!options?.shell)
+            {
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+                options.shell = true;
+            }
```
MiYanni, line 406: nit Q: would this work?
```suggestion
            options.shell ??= true;
```

MiYanni (approved):
Looks good. Thanks for cleaning up the style of files as you were going along. Always leave a place better than when you found it. :)

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -136,6 +138,11 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
     }
     const resolvedTimeoutSeconds = timeoutValue === undefined ? defaultTimeoutValue : timeoutValue;
     const proxyLink = extensionConfiguration.get<string>(configKeys.proxyUrl);
+
+    // Create a cache with the TTL setting that we can only reasonably access from here.
+    const cacheTimeToLiveMultiplier = Math.abs(Number(extensionConfiguration.get<string>(configKeys.cacheTimeToLiveMultiplier)));
+    const _localCache = LocalMemoryCacheSingleton.getInstance(cacheTimeToLiveMultiplier ?? 1);
```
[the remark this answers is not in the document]
nagilson: Hm that's a good point. This is also a number type in vscode so it should be a number regardless - Number(0) is null.

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -136,6 +138,11 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
     }
     const resolvedTimeoutSeconds = timeoutValue === undefined ? defaultTimeoutValue : timeoutValue;
     const proxyLink = extensionConfiguration.get<string>(configKeys.proxyUrl);
+
+    // Create a cache with the TTL setting that we can only reasonably access from here.
+    const cacheTimeToLiveMultiplier = Math.abs(Number(extensionConfiguration.get<string>(configKeys.cacheTimeToLiveMultiplier)));
```
nagilson, line 143:
```suggestion
    const cacheTimeToLiveMultiplier = Math.abs(Number(extensionConfiguration.get<string>(configKeys.cacheTimeToLiveMultiplier ?? 1)));

```

vscode-dotnet-runtime-extension/src/extension.ts
nagilson, line 144:
```suggestion
    const _localCache = LocalMemoryCacheSingleton.getInstance(cacheTimeToLiveMultiplier);

```

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
nagilson: Thanks, that's a great simplification.

Also I forgot, the word key is a total nono in telemetry, it's all going to be destroyed. I need to rename that in the upcoming PR.

Issue #2126, opened 2025-02-10
DotnetInvalidReleasesJSONError - dotnet-sdk-9.0.102-win-x64.exe is hosted on an unexpected domain.


mrbeppo:
while installing the SDK .NET 9 (9.0.102) from Visual Studio, an error message is shown to the user.

DotnetInvalidReleasesJSONError

**Error**
The url: https://builds.dotnet.microsoft.com/dotnet/Sdk/9.0.102/

[DotNetAcquisition-ms-dotnettools.vscode-dotnet-runtime-1739176723902.txt](https://github.com/user-attachments/files/18730812/DotNetAcquisition-ms-dotnettools.vscode-dotnet-runtime-1739176723902.txt)

We cannot verify that .NET downloads are hosted in a secure location, so we have rejected .NET. The url should be download.visualstudio.microsoft.com.

flowoeB:
same

baronfel:
Did you mean Visual Studio Code? Can you try uninstalling and reinstalling this extension to clear it's cache/storage? The `releases.json` is cached for some time and a reinstall should purge the local cache

nagilson:
Thank you for the report and sorry for this issue. This is due to a breaking change on releases.json. We need to release an emergency hot fix to respond to this change so downloading 9.0 will be supported again.

nagilson:
Actually that got reverted so this should be working again but it still needs to be fixed so they can do the break again.

Issue #2125, opened 2025-02-09
dotnet 9.0.102


atray716:
### Describe the bug.

error acquiring .NET

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

Trying to create dotnet ServiceDefault.sln

### Expected Behavior

Create a New ServiceDefault.sln Document

### Exceptions & Errors (if any)

" Error Acquiring .NET "

### .NET Install Tool Version

9.0.102

### Your operating system? Please provide more detail, such as the version and or distro above.

Windows

### VS Code Version

2022

Issue #2123, opened 2025-02-08
Enable installing binaries from multiple domains


richlander:
Related: https://github.com/dotnet/vscode-dotnet-runtime/issues/2121

The following are safe domains:

- `download.visualstudio.microsoft.com`
- `builds.dotnet.microsoft.com`

This code needs to change:

https://github.com/dotnet/vscode-dotnet-runtime/blob/674f3683c5d01d48583bc5c3624a2559cc32866c/vscode-dotnet-runtime-library/src/Acquisition/GlobalInstallerResolver.ts#L250-L255

Blocking: https://github.com/dotnet/core/pull/9724

baronfel:
Frankly I'm not sure we should be applying any additional checks on the content of the releases.json file. If a) we have that file, and b) we can use a MS public signature to verify the contents (this would require publishing sidecar signature files) then we should just trust it.

richlander:
That is a good point. We should be doing something similar to:

https://github.com/dotnet/core/issues/7641#issuecomment-2002216104

I can imagine two things:

- Validate the signature / checksum.
- Validate that the domain ends in microsoft.com.

Validating the specific domain seems like tighter coupling (as demonstrated) than we need/want.

richlander:
I'm hoping to get checksums deployed beside each asset. I have a tracking issue for it. I'm hope that will happen this month.

nagilson:
Extracting the domain from a string is actually a pretty hard problem with a lot of edge cases. I think I'm going to add `builds.dotnet.microsoft.com` to the list of acceptable terms for now as a stopgap and then if the checksums come out this month we can change to doing that.

richlander:
StartsWith has a lot going for it w/rt simplicity.

PR #2129, opened 2025-02-10
Respond to Breaking Change in Releases.Json

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2126 Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2125

Workaround fix for https://github.com/dotnet/vscode-dotnet-runtime/issues/2123

We have a list of possible safe domains as part of our threat model which is needed to verify the source of truth when we run executables with elevated permission. In the past releases json only hosted downloads on `download.visualstudio.microsoft.com` but now it can be `builds.dotnet.microsoft.com`. I've added some more urls to our azure front door and other CDNs in the event that we decide to change to those.

The long term fix for this would be if signatures are ever published for releases.json to verify those instead, but that is not ready yet and this change needs to go in.

This is blocking a change: https://github.com/dotnet/core/pull/9724 -- all of dotnet as the product is broken when we update releases.json, until this is fixed.


vscode-dotnet-runtime-library/src/Acquisition/GlobalInstallerResolver.ts
```diff
@@ -221,36 +224,39 @@ export class GlobalInstallerResolver {
...
 The version may be Out of Support, or the releases json format used by ${indexUrl} may be invalid and the extension needs to be updated.`),
                                 getInstallFromContext(this.context));
                             this.context.eventStream.post(releaseJsonErr);
                             throw releaseJsonErr.error;
                         }
-                        if(!(installerUrl as string).startsWith('https://download.visualstudio.microsoft.com/'))
+                        if (!this.startsWithAny((installerUrl as string), [
+                            'https://download.visualstudio.microsoft.com/', 'https://builds.dotnet.microsoft.com/', 'https://dotnetcli.trafficmanager.net/',
+                            'https://builds.dotnet.microsoft.com.edgesuite.net/', 'https://dotnetcli.azureedge.net/', 'https://dotnetcli.blob.core.windows.net/',
+                        ]))
```
richlander, line 256:
```suggestion
                            'https://download.visualstudio.microsoft.com/', 'https://builds.dotnet.microsoft.com/', 'https://ci.dot.net',
                       'https://dotnetcli.blob.core.windows.net/',
                        ]))
```

These are the only URLs that I can imagine ever using. We don't want the other URLs being documented via code and never plan to use them directly.

richlander (approved):

PR #2134, opened 2025-02-13
Code Cleanup: Improve Null Handling

nagilson:
This is a code clean up PR including 2 commits and fixes:

1.
I noticed in the telemetry some number of errors due to trying to access content-headers when headers was null. I don't know why axios would return a response but with undefined headers: https://github.com/axios/axios/issues/4812 based on this it may be a bug of some sort. Anyway, we want to handle that and I didn't know you could chain ?.['index'].

At the same time, many calls checked if something !== null. null is falsey, as is undefined, as is calling foo?.something when foo or something is undefined as that short-circuit returns undefined: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining. Now there is some logic where the difference between null and undefined matter and I didnt touch those. Or cases where it easier to read but those are far and few between.

2.
Sometimes things can end up being undefined even if their typescript type is only string and not string | undefined. This was demonstrated in the above commit where axios object headers could be undefined even though their type did not declare as such. It's certainly possible to circumvent the type system. We do handle some objects with any types due to the dynamic nature of objects which could change form (I dont think this was the best decision but it is what it is, and I don't see us prioritizing a refactor.) Some of these types are safe but to be more cautious and or make the compiler happy, I've done this.

If you execute string?.length === 0, if the string is undefined, undefined === 0 is not true. So I've changed this to (string?.length ?? 0) -> it needs to go in parans since >/==/< operators go first over ??. It's not needed for some of them like if its > 0 since undefined is not > 0 but I'd rather be safe on all of them. Generally in a check like this I would want undefined lengths to be treated as 0. This is pretty much safe guarding our code somewhat to handle this.

There is an additional check at the end to disable other extensions when testing locally inside vscode so they dont interefere, which is what I was facing when testing this.

### DIFF
--- a/sample/src/extension.ts
+++ b/sample/src/extension.ts
@@ -71,7 +71,7 @@ export function activate(context: vscode.ExtensionContext)
 
             const result = cp.spawnSync(dotnetPath, helloWorldArgs);
             const stderr = result?.stderr?.toString();
-            if (stderr?.length > 0)
+            if ((stderr?.length ?? 0) > 0)
             {
                 vscode.window.showErrorMessage(`Failed to run Hello World:
 ${stderr}`);
@@ -80,7 +80,8 @@ ${stderr}`);
 
             const appOutput = result?.stdout?.toString();
             vscode.window.showInformationMessage(`.NET Output: ${appOutput}`);
-        } catch (error)
+        }
+        catch (error)
         {
             vscode.window.showErrorMessage((error as Error).toString());
         }
@@ -101,7 +102,8 @@ ${stderr}`);
         {
             await vscode.commands.executeCommand('dotnet.showAcquisitionLog');
             await vscode.commands.executeCommand('dotnet.acquire', { version, requestingExtensionId, mode: installMode });
-        } catch (error)
+        }
+        catch (error)
         {
             vscode.window.showErrorMessage((error as Error).toString());
         }
@@ -133,7 +135,8 @@ ${stderr}`);
             await vscode.commands.executeCommand('dotnet.showAcquisitionLog');
             const status = await vscode.commands.executeCommand<IDotnetAcquireResult>('dotnet.acquireStatus', { version, requestingExtensionId });
             vscode.window.showInformationMessage(status === undefined ? '.NET is not installed' : `.NET version ${version} installed at ${status.dotnetPath}`);
-        } catch (error)
+        }
+        catch (error)
         {
             vscode.window.showErrorMessage((error as Error).toString());
         }
@@ -166,7 +169,8 @@ ${stderr}`);
                 // Await here so we can detect errors
                 await promise;
             }
-        } catch (error)
+        }
+        catch (error)
         {
             vscode.window.showErrorMessage((error as Error).toString());
         }
@@ -188,7 +192,8 @@ ${stderr}`);
         try
         {
             await vscode.commands.executeCommand('dotnet.showAcquisitionLog');
-        } catch (error)
+        }
+        catch (error)
         {
             vscode.window.showErrorMessage((error as Error).toString());
         }
@@ -292,7 +297,8 @@ ${JSON.stringify(result) ?? 'undefined'}`);
         {
             await vscode.commands.executeCommand('dotnet-sdk.showAcquisitionLog');
             await vscode.commands.executeCommand('dotnet-sdk.acquire', { version, requestingExtensionId });
-        } catch (error)
+        }
+        catch (error)
         {
             vscode.window.showErrorMessage((error as Error).toString());
         }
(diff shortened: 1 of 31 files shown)


vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -118,7 +118,7 @@ If you cannot change this flag, try setting a custom existingDotnetPath via the
                                 reject(newError);
                             }
                         }
-                        else if (stderr && stderr.length > 0)
+                        else if (stderr?.length > 0)
```
nagilson, line 121:
```suggestion
                        else if ((stderr?.length ?? 0) > 0)

```

This is correct previously too but this feels easier to read to me. Subjective.

sample/src/extension.ts
```diff
@@ -65,74 +70,91 @@ export function activate(context: vscode.ExtensionContext) {
...
-        try {
+    const sampleDotnetUninstallAllRegistration = vscode.commands.registerCommand('sample.dotnet.uninstallAll', async () =>
+    {
+        try
+        {
             await vscode.commands.executeCommand('dotnet.uninstallAll');
             vscode.window.showInformationMessage('.NET runtimes uninstalled.');
-        } catch (error) {
+        } catch (error)
+        {
```
MiYanni, line 152: nit: catch to new line

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -174,133 +176,133 @@ export class InstallTrackerSingleton
...
     {
-        await this.executeWithLock( false, this.installingVersionsId, async () =>
+        await this.executeWithLock(false, this.installingVersionsId, async () =>
         {
             // This does not uninstall global things yet, so don't remove their ids.
             const installingVersions = await this.getExistingInstalls(false, true);
             const remainingInstallingVersions = installingVersions.filter(x => x.dotnetInstall.isGlobal);
             await this.extensionState.update(this.installingVersionsId, remainingInstallingVersions);
-        }, );
+        },);
```
MiYanni, line 200: nit Is this normal to have comma-nothing?

vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
```diff
@@ -66,16 +71,19 @@ export class OutputChannelObserver implements IEventStreamObserver {
...
                 this.inProgressVersionDone(acquisitionCompleted.install.installId);
 
-                if (this.inProgressDownloads.length > 0) {
+                if ((this.inProgressDownloads?.length ?? 0) > 0)
+                {
                     const completedVersionString = `'${this.inProgressDownloads.join('\', \'')}'`;
                     this.outputChannel.append(`Still downloading .NET version(s) ${completedVersionString} ...`);
-                } else {
+                } else
+                {
```
MiYanni, line 79: nit: separate line for else

MiYanni (approved):

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
nagilson: It is normal, mostly a personal preference IMO, it doesn't do anything. Good question

vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
```diff
@@ -66,16 +71,19 @@ export class OutputChannelObserver implements IEventStreamObserver {
 
                 this.inProgressVersionDone(acquisitionCompleted.install.installId);
 
-                if (this.inProgressDownloads.length > 0) {
+                if ((this.inProgressDownloads?.length ?? 0) > 0)
+                {
                     const completedVersionString = `'${this.inProgressDownloads.join('\', \'')}'`;
                     this.outputChannel.append(`Still downloading .NET version(s) ${completedVersionString} ...`);
-                } else {
+                } else
```
nagilson, line 78:
```suggestion
                }
                else
```

sample/src/extension.ts
```diff
@@ -65,74 +70,91 @@ export function activate(context: vscode.ExtensionContext) {
...
-    const sampleDotnetUninstallAllRegistration = vscode.commands.registerCommand('sample.dotnet.uninstallAll', async () => {
-        try {
+    const sampleDotnetUninstallAllRegistration = vscode.commands.registerCommand('sample.dotnet.uninstallAll', async () =>
+    {
+        try
+        {
             await vscode.commands.executeCommand('dotnet.uninstallAll');
             vscode.window.showInformationMessage('.NET runtimes uninstalled.');
-        } catch (error) {
+        } catch (error)
```
nagilson, line 151:
```suggestion
        }
        catch (error)
```

PR #2135, opened 2025-02-13
Use a random distribution for cache reporting

nagilson:
So, the information we have about caching is useful to see if it's helpful or not and giving a perf boost/causing issues. But also, we are sending like 500k of them per hour and the data team is not very enthused about this. We only need a small random sample to get what we need so this is a better use of $

suppressTelemetry is checked in the TelemetryObserver to not send it.

### DIFF
--- a/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
@@ -1098,7 +1098,7 @@ export class CachePutEvent extends DotnetCustomMessageEvent
 
     public getProperties()
     {
-        return { Message: this.eventMessage, indexStr: this.indexStr, value: this.value, ttl: this.ttl };
+        return { Message: this.eventMessage, indexStr: this.indexStr, value: this.value, ttl: this.ttl, ...getDisabledTelemetryOnChance(1) };
     };
 }
 
@@ -1109,7 +1109,7 @@ export class CacheGetEvent extends DotnetCustomMessageEvent
 
     public getProperties()
     {
-        return { Message: this.eventMessage, indexStr: this.indexStr, value: this.value };
+        return { Message: this.eventMessage, indexStr: this.indexStr, value: this.value, ...getDisabledTelemetryOnChance(1) };
     };
 }
 
@@ -1753,4 +1753,9 @@ export class TestAcquireCalled extends IEvent
     {
         return undefined;
     }
+}
+
+function getDisabledTelemetryOnChance(percentIntToSend: number): { [disableTelemetryId: string]: boolean }
+{
+    return { suppressTelemetry: Math.random() * 100 < percentIntToSend ? true : false };
 }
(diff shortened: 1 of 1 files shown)


vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
```diff
@@ -1754,3 +1754,8 @@ export class TestAcquireCalled extends IEvent
         return undefined;
     }
 }
+
+function getDisabledTelemetryOnChance(percentIntToSend: number): { [disableTelemetryId: string]: boolean }
+{
+    return { suppressTelemetry: Math.random() * 100 < percentIntToSend ? true : false };
```
MiYanni, line 1760: Took a peek at what Math.random() does in Javascript. So, this'll only be a value between 0 and 99 inclusive. Which is why it works for `< 1` as 1%. However, that predicate is a boolean itself. So, the `? true : false` is redundant.
```suggestion
    return { suppressTelemetry: Math.random() * 100 < percentIntToSend };
```

MiYanni (approved):
Minor suggestion but everything else looks fine.

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
nagilson: Great point, ty.

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
```diff
@@ -1754,3 +1754,8 @@ export class TestAcquireCalled extends IEvent
         return undefined;
     }
 }
+
+function getDisabledTelemetryOnChance(percentIntToSend: number): { [disableTelemetryId: string]: boolean }
+{
+    return { suppressTelemetry: Math.random() * 100 < percentIntToSend };
```
Forgind, line 1760: Forgive me if this is stupid, but as I understand it, this adds an extra property to the object you're returning: suppressTelemetry. suppressTelemetry is presumably consumed (if present) at send time and, if true, doesn't send the telemetry. If false, it sends it. Is that correct?

If so, I think this is backwards. This should give a 99% chance of suppressTelemetry being false and a 1% chance of it being true, so you'd be sending 99% of data.

nagilson: Thank you, that's a fantastic observation. This correctly does something 'x%' of the time as it's 'true' x% of the time. But it should send the data 1% of the time, which means it should suppress data 99% of the time.

https://github.com/dotnet/vscode-dotnet-runtime/pull/2139/commits/6a664c492b1a326af2ccd92468fe56339a62a626 I fixed it here.

Btw this helped me find another problem: math.random does not include 1 in its output so it will never be 100% if you ever tried that. I fixed that too.

Forgind: > Btw this helped me find another problem: math.random does not include 1 in its output so it will never be 100% if you ever tried that. I fixed that too.

This isn't really meaningful in practice, and it's how most .random() functions are implemented. If you know you should return a random n-bit number between 0 and 1, and you have a random bit generator, you just return 0.bit1bit2bit3...bitn, which can't return 1. The only range that isn't covered is between (1- 2^-n) and 1, so if you want a chance closer to 1% with n=16, instead of saying Math.random() < 0.01, you'd say Math.random() < 0.00999985, but few care about that level of specificity, and even that isn't truly 1%. Out of the 65536 possible values Math.random() could return, you'd really want 655.36 of them to return true, which is impossible to get.

Issue #2138, opened 2025-02-14
Support rollForward global.json version matching pattern in findPath API


nagilson:
### Is your feature request related to a problem? Please describe.

The rollForward value can be disabled, major, minor, latestPatch etc. Due to this issue with aspire: https://github.com/dotnet/aspire/issues/6701, we have been asked to findPaths only with runtime 9.0.2 or higher. Our >= requirement only is sufficient for allowing any 9.0 patch >= that patch, so 10 would be allowed.

### A clear and concise description of what you want to happen. Include any alternative solutions you've considered.

^

### Describe the solution you'd like

Add a different option that you can provide in the find path context.

PR #2139, opened 2025-02-14
Add additional version spec options to match rollforward from global.…

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2138

Need to add some tests, this logic is very WIP
As requested by @lifengl

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
@@ -2,19 +2,21 @@
 *  Licensed to the .NET Foundation under one or more agreements.
 *  The .NET Foundation licenses this file to you under the MIT license.
 *--------------------------------------------------------------------------------------------*/
+import * as os from 'os';
+import { DotnetFindPathDidNotMeetCondition, DotnetUnableToCheckPATHArchitecture } from '../EventStream/EventStreamEvents';
 import { IDotnetFindPathContext } from '../IDotnetFindPathContext';
 import { CommandExecutor } from '../Utils/CommandExecutor';
+import { FileUtilities } from '../Utils/FileUtilities';
 import { ICommandExecutor } from '../Utils/ICommandExecutor';
 import { IUtilityContext } from '../Utils/IUtilityContext';
-import { IDotnetListInfo } from './IDotnetListInfo';
+import { DOTNET_INFORMATION_CACHE_DURATION_MS, SYS_CMD_SEARCH_CACHE_DURATION_MS } from './CacheTimeConstants';
 import { IAcquisitionWorkerContext } from './IAcquisitionWorkerContext';
 import { IDotnetConditionValidator } from './IDotnetConditionValidator';
+import { IDotnetListInfo } from './IDotnetListInfo';
 import * as versionUtils from './VersionUtilities';
-import * as os from 'os';
-import { FileUtilities } from '../Utils/FileUtilities';
-import { DotnetFindPathDidNotMeetCondition, DotnetUnableToCheckPATHArchitecture } from '../EventStream/EventStreamEvents';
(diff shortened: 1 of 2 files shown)


vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
```diff
@@ -152,40 +167,55 @@ Please set the PATH to a dotnet host that matches the architecture ${requirement
...
-                switch (requirement.versionSpecRequirement)
+                switch (adjustedVersionSpec)
                 {
                     case 'equal':
                         return availableMinor === requestedMinor;
                     case 'greater_than_or_equal':
                         return availableMinor >= requestedMinor;
                     case 'less_than_or_equal':
                         return availableMinor <= requestedMinor;
+                    case 'latestPatch':
```
joeloff, line 200: Should there be `default` cases?

vscode-dotnet-runtime-library/src/test/unit/DotnetConditionValidator.test.ts
```diff
@@ -84,4 +85,119 @@ suite('DotnetConditionValidator Unit Tests', () => {
...
     });
+
+    test('It does not take newer major SDK if latestPatch or feature used', async () =>
+    {
+        const conditionValidator = new DotnetConditionValidator(acquisitionContext, utilityContext, mockExecutor);
+        const contextThatCanBeIgnoredExceptMode = lodash.cloneDeep(acquisitionContext.acquisitionContext);
+        contextThatCanBeIgnoredExceptMode.mode = 'sdk';
+
+        let isAccepted = conditionValidator.stringVersionMeetsRequirement('9.0.100', '8.0.201', { acquireContext: contextThatCanBeIgnoredExceptMode, versionSpecRequirement: 'latestPatch' });
+        assert.isNotTrue(isAccepted, 'It doesnt take 9.0 sdk for 8.0 latestPatch');
```
joeloff, line 96: missing apostrophe

nagilson, line 96:
```suggestion
        assert.isNotTrue(isAccepted, 'It doesn't take 9.0 sdk for 8.0 latestPatch');
```

nagilson: fix typo

vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
```diff
@@ -152,40 +167,55 @@ Please set the PATH to a dotnet host that matches the architecture ${requirement
                 const availablePatchStr: string | null = requirement.acquireContext.mode !== 'sdk' ? versionUtils.getRuntimePatchVersionString(availableVersion, this.workerContext.eventStream, this.workerContext)
                     : versionUtils.getSDKCompleteBandAndPatchVersionString(availableVersion, this.workerContext.eventStream, this.workerContext);
                 const availablePatch = availablePatchStr ? Number(availablePatchStr) : null;
-                switch (requirement.versionSpecRequirement)
+
+                const availableBandStr = requirement.acquireContext.mode === 'sdk' ? versionUtils.getFeatureBandFromVersion(availableVersion, this.workerContext.eventStream, this.workerContext, false) ?? null : null;
```
MiYanni, line 171: So, is `?? null` saying, "If it is null or undefined, return null"? If `getFeatureBandFromVersion` cannot return `undefined`, it doesn't do anything.

MiYanni (approved):

vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
nagilson: This was supposed to handle a return value of '' which evaluates to false in JS. But even though '' is false, that doesn't mean ?? will convert it to null -- good catch.

PR #2151, opened 2025-02-28
Improve Performance by Utilizing Shells Properly

nagilson:
There are many ways to execute commands in nodejs via `child_process`. Unfortunately, we were using `spawnSync` which blocks the entire thread (all of vscode extensions share one thread) which is highly inefficient.

I learned MANY things by reading various docs and such about all of the ways to do this to figure out the 'best' thing to do for our situation.

Here are some other fun improvments:

- You can spawn with powershell.exe shell! Running it with cmd (default shell) spawns a shell which spawns a shell, which is a huge unnecessary perf hit.

- You can go shell-less when running direct file executables to improve performance.

What more we could do (this is already a huge boost):
- The WSL check calls sync methods still because it needs to be called in the activate function to disable commands from showing up in WSL that are not supported. This makes startup slow. I would rather just have worse UX for WSL by allowing the command to appear but saying it doesn't work so it doesn't make vscode way slower for everybody.

Try to find places where executables can be used to reduce shell spawns.

# Ways to Run Commands

exec - spawns a shell. returns callback with error, stderr, stdout after waiting for it to complete. Any exit code other than 0 will be an error. error.signal for when a signal terminated it.

util.promisify(function) turns a callback into await or thenable (same thing) object with stderr or stdout and rejects with error with same object but with an error instead with exec.

execFile - same as exec but shellless - must be a file to exec instead of any generic command. No shell spawn by default thus more efficient as you should give it the path to the executable instead of expecting it to be found by the shell. Utilize when possible

spawn - shelless by default (which is faster if you dont need a shell but both have this option) spawns a new process with open stream communication for the command (shell or no shell) and streams data live. meant for 'interactive' commands and is better for huge data streams as you can chunk it.

fork - spawn a new process that can communicate with the parent, no shell support

spawnSync - Blocks the Entire Node Thread until its done. Same with execSync.

shelljs - a library for running native unix commands cross-plat

The ways to access the 'signal' or 'status code' are different and while spawn provides an easier interface to get to them in theory, if you read the docs on the callback return object possibilities from node you will see its applicable in either situation.

certain commands are implemented in nodejs such as chmod

# more reading
https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback

https://stackoverflow.com/questions/14332721/node-js-spawn-child-process-and-get-terminal-output-live https://stackoverflow.com/questions/48698234/node-js-spawn-vs-execute

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
@@ -22,26 +22,26 @@ import
     PowershellBadLanguageMode,
 } from '../EventStream/EventStreamEvents';
 
-import { timeoutConstants } from '../Utils/ErrorHandler'
-import { InstallScriptAcquisitionWorker } from './InstallScriptAcquisitionWorker';
 import { TelemetryUtilities } from '../EventStream/TelemetryUtilities';
-import { FileUtilities } from '../Utils/FileUtilities';
 import { CommandExecutor } from '../Utils/CommandExecutor';
+import { timeoutConstants } from '../Utils/ErrorHandler';
+import { FileUtilities } from '../Utils/FileUtilities';
+import { InstallScriptAcquisitionWorker } from './InstallScriptAcquisitionWorker';
 
 import { IUtilityContext } from '../Utils/IUtilityContext';
-import { IAcquisitionWorkerContext } from './IAcquisitionWorkerContext';
-import { IAcquisitionInvoker } from './IAcquisitionInvoker';
-import { IDotnetInstallationContext } from './IDotnetInstallationContext';
-import { IInstallScriptAcquisitionWorker } from './IInstallScriptAcquisitionWorker';
+import { WebRequestWorker } from '../Utils/WebRequestWorker';
 import { DotnetInstall } from './DotnetInstall';
 import { DotnetInstallMode } from './DotnetInstallMode';
-import { WebRequestWorker } from '../Utils/WebRequestWorker';
+import { IAcquisitionInvoker } from './IAcquisitionInvoker';
+import { IAcquisitionWorkerContext } from './IAcquisitionWorkerContext';
+import { IDotnetInstallationContext } from './IDotnetInstallationContext';
+import { IInstallScriptAcquisitionWorker } from './IInstallScriptAcquisitionWorker';
 
 export class AcquisitionInvoker extends IAcquisitionInvoker
 {
     protected readonly scriptWorker: IInstallScriptAcquisitionWorker;
     protected fileUtilities: FileUtilities;
-    private noPowershellError = `powershell.exe is not discoverable on your system. Is PowerShell added to your PATH and correctly installed? Please visit: https://learn.microsoft.com/powershell/scripting/install/installing-powershell-on-windows.
+    private noPowershellError = `powershell is not discoverable on your system. Is PowerShell added to your PATH and correctly installed? Please visit: https://learn.microsoft.com/powershell/scripting/install/installing-powershell-on-windows.
 You will need to restart VS Code after these changes. If PowerShell is still not discoverable, try setting a custom existingDotnetPath following our instructions here: https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md.`
 
     constructor(private readonly workerContext: IAcquisitionWorkerContext, private readonly utilityContext: IUtilityContext)
@@ -88,7 +88,7 @@ Please read more at https://go.microsoft.com/fwlink/?LinkID=135170`);
                             this.eventStream.post(new PowershellBadExecutionPolicy(badPolicyError, install));
                             reject(badPolicyError);
                         }
-                        if ((this.looksLikeBadLanguageModeError(stderr) || error?.code === 1) && this.badLanguageModeSet(powershellReference))
+                        if ((this.looksLikeBadLanguageModeError(stderr) || error?.code === 1) && await this.badLanguageModeSet(powershellReference))
                         {
                             const badModeError = new EventBasedError('PowershellBadLanguageMode', `Your Language Mode disables PowerShell language features needed to install .NET. Read more at: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_language_modes.
 If you cannot change this flag, try setting a custom existingDotnetPath via the instructions here: https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/troubleshooting-runtime.md.`);
@@ -167,7 +167,7 @@ At dotnet-install.ps1:1189 char:5
         return stderr.includes('FullyQualifiedErrorId') || stderr.includes('unexpectedToken')
     }
 
-    private badLanguageModeSet(powershellReference: string): boolean
+    private async badLanguageModeSet(powershellReference: string): boolean
     {
         if (os.platform() !== 'win32')
         {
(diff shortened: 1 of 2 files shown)


nagilson:
I think the rejected obj isnt what I thought it was from the docs, need to investigate.

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -206,13 +214,13 @@ To keep your .NET version up to date, please reconnect to the internet at your s
             context.eventStream.post(new DotnetAcquisitionStatusResolved(install, version));
             return { dotnetPath };
         }
-        else if ((installedVersions?.length ?? 0) === 0 && fs.existsSync(dotnetPath) && installMode === 'sdk')
+        else if ((installedVersions?.length ?? 0) && await this.file.exists(dotnetPath) && installMode === 'sdk')
```
nagilson, line 217:
```suggestion
        else if ((installedVersions?.length ?? 0) === 0 && await this.file.exists(dotnetPath) && installMode === 'sdk')
```

vscode-dotnet-runtime-library/src/test/mocks/MockObjects.ts
```diff
@@ -256,27 +291,34 @@ export class MockVSCodeEnvironment extends IVSCodeEnvironment
...
 
-    protected getFallbackScriptPath(): string {
-        if (this.fallback) {
+    protected getFallbackScriptPath(): string
+    {
+        if (this.fallback)
+        {
             return path.join(__dirname, '..');
-        } else {
+        } else
```
nagilson, line 320:
```suggestion
        }
        else
```

nagilson:
For some reason the pipeline machines for Mac and Linux are leaving behind an old test file. I think this is ready for review nonetheless and I'll keep looking into this.

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -41,7 +41,7 @@ export class AcquisitionInvoker extends IAcquisitionInvoker
 {
     protected readonly scriptWorker: IInstallScriptAcquisitionWorker;
     protected fileUtilities: FileUtilities;
-    private noPowershellError = `powershell.exe is not discoverable on your system. Is PowerShell added to your PATH and correctly installed? Please visit: https://learn.microsoft.com/powershell/scripting/install/installing-powershell-on-windows.
+    private noPowershellError = `powershell is not discoverable on your system. Is PowerShell added to your PATH and correctly installed? Please visit: https://learn.microsoft.com/powershell/scripting/install/installing-powershell-on-windows.
```
joeloff, line 44: should we have aka.ms links instead? documentation tends to move around

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -237,19 +238,26 @@ At dotnet-install.ps1:1189 char:5
...
+                CommandExecutor.makeCommand('0', []),
+                CommandExecutor.makeCommand('1', []),
+                CommandExecutor.makeCommand('2', []),
+                CommandExecutor.makeCommand('3', []),
+                CommandExecutor.makeCommand('4', []),
             ];
-
+        const possiblePowershellPaths =
+            [ // use shell as powershell and see if it passes or not. This is faster than doing it with the default shell, as that spawns a cmd to spawn a pwsh
+                { shell: `%SystemRoot%\\System32\\WindowsPowerShell\\v1.0\\powershell.exe` },
```
joeloff, line 251: These variables can change. You really should be calling `SHGetSpecialFolderPathW`. Does TS not support any native APIs at all?

Q2: What is the significance of the ordering here?

joeloff: Why check for `pwsh.exe` and `pwsh`? Isn't one for Windows and the other for Mac? Should you construct this based on the OS instead?

vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
```diff
@@ -144,10 +162,10 @@ Bin Bash Path: ${os.platform() !== 'win32' ? (await this.executor?.execute(Comma
 
         options.dotnetInstallToolCacheTtlMs = DOTNET_INFORMATION_CACHE_DURATION_MS;
         const findCommand = CommandExecutor.makeCommand(pathLocatorCommand, ['dotnet']);
-        const dotnetsOnPATH = (await this.executor?.execute(findCommand, options))?.stdout.split('\n').map(x => x.trim()).filter(x => x !== '');
+        const dotnetsOnPATH = (await this.executor?.execute(findCommand, options, false))?.stdout.split('\n').map(x => x.trim()).filter(x => x !== '');
```
joeloff, line 165: is this looking at the PATH environment variable to see if there's anything called .NET or is it trying to resolve the dotnet executable for each directory in the path variable to see whether it exists?

vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
```diff
@@ -201,20 +219,30 @@ Bin Bash Path: ${os.platform() !== 'win32' ? (await this.executor?.execute(Comma
...
+        {
+            return undefined;
+        }
+        return this.getTruePath(realPaths);
     }
 
     public async findHostInstallPaths(requestedArchitecture: string): Promise<string[] | undefined>
     {
-        this.workerContext.eventStream.post(new DotnetFindPathHostFxrResolutionLookup(`Looking up .NET without checking the PATH.`));
+        this.workerContext.eventStream.post(new DotnetFindPathHostFxrResolutionLookup(`Looking up.NET without checking the PATH.`));
```
joeloff, line 245:
```suggestion
        this.workerContext.eventStream.post(new DotnetFindPathHostFxrResolutionLookup(`Looking up .NET without checking the PATH.`));

```

vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
```diff
@@ -316,7 +357,7 @@ Bin Bash Path: ${os.platform() !== 'win32' ? (await this.executor?.execute(Comma
             }
             else
             {
-                this.workerContext.eventStream.post(new DotnetFindPathNoRuntimesOnHost(`The host: ${tentativePath} does not contain a .NET runtime installation.`));
+                this.workerContext.eventStream.post(new DotnetFindPathNoRuntimesOnHost(`The host: ${tentativePath} does not contain a.NET runtime installation.`));
```
joeloff, line 360:
```suggestion
                this.workerContext.eventStream.post(new DotnetFindPathNoRuntimesOnHost(`The host: ${tentativePath} does not contain a .NET runtime installation.`));

```

vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
nagilson: Good catch!

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
nagilson, line 44:
```suggestion
    private noPowershellError = `powershell is not discoverable on your system. Is PowerShell added to your PATH and correctly installed? Please visit: https://aka.ms/install-powershell

```

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
nagilson: Q2: The ordering determines which one it checks first, and if it works then it uses that one.
image

I think your API suggestion is a good idea and the better way to go but https://github.com/search?q=SHGetSpecialFolderPathW+language%3Atypescript&type=code that doesn't seem to ever be done before. I think I will investigate this but don't want to block this PR for it.

Why check for pwsh.exe and pwsh and powershell.exe and powershell? Based on the data there is a very odd number of people where it works for one but not the other. For the (Vast) majority the system root one works.
image

joeloff: sounds good. Agree not to block on the special folder change, that should probably be looked at on its own,

nagilson: Based on my sample data literally nobody in this large set: didnt have poweshell.exe but did have pwsh.exe, so I can remove that one. But there were a small number who had powershell but not powershell.exe. Pwsh is the newer version of pwsh. This is all Windows only so why is this the case, I dont know ._.

vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
nagilson: Its using which or where to list every possible thing called 'dotnet' and later it will check if the file works and is the right version + arch of dotnet, etc etc

PR #2157, opened 2025-03-05
Do not consider SDKs when validating the runtime requirements.

JoeRobich:
The runtimes installed with SDKs are already includes in the runtimes list. Comparing runtime and SDK versions does not make sense.

Part of the fix for https://github.com/dotnet/vscode-csharp/issues/8034


JoeRobich:
reopened from a fork as https://github.com/dotnet/vscode-dotnet-runtime/pull/2158

PR #2158, opened 2025-03-05
Do not consider SDKs when validating the runtime requirements. 

JoeRobich:
The runtimes installed with SDKs are already includes in the runtimes list. Comparing runtime and SDK versions does not make sense.

Part of the fix for https://github.com/dotnet/vscode-csharp/issues/8034

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
@@ -32,7 +32,6 @@ export class DotnetConditionValidator implements IDotnetConditionValidator
         if (requirement.acquireContext.mode === 'sdk') {
             const availableSDKs = await this.getSDKs(dotnetExecutablePath);
             if (availableSDKs.some((sdk) => {
-                // The SDK includes the Runtime, ASP.NET Core Runtime, and Windows Desktop Runtime. So, we don't need to check the mode.
                 return this.stringArchitectureMeetsRequirement(hostArch, requirement.acquireContext.architecture) &&
                     this.stringVersionMeetsRequirement(sdk.version, requirement.acquireContext.version, requirement) && this.allowPreview(sdk.version, requirement);
             })) {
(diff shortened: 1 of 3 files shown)


vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
```diff
@@ -27,36 +27,32 @@ export class DotnetConditionValidator implements IDotnetConditionValidator
...
-            return true;
-        }
-        else
-        {
+        if (requirement.acquireContext.mode === 'sdk') {
             const availableSDKs = await this.getSDKs(dotnetExecutablePath);
-            if (availableSDKs.some((sdk) =>
-            {
+            if (availableSDKs.some((sdk) => {
                 // The SDK includes the Runtime, ASP.NET Core Runtime, and Windows Desktop Runtime. So, we don't need to check the mode.
```
nagilson, line 35:
```suggestion
```
Remove erroneous comment

nagilson (approved):
Thank you for making a PR for this! That was speedy and you did a great job. I appreciate the modifications to the test as well. Sorry for the misconception here, I didn't consider that the 9.0.101 sdk might still have a 9.0.100 runtime.

nagilson:
@JoeRobich Btw, how did you get this to pop up in a display from the extension host? 
image
 Is this your own logging that you added in c#?

JoeRobich:
> @JoeRobich Btw, how did you get this to pop up in a display from the extension host?

Very low tech way. Opened your $home\.vscode\extensions\msdoten-...\dist\extension.js, ran it through prettier, found the bits of code I wanted to add logging to, added `console.log()`, saved file, and reloaded vscode. haha

Now that VS Code supports output channels with log levels, it might be worth while going through and adding trace level logging for discovery and acquisition.

nagilson:
Ah, very nice 😁 The log.txt we have is very useful but this is so nice that it has the data objects instead of just a string. So thank you and thats a good suggestion

PR #2166, opened 2025-03-10
Uninstall Logic Handles Corrupt Save Data

nagilson:
# A - Proper Shared State
https://github.com/dotnet/vscode-dotnet-runtime/pull/2248, which should be reviewed first, (this contains the same changes as it builds on top of them) protects the entire installation process using the mutex. As a follow up to that, it's clear that the longstanding methods of checking whether an install was in progress or not was isolated to each individual vs code instance (promise storage in local memory - not shared, and thus prone to error or invalid logic with multiple windows open) and is no longer needed as the operation is protected under a share-able atomic lock.

# B - Handling Corrupt Save Data
In addition, we had an old thing called the 'graveyard' which held very old data from extensions that edited the vscode save state. That data could become corrupted over time and trying to read it at this point is not very relevant. This was initially added  in https://github.com/dotnet/vscode-dotnet-runtime/pull/1443 2023 in version 1.8.1 to handle extension versions that were stored as strings. It was rehauled in 2024 https://github.com/dotnet/vscode-dotnet-runtime/pull/1748 2024 when we changed the state of extension storage again. An example of corruption is here: https://github.com/microsoft/GitHub-Copilot-for-Azure/issues/163. The set of installs was treated as a 'string' and then expanded out to an extremely large list of datum. None of that data really matters anymore, so we should just try to clean it out and not process it. I will have vendors test 1.8.0, 1.8.1, 2.0.6, and now 2.3.4 which represent inflection points of data handling.

Data Formats in Extension State / Vs Code Global Storage (User Save Data Memory)

### 1.8.0:
installPathsGraveyard: undefined
    installing: string[]: `['9.0~x64']` (for installs before 1.8.0, `9.0`)
    installed: string[]
    private acquisitionPromises: { [version: string]: Promise<string> | undefined };

### 1.8.1:
installPathsGraveyard: `{ [installKeys: string]: string //path }` Example: {'8.0.0': 'dotnet/8.0.0}'
installing: string[]: `['9.0~x64']` (for installs before 1.8.0, `9.0`)
installed: string[]

local memory only: acquisitionPromises: `{ [installKeys: string]: Promise<string> | undefined }`
### 2.0.6:
```
interface LocalDotnetInstall
{
    dotnetInstall: DotnetInstall;
    // The string is the path of the install once completed.
    path: string;
}
```
installPathsGraveyard: `{ [installKeys: string]: string } | Set<LocalDotnetInstall>`
* creates a new Set<> or auto maps the old type:
```
Object.entries(graveyard).map(([key, path]) => ({ dotnetInstall: getAssumedInstallInfo(this.context, key, null), path }) as LocalDotnetInstall)
```

```
export interface InstallRecord
{
    dotnetInstall: DotnetInstall;
    installingExtensions: InstallOwner[] (`string` | null`);
}
```
installing and installed: `InstallRecord[]` ( auto converted)

local memory:
```
interface InProgressInstall
{
    dotnetInstall: DotnetInstall;
    // The string is the path of the install once completed.
    installingPromise: Promise<string>;
}
```
`Set<InProgressInstall>`

### 2.3.4:

installPathsGraveyard: undefined
installed: same
installing: gone
promises: gone

# C - Exposing API Surface

Adds `resetData` command to uninstall everything and also clear our related datum. This is a public method that could be used in scenarios like the above in B. This is not enabled by default and the setting must be checked to show it in the command pallet (ctrl shift p)

To test:
image
Build and run the extension in the workspace.
image
Go to the extension setting and enable the command:
Screenshot 2025-05-02 112307

# D - Removing Deprecated API RelatedCode

Local sdk installs are no longer needed, thus neither is the related code.

# E - Dont Uninstall if In Use

On Unix, you can do fs.promises.rm, etc, to remove the file from the file system. The fd is still held in memory by the other processes but the file will appear deleted. On Windows, you cannot. We now check during uninstall if dotnet.exe is in use to prevent wiping the runtime files, etc, when the host is in use and cannot be deleted. 

# Changes
- Remove 'installing' state.

- Remove promises store.

The promises was a way to check if an existing function was running and executing the same install. This was added at the very beginning. However, it was stored in local memory. This is not good as it is not persistent across multiple instances of vscode running at the same time. Now we have a mutex over the local storage folder that works cross-process and intra-process. The mutex should protect the entire install operation. Other similar installs will wait for the mutex to be free (or if the application dies.) Then if it is valid they will see its an existing install and validate it, then use it. If not, the invoker now wipes the install folder.

Now, one problem is validation only checks --info arch and --list-runtimes or sdks and that it matches the version requirement, as well as that the dotnet host files and main expected files are there. There is no easy way to fully validate that everything worked without enumerating every file and keeping a list of every valid file for every valid build, or rezipping it exactly how our tooling does and comparing the sha (possibly easier but still hard.) This is not a new bug that has been introduced, however, considering before the behavior must have also existed, because the 'installed' state key check only validated that the executable was present, and the check for partial installs only removed installs from the 'installing' list, not the 'installed' list.

Now, the 'installing' list is no longer relevant as it was only used for this promise holding based approach.


vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -299,9 +303,9 @@ ${installRecord.map(x => `${x.installingExtensions.join(' ')} ${JSON.stringify(I
         return executeWithLock(this.eventStream, alreadyHoldingLock, this.getLockFilePathForKey(context.installDirectoryProvider, installState), 5, 200000,
             async (installationState: InstallState, install: DotnetInstall, ctx: IAcquisitionWorkerContext) =>
             {
-                this.eventStream.post(new RemovingVersionFromExtensionState(`Adding ${JSON.stringify(install)} with id ${installationState} from the state.`));
+                this.eventStream.post(new AddTrackingVersions(`Adding ${JSON.stringify(install)} with id ${id} from the state.`));
 
-                const existingVersions = await this.getExistingInstalls(installationState, ctx.installDirectoryProvider, true);
+                const existingVersions = await this.getExistingInstalls(installState, context.installDirectoryProvider, true);
```
nagilson, line 308: is this ok?

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -142,6 +145,7 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
     }
     const resolvedTimeoutSeconds = timeoutValue === undefined ? defaultTimeoutValue : timeoutValue;
     const proxyLink = extensionConfiguration.get<string>(configKeys.proxyUrl);
+    const showResetDataCommand = extensionConfiguration.get<boolean>(configKeys.showResetDataCommand);
```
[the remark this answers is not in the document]
nagilson: This is documented in the package json and will mainly be used when we tell people to use it on GH

vscode-dotnet-runtime-library/src/Utils/FileUtilities.ts
```diff
@@ -67,29 +71,33 @@ export class FileUtilities extends IFileUtilities
...
-    public async wipeDirectory(directoryToWipe: string, eventStream?: IEventStream, fileExtensionsToDelete?: string[])
+    public async wipeDirectory(directoryToWipe: string, eventStream?: IEventStream, fileExtensionsToDelete?: string[], verifyDotnetNotInUse = false)
     {
         if (!(await this.exists(directoryToWipe)))
         {
             eventStream?.post(new EmptyDirectoryToWipe(`The directory ${directoryToWipe} did not exist, so it was not wiped.`))
             return;
         }
+        else if (verifyDotnetNotInUse && await FileUtilities.fileIsOpen(path.join(directoryToWipe, getDotnetExecutable()), eventStream))
+        {
```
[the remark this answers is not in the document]
nagilson: That's a fair point, although I think the code is pretty clear here and this comment just says what the code does. If another person would like to have this comment, happy to add it (I may have too much context since I just wrote the code)

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -845,6 +864,11 @@ We will try to install .NET, but are unlikely to be able to connect to the serve
         ensureDependenciesRegistration,
         reportIssueRegistration,
         ...eventStreamObservers);
+
+    if (showResetDataCommand)
+    {
+        vsCodeContext.subscriptions.push(resetDataPublicRegistration);
```
Forgind, line 870: Is there an intended order for these actions?

Forgind (approved):
I have to be honest here. I got very lost very fast. There This PR combines a few different things, admittedly with some related parts, but I couldn't always tell which problem certain changes were intended to address. I tried to review little pieces to see if I could build up to something deeper, but I never got all the way there. So this approval is more like "I read your PR description, and it sounded good" than "I fully understood all the code, and it's correct." Sorry about that ☹️

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: Good question -- no, the order shouldn't matter for the behavior of the application

nagilson:
> I have to be honest here. I got very lost very fast. This PR combines a few different things, admittedly with some related parts, but I couldn't always tell which problem certain changes were intended to address. I tried to review little pieces to see if I could build up to something deeper, but I never got all the way there. So this approval is more like "I read your PR description, and it sounded good" than "I fully understood all the code, and it's correct." Sorry about that ☹️

Thanks for taking a look, I think reviewing this PR would kind of suck without a lot of code context. I'm happy to do a call with either you or the team (the team because some of this logic I think would probably be relevant to discuss for the bootstrapper as it will need to implement something similar in the long-term.) I'll chat with marcpop about this.

Issue #2164, opened 2025-03-10
[NETE2ESDK] The output was inaccurate if user canceled the SDK installation


ChenhuiYuan01:
### Describe the bug.

The output was inaccurate if user canceled the SDK installation

### Please attach log file(s) here if you have any.

[logs.zip](https://github.com/user-attachments/files/19159976/logs.zip)

### Reproduction Steps

1. On windows, run the sample command ‘Install .NET SDK Globally via .NET Install Tool (Former Runtime Extension)‘.
2. Use the version it provided, which is 8.0.112. Hit enter.
3. Click cancel during the installation process.
4. Observe the output

### Expected Behavior

The output shows that SDK installation failed as the user canceled the SDK installation

### Exceptions & Errors (if any)

An error message appears
![Image](https://github.com/user-attachments/assets/a42ab095-72be-4ee8-9df7-0c45850866b2)

**Note**
A warning message shows if try to install the SDK again
![Image](https://github.com/user-attachments/assets/0f44fb5a-e8a5-4c99-b7bd-f347e257bfb2)

The installation failed even the user click 'Install Anyways'
![Image](https://github.com/user-attachments/assets/10feb1be-002f-479c-a7fd-4695f413a4ef)

### .NET Install Tool Version

2.2.9

### Your operating system? Please provide more detail, such as the version and or distro above.

N/A

### VS Code Version

1.98.0

NicoleWang001:
Fixed on the new build.
![Image](https://github.com/user-attachments/assets/996cbb35-fb20-4e82-9ae6-2250da5fa204)

PR #2169, opened 2025-03-10
Fix 2.2.9 Bugs for Release. Command Status, Web Cache, & Web Timing

nagilson:
# Improvements for 2.2.9 Release (Bugs found during Vendor Testing)

# Issue 1 - Getting the Status of a Command 

Fixes https://github.com/dotnet/vscode-dotnet-runtime/issues/2164

The nodejs spec for the special cased promisify version of exec mentions that it throws an object with an error, stdout and stderr. It looks like the object it throws is not an error in an object but rather an error object. So calling .error is incorrect.

This fixes the handling of cancellations.

image

# Issue 2 - Caching Web Requests

The old web request library was a class which each owned their own axios instance.
This used to work when the cache storage was persistent and stored within VS Code. But now that its a local memory cache, I migrated the class to a singleton so the cache actually gets used by everyone.
For why its not persistent storage, there is a bug in VSCode described here which makes doing that not work well: https://github.com/dotnet/vscode-dotnet-runtime/pull/1540

That was a very subtle bug and rare so I dont want to deal with figuring out if it's better or not in a hot fix of releases, but I figured this was simple enough that it would be ok to fit it in. (Maybe I was a bit optimistic.)

Thank you to @baronfel as well for doing some testing around with the axios cache library.

Now you can see cached events are being published, and took .02 ms instead of the longer time it'd take before
image

Each test also now has to clear the cache between each run to prevent any funkiness.

# Issue 3 - Timing Web Requests

The file download response type is a 'stream' which means the response gets sent back with an open stream to pipe the file.
This means the recorded end time for the webrequest was for the first few chunks of data and the initial stream pipe handover.
Now it waits to record the data until after that is done.

To simplify the logic, it still records every web request but recognizes if its in a stream mode not to record that data and let the end of the stream handler callback report it.
image

Now you can see it took 51 seconds instead of the average of the old data (about 2 s)


vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -110,10 +108,68 @@ export class WebRequestWorker
...
+                return true; // the class object was deleted externally (close tab / pkill from vscode ext service), so we should stop the cleanup.
+            }
+            return this.stopCacheCleanup.valueOf();
+        }
+
+        loopWithTimeoutOnCond(500, Number.POSITIVE_INFINITY,
+            this.cacheCleanupRunner,
+            function stopCleaning(): void {},
+            null,
+            new WebCacheClearEvent(`Clearing the web cache.`)
```
nagilson, line 147: this doesn't do anything but that's ok for now (event stream is null so its never posted)

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -45,29 +47,25 @@ import
...
      * The responses from GET requests are cached with a 'time-to-live' of 5 minutes by default.
      */
     private client: AxiosCacheInstance;
-    private websiteTimeoutMs: number;
-    private proxy: string | undefined;
-
+    protected static instance: WebRequestWorkerSingleton;
     private proxyAgent: HttpsProxyAgent<string> | null = null;
+    private cacheTtl: number = WEB_CACHE_DURATION_MS;
+    private stopCacheCleanup = Boolean(false); // must be object and not primitive so we can hold a reference to it
```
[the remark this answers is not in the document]
nagilson: It doesn't need to be a non primitive anymore so I changed the comment, this was a leftover from when I was using a reference object with a binding

vscode-dotnet-runtime-extension/CHANGELOG.md
```diff
@@ -11,21 +11,20 @@ and this project adheres to [Semantic Versioning].
...
-
 - Fixes an issue with the Uninstall command.
 - Improvements to how installations are managed and tracked.
-- Performance improvements by reducing network activity requirements.
-- Fixed auditing of internal CDN performance metrics.
 
 ## [2.2.9] - 2025-3-12
 
-- Widescale performance improvements throughout the entire extension. Better asynchronicity, efficient utilization of shells, eliminating process calls, improving file system / io.
+- Wide-scale performance improvements throughout the entire extension. Better asynchronicity, efficient utilization of shells, eliminating process calls, improving file system / io.
```
joeloff, line 19:
```suggestion
- Wide-scale performance improvements throughout the entire extension. Better asynchronicity, efficient utilization of shells, eliminating process calls, improving file system I/O.

```

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -302,13 +302,13 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
     });
 
     const dotnetListVersionsRegistration = vscode.commands.registerCommand(`${commandPrefix}.${commandKeys.listVersions}`,
-        async (commandContext: IDotnetListVersionsContext | undefined, customWebWorker: WebRequestWorker | undefined): Promise<IDotnetListVersionsResult | undefined> =>
+        async (commandContext: IDotnetListVersionsContext | undefined, customWebWorker: WebRequestWorkerSingleton | undefined): Promise<IDotnetListVersionsResult | undefined> =>
```
Forgind, line 305: I haven't gotten to where this is defined yet, but I'm very optimistic based on the name

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -235,16 +235,23 @@ This report should be made at https://github.com/dotnet/vscode-dotnet-runtime/is
     private async downloadInstaller(installerUrl: string): Promise<string>
     {
         const ourInstallerDownloadFolder = IGlobalInstaller.getDownloadedInstallFilesFolder(installerUrl);
-        await this.file.wipeDirectory(ourInstallerDownloadFolder, this.acquisitionContext.eventStream);
         const installerPath = path.join(ourInstallerDownloadFolder, `${installerUrl.split('/').slice(-1)}`);
 
+        if (await this.file.exists(installerPath) && await this.installerFileHasValidIntegrity(installerPath))
```
Forgind, line 240: Is it always the same installer?

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -110,10 +108,68 @@ export class WebRequestWorker
...
+                try
+                {
+                    if ((process.hrtime.bigint() - this.lastCacheCleanTimeNs) >= this.cacheTtlMs * 1e6)
+                    {
+                        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+                        (this.client.storage as any)?.clear();
+                        this.lastCacheCleanTimeNs = process.hrtime.bigint();
+                    }
+                }
+                catch (err)
```
Forgind, line 131: Can you be more precise with these errors then? Not sure if that's a thing in ts

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -125,74 +181,92 @@ export class WebRequestWorker
...
         {
-            this.context.eventStream.post(new WebRequestTime(`Timer for request:`, durationMs, 'true', url, String(response.status)));
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            ctx.eventStream.post(new WebRequestTime(`Timer for request:`, durationMs, 'true', url, String(response?.status)));
         }
         else
         {
-            this.context.eventStream.post(new WebRequestCachedTime(`Cached Timer for request:`, durationMs, 'true', url, String(response.status)));
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            ctx.eventStream.post(new WebRequestCachedTime(`Cached Timer for request:`, durationMs, 'true', url, String(response?.status)));
```
Forgind, line 265: Is it possible to give real types for these cases?

Forgind (approved):

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
nagilson: The String() constructor here will produce a real string type, so that is good, but good question :)

If you mean the response object, I dont trust any of the types that the web libraries may return as they have never followed the correct logic and bypassed the type system, returning types their API said they would never return. So sadly I dont trust it

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
nagilson: The thrown object can be any type, while we could write handlers for specific error types, the action we'd take would generally be the same. I think thats a good question though. There is likely a better way to do this and I remember briefly looking into it and not finding much, though. You can at least try to assume it will throw an Error type but that is not necessarily a requirement. If any dependency in the code chain breaks the throw error convention then it would fail unexpectedly here 🤔

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
nagilson: No, it will be different depending upon the version / architecture

Issue #2174, opened 2025-03-12
[NETE2ESDK][Windows]Failed to uninstall some .NET SDKs if run several uninstalls at the same time


NicoleWang001:
### Describe the bug.

Failed to uninstall some .NET SDKs if run several uninstalls at the same time

### Please attach log file(s) here if you have any.

[logs.zip](https://github.com/user-attachments/files/19204772/logs.zip)

### Reproduction Steps

1. On windows, open VSCode with administrator, and install .NET SDK 9.0.2xx, 9.0.1xx and 8.0.1xx via running command ‘.NET Install Tool: Install .NET SDK System-Wide' for several times
2. Run the ‘,NET Install Tool: Uninstall’ command and try to run it several times to uninstall NET SDK 9.0.2xx, 9.0.1xx and 8.0.1xx
3. Observe the uninstallation result

### Expected Behavior

All commands should succeed. 

### Exceptions & Errors (if any)

Failed to uninstall .NET 9.0.1xx and 9.0.2xx
![Image](https://github.com/user-attachments/assets/f6d4a072-2033-49a7-9ffd-42d511e27494)

### .NET Install Tool Version

v2.2.9-b

### Your operating system? Please provide more detail, such as the version and or distro above.

Windows

### VS Code Version

1.98.0

nagilson:
I believe this will be fixed by the same thing as https://github.com/dotnet/vscode-dotnet-runtime/issues/2183?reload=1

NicoleWang001:
This issue is not reproduced on the new build 2.3.0
![Image](https://github.com/user-attachments/assets/f911a868-9a9b-4eaf-9292-20183ed1242d)

nagilson:
This one also got truncated by the log. Might be due to a timeout.

NicoleWang001:
Is reproduced with v2.3.1, especially when multiple SDKs are installed by .NET Install Tool and Sample command.

[Logs.zip](https://github.com/user-attachments/files/19548249/Logs.zip)

ChenhuiYuan01:
This issue is still reproduced with the new build v 2.3.1 on win11 arm64
![Image](https://github.com/user-attachments/assets/ccb8c1ae-26d0-46d2-a0f3-10c4e2b0c1ea)

NicoleWang001:
Not repro on v2.3.7

[...]

Issue #2173, opened 2025-03-12
[NETE2ESDK][Windows].NET SDK could not be installed normally if run several installs at the same time


NicoleWang001:
### Describe the bug.

.NET SDK 9.0.114 could not be installed normally if run several installs at the same time.

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

[logs.zip](https://github.com/user-attachments/files/19204314/logs.zip)

### Reproduction Steps

1. On Windows Run a command ‘.NET Install Tool: Install .NET SDK System-Wide' to install the .NET SDK and enter the newest version 9.0.201
2. While this command is running, run the command ‘Sample: Acquire ASPNET Runtime’ 
3. While this command is running, run the command ‘Sample” concurrently acquire 2.x runtimes’ 
4. Run a command ‘.NET Install Tool: Install .NET SDK System-Wide' twice to install the .NET SDK 8.0.1xx and 9.0.1xx
5. Observe the installation

### Expected Behavior

All SDKs and Runtimes that required should be installed normally.

### Exceptions & Errors (if any)

.NET SDK 9.0.104 is not list on 'dotnet --list-sdks' and 'C:\Program Files\dotnet\sdk' even shows installed succeed in output.

![Image](https://github.com/user-attachments/assets/e4980568-9890-4d8a-ba79-ad61e3b82e9c)

Outputs:
```
user requested to download the .NET SDK.
Downloading .NET version(s) 9.0.201-global~x64 .............ms-dotnettools.sample-extension requested to download the .NET Runtime.
 -- Concurrent download of '6.0.36~x64' started!

Downloading .NET version(s) 9.0.201-global~x64, 6.0.36~x64 ...ms-dotnettools.sample-extension requested to download the .NET Runtime.
 -- Concurrent download of '7.0.20~x64' started!

Downloading .NET version(s) 9.0.201-global~x64, 6.0.36~x64, 7.0.20~x64 ...ms-dotnettools.sample-extension requested to download the .NET Runtime.
 -- Concurrent download of '8.0.14~x64' started!

Downloading .NET version(s) 9.0.201-global~x64, 6.0.36~x64, 7.0.20~x64, 8.0.14~x64 ...ms-dotnettools.sample-extension requested to download the .NET ASP.NET Runtime.
 -- Concurrent download of '7.0.20~x64~aspnetcore' started!

[...]

PR #2177, opened 2025-03-13
Hold Lock Over Machine State Changing Operations (Sync Code Is Now Async)

nagilson:
> [!TIP]
> `Hide whitespace changes` to review this PR.

# Motivation
https://github.com/dotnet/vscode-dotnet-runtime/pulls?q=is%3Apr+is%3Aclosed Introduced many bugs with concurrency and the ordering of things because this was not coded correctly. The fact that the sync methods blocked the entire thread from continuing prevented other things from operating so it was essentially a 'lock'. 

There was already a lock for when we update the records of whats installed but that isnt held and is completely separate from the execution of installing or uninstalling.

# Broad Changes
- globalMachineStateLock Now, we need to manually hold a lock when modifying global machine state ( so we dont try to launch 2 installers at once, as these msis cannot run simultaneously (error 1618) or apt-get / dnf requires a lock to be held.) 

- LocalFolderInstallLocks For local installs we also want to hold a lock so that multiple installs dont try to write to the folder at the same time, this is only a per folder lock however since it is local.

- TalkingToSudoProcessLock If you hold one of those locks, you should respect it by not touching the other things in that folder. Aka if youre also accessing sudo state then you must hold the lock.

# Individualized Changes: Sudo Prompt

The sudo commands also hold a lock to operate and communicate with a master sudo process that runs commands forwarded to it (see: threat model if this raises eyebrows) This is an anti pattern but necessary with vscode/sudo-prompt unless we want to edit the library and fork it, as it runs only 1 process at a time which makes it hard when we need to run dozens of commands with elevation, we don't want to elevate every time. And no, we can't use IPC / Fork() and talk to the process in a regular fashion without rewriting Vs Code Sudo Library and we don't want to own that.

I migrated the locking code here to be all the same. Here were some issues with the old code:

1. If someone PKills Vs Code, the lock will get into a 'bad' state where it was never unlocked. The 'lock' is just a protected file on disk that has a 'stale' timer where it's touched. If a lock is 'stale' it can be assumed to be dropped. This means we would wait for the lock to get stale (could be 10 minutes) if someone pkills vscode and reopens which is unnecessary.

2. Migrates error flow from VS Code Async Code / Sudo Master Process to a singleton that is later accessed by our process. This means we can properly control the code execution flow, where as before the error could be thrown by the sub Master Process in the middle of any code scope anywhere, which leads to unpredictable results. 

3. Protect during the entire operation of sudo, so that when we check if the sudo process is alive / dead / should we kill it / execute with it, those are all holding 1 lock to prevent race conditions.

4. Try to unlock if the lock library failed. Sometimes, it seems to try to unlock even if it never got the lock. That sucks. The best we can do is accept it may try that and fail but we can catch it. The ramification of that is that the lock would theoretically be released when someone else is holding it by a different prior holder, except for the fact that checkSync and unlock checks that the unlocker is the one who holds the lock, so it's fine. 

5. Remove the lock from the directory functions as the mental concept is that to operate in this directory, you need to hold the sudo lock anyways. 

6. We kill the sudo master process once we're done. That's to prevent it from hanging around. But, other code should not wait 10 minutes to check if the proc is still alive to do work if we killed it. So I added that logic too. 

7. If someone kills the master process we spawn it will not drop the dpkg lock which is required on Ubuntu to run apt. This also tells the apt lock to finish and drop the dpkg lock if it was killed. 

# Context and Example

This manifested in issues 
https://github.com/dotnet/vscode-dotnet-runtime/issues/2173
https://github.com/dotnet/vscode-dotnet-runtime/issues/2174

Here's an example of it now working and processing many different concurrent things at once:
```
user requested to download the .NET SDK.
Downloading .NET version(s) 9.0.201-global~x64 ....user: Trying to install .NET 9.0.201-global~x64 but it already exists. No downloads or changes were made.
 Done!
.NET 9.0.201-global~x64 executable path: C:\Program Files\dotnet\dotnet.exe

user requested to download the .NET SDK.
Downloading .NET version(s) 7.0.317-global~x64 ......user requested to download the .NET SDK.
 -- Concurrent download of '7.0.120-global~x64' started!

Downloading .NET version(s) 7.0.317-global~x64, 7.0.120-global~x64 .......user: Trying to install .NET 9.0.201-global~x64 but it already exists. No downloads or changes were made.
.ms-dotnettools.sample-extension: Trying to install .NET 8.0.14~x64~aspnetcore but it already exists. No downloads or changes were made.
ms-dotnettools.sample-extension: Trying to install .NET 7.0.20~x64 but it already exists. No downloads or changes were made.
ms-dotnettools.sample-extension: Trying to install .NET 7.0.20~x64~aspnetcore but it already exists. No downloads or changes were made.
ms-dotnettools.sample-extension: Trying to install .NET 6.0.36~x64~aspnetcore but it already exists. No downloads or changes were made.
ms-dotnettools.sample-extension requested to download the .NET Runtime.
 -- Concurrent download of '8.0.14~x64' started!

Downloading .NET version(s) 7.0.317-global~x64, 7.0.120-global~x64, 8.0.14~x64 ....ms-dotnettools.sample-extension: Trying to install .NET 6.0.36~x64 but it already exists. No downloads or changes were made.
..ms-dotnettools.sample-extension tried to install .NET 6.0.36~x64~aspnetcore but that install had already been requested. No downloads or changes were made.
..Attempting to remove .NET 9.0.201-global~x64.
.. Done!
.NET 8.0.14~x64 executable path: c:\Users\noahgilson\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.vscode-dotnet-runtime\.dotnet\8.0.14~x64\dotnet.exe

Still downloading .NET version(s) '7.0.317-global~x64', '7.0.120-global~x64' ......Attempting to remove .NET 8.0.114-global~x64.
....................................... Done!
.NET 7.0.317-global~x64 executable path: C:\Program Files\dotnet\dotnet.exe

Still downloading .NET version(s) '7.0.120-global~x64' ............................................. Done!
.NET 7.0.120-global~x64 executable path: C:\Program Files\dotnet\dotnet.exe

Attempting to remove .NET 6.0.36~x64~aspnetcore.
Uninstalled .NET 6.0.36~x64~aspnetcore.
ms-dotnettools.sample-extension: Trying to install .NET 2.0.9~x64 but it already exists. No downloads or changes were made.
ms-dotnettools.sample-extension: Trying to install .NET 2.2.8~x64 but it already exists. No downloads or changes were made.
ms-dotnettools.sample-extension: Trying to install .NET 2.1.30~x64 but it already exists. No downloads or changes were made.
ms-dotnettools.sample-extension requested to download the .NET Runtime.
Downloading .NET version(s) 2.0.9~x64 ...ms-dotnettools.sample-extension requested to download the .NET Runtime.
 -- Concurrent download of '2.2.8~x64' started!

Downloading .NET version(s) 2.0.9~x64, 2.2.8~x64 ...ms-dotnettools.sample-extension requested to download the .NET Runtime.
 -- Concurrent download of '2.1.30~x64' started!

Downloading .NET version(s) 2.0.9~x64, 2.2.8~x64, 2.1.30~x64 ..........Uninstalled .NET 8.0.114-global~x64.
 Done!
.NET 2.0.9~x64 executable path: c:\Users\noahgilson\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.vscode-dotnet-runtime\.dotnet\2.0.9~x64\dotnet.exe

Still downloading .NET version(s) '2.2.8~x64', '2.1.30~x64' .... Done!
.NET 2.2.8~x64 executable path: c:\Users\noahgilson\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.vscode-dotnet-runtime\.dotnet\2.2.8~x64\dotnet.exe

Still downloading .NET version(s) '2.1.30~x64' ... Done!
.NET 2.1.30~x64 executable path: c:\Users\noahgilson\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.vscode-dotnet-runtime\.dotnet\2.1.30~x64\dotnet.exe
```


joeloff:
This will only work within the extension. You're not going to stop installers from being run by other services like automatic updates or users running winget. The best practice for Windows is this: https://learn.microsoft.com/en-us/windows/win32/msi/-msiexecute-mutex

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -334,67 +289,69 @@ ${installRecord.map(x => `${x.installingExtensions.join(' ')} ${JSON.stringify(I
 
     protected async addVersionToExtensionState(context: IAcquisitionWorkerContext, idStr: string, installObj: DotnetInstall, alreadyHoldingLock = false)
     {
-        return this.executeWithLock(alreadyHoldingLock, idStr, async (id: string, install: DotnetInstall) =>
-        {
-            this.eventStream.post(new RemovingVersionFromExtensionState(`Adding ${JSON.stringify(install)} with id ${id} from the state.`));
+        return executeWithLock(this.eventStream, alreadyHoldingLock, this.getLockFilePathForKey(idStr), 5, 200000,
```
nagilson, line 292: nit: these retry numbers could be part of the class

nagilson:
> This will only work within the extension. You're not going to stop installers from being run by other services like automatic updates or users running winget. The best practice for Windows is this: https://learn.microsoft.com/en-us/windows/win32/msi/-msiexecute-mutex

Absolutely, thanks for calling that out. I agree and put that in a comment in the code but looks like I deleted it during a refactor. 
I think this isnt a perfect solution but its a significant improvement. 1618 being an issue was already present in the extension before we moved off sync code when other external MSIs were running, and we do have some logic retry in this case. (see `handleStatus` where it will call `executeInstall` a 2nd time... although it doesnt wait for very long). The problem this solves is the extensions own multiple requests (which happens all of the time because install api calls are generally made by other extensions at startup) not failing one another which was not a problem with sync code but is a problem with async code.

I would prefer to use msiexecute-mutex on windows but doing so in typescript seems mildly sketchy and will take a lot more work. Id like to fix this part for now since it was the regression and consider this next part later.

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
```diff
@@ -58,6 +67,58 @@ export async function isRunningUnderWSL(acquisitionContext: IAcquisitionWorkerCo
...
 
+/*
+@remarks lockPath should be a full path to a shared lock file ending in .lock (that may or may not exist on disk) and the file content does not matter
+*/
+export async function executeWithLock<A extends any[], R>(eventStream: IEventStream, alreadyHoldingLock: boolean, lockPath: string, retryTimeMs: number, timeoutTimeMs: number, f: (...args: A) => R, ...args: A): Promise<R>
+{
+    retryTimeMs = retryTimeMs > 0 ? retryTimeMs : 100;
+    const retryCountToEndRoughlyAtTimeoutMs = timeoutTimeMs / retryTimeMs;
+
+    try
```
joeloff, line 78: Are there any risks around race conditions trying to create the lock file?

nagilson: Not that I can think of. `mkdirSync` will block the entire JS thread pool from doing anything until it is done. same with WriteFileSync. Someone else could try to make it again and they may get to the lock code first but that should be fine. 

Here is an interesting situation:  that would failure if the flag `lockfilePath` is used (which we dont use):
Process A Writes the Lock File
Process B Writes the Same Lock File (OK)
Process B Locks 
Process B Unlocks (deletes the file)
Process A Locks - File DNE, Exception Thrown

As the file to lock on must exist, and lockFilePath treats the file as a lock. Luckily without this option, the file that gets deleted is the lock file created with lock() and not earlier. lockFilePath is meant to be used to lock an entire directory with a hand-over-hand lock approach to prevent this. https://www.npmjs.com/package/proper-lockfile

I also confirmed this in their source code https://github.com/search?q=repo%3Amoxystudio%2Fnode-proper-lockfile%20release&type=code

nagilson: I did just realize it wont release it if the f() call throws in some cases, fixing that...

nagilson:
I found some issues with this, marking as draft until Im more confident.

.eslintrc.js
```diff
@@ -71,10 +71,8 @@ module.exports = {
         "@typescript-eslint/no-unsafe-return": "warn",
         "@typescript-eslint/no-unsafe-call": "warn",
         "@typescript-eslint/no-explicit-any": "error",
-        "@typescript-eslint/prefer-promise-reject-errors": "warn",
```
nagilson, line 74: This is turning them from warnings to errors

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -532,7 +532,7 @@ ${WinMacGlobalInstaller.InterpretExitCode(installerResult)}`), install);
         dotnetPath = await installer.getExpectedGlobalSDKPath(installingVersion,
             context.acquisitionContext.architecture ?? this.getDefaultInternalArchitecture(context.acquisitionContext.architecture));
 
-        context.installationValidator.validateDotnetInstall(install, dotnetPath, os.platform() !== 'win32', os.platform() !== 'darwin');
+        context.installationValidator.validateDotnetInstall(install, dotnetPath, os.platform() === 'darwin', os.platform() !== 'darwin');
```
joeloff, line 535: Was looking at what this method does, confused by one of the errors this returns: `Expected executable file exists but is not a file` I'm guessing this is along the lines of you expect there to be a dotnet executable file, but instead you have a folder named dotnet

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
```diff
@@ -58,6 +65,97 @@ export async function isRunningUnderWSL(acquisitionContext: IAcquisitionWorkerCo
...
+    if (alreadyHoldingLock)
+    {
+        // eslint-disable-next-line @typescript-eslint/await-thenable
+        return f(...(args));
+    }
+
+    // Someone PKilled Vscode while we held the lock previously. Need to clean up the lock created by the lib (lib adds .lock unless you use LockFilePath option)
+    if (fs.existsSync(`${lockPath}.lock`) && !(LockUsedByThisInstanceSingleton.getInstance().hasVsCodeInstanceInteractedWithLock(lockPath)))
+    {
+        eventStream?.post(new DotnetLockReleasedEvent(`Lock about to be released, But we never touched it (pkilled vscode?)`, new Date().toISOString(), lockPath, lockPath));
```
joeloff, line 84:
```suggestion
        eventStream?.post(new DotnetLockReleasedEvent(`Lock about to be released, but we never touched it (pkilled vscode?)`, new Date().toISOString(), lockPath, lockPath));

```

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
```diff
@@ -84,12 +84,13 @@ export async function executeWithLock<A extends any[], R>(eventStream: IEventStr
...
         if (lockfile.checkSync(lockPath))
         {
+            eventStream?.post(new DotnetLockReleasedEvent(`Lock about to be released, and checkSync showed it.`, new Date().toISOString(), lockPath, lockPath));
             lockfile.unlockSync(lockPath);
         }
         else
         {
             eventStream?.post(new DotnetLockReleasedEvent(`Lock is not owned by us, delete it`, new Date().toISOString(), lockPath, lockPath));
-            fs.unlinkSync(`${lockPath}.lock`);
+            fs.rmdirSync(`${lockPath}.lock`, { recursive: true });
```
joeloff, line 93: Do you need to check if the lock is actually a dir, like fstat or something? Why delete recursively? do you expect the lock dir to have content, and if so, is there a chance any of the content could be a symlink?

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -174,17 +174,18 @@ ${stderr}`));
         let isLive = false;
 
         const processAliveOkSentinelFile = path.join(this.sudoProcessCommunicationDir, 'ok.txt');
+        const waitForLockTimeMs = maxTimeoutTimeMs ?? this.context?.timeoutSeconds ? (this.context?.timeoutSeconds * 1000 / 5) : 180000;
```
joeloff: Doesn't this imply if the timeout is set to 1s, you actually set it to 200ms?

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
nagilson: That is correct -- it also checks whether the dotnet folder exists and has an executable in it, though in this case we just check that the executable is there and a file instead of a directory called dotnet

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
nagilson: The lock should always be a dir, and the recursively is to make sure in case the logic changes in the future from the lock library that we delete anything in the dir. At the moment, I don't expect there to be any content in the directory. But, that's a good question 🤔

joeloff (approved):

PR #2182, opened 2025-03-18
Revert #2151 (Sync to Async Methods)

nagilson:
Reverts the changes in https://github.com/dotnet/vscode-dotnet-runtime/pull/2151
This PR made the sync code async. This is because sync functions block the entire thread.

These changes introduced many race conditions. I have been working on fixing them in https://github.com/dotnet/vscode-dotnet-runtime/pull/2177, but they are pretty complicated, unfortunately. There are a ton of fixes that need to be released, and this will delay the release if I keep trying to fix it, most likely. Given the large number of changes, this will also make it easier to understand if new bugs are introduced whether they are due to the other included changes or the async to sync changes. Both include a large number of code rewrites.

Once #2177 is done, I'll merge taking in everything from that branch, for the most part.


nagilson:
Hm, it looks like the race cond on linux were always present, so maybe this isnt a regression after all.

PR #2191, opened 2025-03-24
Exclude 'SDK Education Bundle' From Sample Extension as Extension Dependency

nagilson:
The sample extension tests and shows how to call the .NET Install Tool Extension and .NET SDK Education Bundle Extension.
Since the SDK Education Bundle was deprecated and archived on Dec 2nd by that team, we removed the SDK Extension for the Education Bundle from the marketplace.

The extension dependency in vscode makes it so, for this extension to load, the dependent ones must be installed. VS Code tries to fetch them from the marketplace. Since the SDK Extension is now gone, we don't want that. Instead we can include its code manually by import (which we already did when debugging locally, so this is not new code.) The .NET Install Tool remains a dependent since its a valid extension on the marketplace.

The commented out code is for when you want to debug the sample extension code locally. But if that code is enabled it causes file conflicts. It makes sense to have the pipeline be able to build the sample extension by default rather than be able to debug the sample extension locally by default, hence the runtime change.

This bumps the sample extension to match the current version of the install tool: 2.3.0.

### DIFF
--- a/sample/src/extension.ts
+++ b/sample/src/extension.ts
@@ -6,7 +6,7 @@
 import * as cp from 'child_process';
 import * as path from 'path';
 import * as vscode from 'vscode';
-import * as runtimeExtension from 'vscode-dotnet-runtime'; // comment this out when packing the extension
+// import * as runtimeExtension from 'vscode-dotnet-runtime'; // comment this out when packing the extension
 import
 {
     DotnetInstallMode,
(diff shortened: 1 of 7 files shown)


sample/src/extension.ts
```diff
@@ -6,7 +6,7 @@
 import * as cp from 'child_process';
 import * as path from 'path';
 import * as vscode from 'vscode';
-import * as runtimeExtension from 'vscode-dotnet-runtime'; // comment this out when packing the extension
+// import * as runtimeExtension from 'vscode-dotnet-runtime'; // comment this out when packing the extension
```
joeloff, line 9: silly question, but doe TS support any preprocessor directives like `#ifdef` to wrap the import around and pass a value through to the compiler to make it true for packing, but undefined for building?

joeloff (approved):

sample/src/extension.ts
nagilson: That would be pretty cool... it looks like it does not but there are tools out there that add a build step to mimic that behavior.

PR #2192, opened 2025-03-24
Fix Issue when VS Code is pkilled and we hold a lock

nagilson:
# Context

- proper_lockfile is the approach we take to have a concept of a mutex/lock in nodejs.
- This writes a dir on disk based on another existing file on disk, so if you lock foo.txt, it creates foo.txt.lock. This represents the lock.
- If someone pkills vscode, the lock is not cleaned up. The lock library will eventually make it stale, but that means the extension will fail to acquire a dead lock for the timeout time (approx 10 minutes) so it will seem like it fails.
- We can rm -rf the lock in this case by having a singleton keep track of if this instance of vscode ever touched the lock.

# Fix

- unlockSync will fail if we don't own the lock. We don't own the lock if an old process owned it as their ownership is an in-memory local dictionary. We should always just rmDir it. This has no consequences I know of on their lib based on their code, since everything is stored in local memory, it is blown away when vscode gets pkilled. Any of our sub processes would also get blown away, except for perhaps the sudo master process, which is capable of communicating with a new vscode window; thankfully, a lock is required to talk to that process. There is a slight bug possibility, where it was told to run some command and pipes the output back out to vscode if it is quickly pkilled and relaunched, but I think this is lower priority to fix.

- This also makes it so that the lockFile key in the list of touched lock files is based on the ending file name instead of the full path, to prevent issues where the path may be abbreviated / relative.

# Importance

Required for 2.3.0 to release.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Utils/LockUsedByThisInstanceSingleton.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/LockUsedByThisInstanceSingleton.ts
@@ -30,7 +30,7 @@ export class LockUsedByThisInstanceSingleton
 
     public hasVsCodeInstanceInteractedWithLock(lockKey: string): boolean
     {
-        lockKey = path.resolve(lockKey);
+        lockKey = path.basename(lockKey).trim();
         const hasInteracted = this.lockStringAndThisVsCodeInstanceOwnsIt[lockKey] === true;
         this.lockStringAndThisVsCodeInstanceOwnsIt[lockKey] = true; // This could be a set but this is also fine
         return hasInteracted;
--- a/vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
@@ -83,9 +83,15 @@ export async function executeWithLock<A extends any[], R>(eventStream: IEventStr
     {
         eventStream?.post(new DotnetLockReleasedEvent(`Lock about to be released, but we never touched it (pkilled vscode?)`, new Date().toISOString(), lockPath, lockPath));
 
-        eventStream?.post(new DotnetLockReleasedEvent(`Lock is not owned by us, delete it`, new Date().toISOString(), lockPath, lockPath));
-        fs.rmdirSync(`${lockPath}.lock`, { recursive: true });
-
+        try
+        {
+            fs.rmdirSync(`${lockPath}.lock`, { recursive: true });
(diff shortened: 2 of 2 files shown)


Forgind (commented):
What if you have two VS Code instances at the same time both trying to acquire this lock? And how are you verifying that the lock isn't held by this process?

nagilson:
That's a great point. I didn't even think about multiple vscode instances on the same machine 😵 Ack.
It has a global shared state (singleton) which keeps track of if its ever held the lock for each instance, so it wont delete something its held. I think the best I can do here is disable this deletion if theres >1 main vscode instances running at the same time, such that:

- Yes, it will be slower if you kill a vscode window and have multiple running as it will have to wait for a timeout to be able to get the lock back
- I should confirm that it can get the lock back after a timeout
- It will still not get stuck if you pkill and only have one window open
- Deleting the lock if it is in use by another process is quite bad, and leads to unexpected behavior. 

In the long term, we could: 
Find a way to manage each process PID for each vscode instance and see if it was killed or not to prevent slow behavior with multiple simultaneous windows if another window is pkilled. I don't want to deal with that RN.

Forgind:
What if you name the lock file `${lockPath}.${PID}.lock`, and if you want to acquire the lock you check to see if there are any files matching that pattern, and if so whether the associated process is currently alive?

nagilson:
PID ReUse may be a problem w/ the above approach. VS Code Session ID or some alternative may be another good solution.

Forgind:
> PID ReUse may be a problem w/ the above approach. VS Code Session ID or some alternative may be another good solution, but not sure how to expose that / check for other session ids within a different session.

That's a good point...I'll try to think about it and let you know if I come up with something.

nagilson:
Another point, is that VS Code has many subprocesses. I'm not actually sure if we can rely on the code that's running 1 instance of vscode to use a consistent PID for the entirety of its execution. The extension host may be its own process but I'm not sure yet.

nagilson:
TIL there is a 'bindings' library that lets you bind a c lib or other language lib into a nodejs module https://www.npmjs.com/package/bindings

Which vscode has used to implement a mutex: https://github.com/microsoft/vscode/blob/b25957bf5c5abae45ec0f57c46942241c64ad2a0/src/vs/code/electron-main/app.ts#L1385

Although only on Windows.

nagilson:
I'm going to talk to one of _the_ vs code experts hopefully tomorrow about this :)

nagilson:
We can implement this better with IPC Pipes. Basically, check if its alive by talking to it as it holds the lock. If it responds, great, if not, its dead. This is the way to implement locks in node.js.

I will do this in another PR, since this Invalid URL Issue appears to be more pressing and without this, we will face other problems.

PR #2199, opened 2025-03-26
Don't record all of the environment in the log

nagilson:
We only need certain information to help debug logs, and when customers submit them to us, they are less likely to redact everything. It's been helpful having this information for debugging and this will make it harder to debug in case there are variables that I didn't think of, that a user might set that'd cause a bug, but for the sake of collecting less information on the log file we should do this.

This also hashes the string of the env used as the key to cache commands, since each command environment may result in different results, we want to still have a 'unique' enough ID for it without storing this data in our cache.

### DIFF
--- a/vscode-dotnet-runtime-library/src/LocalMemoryCacheSingleton.ts
+++ b/vscode-dotnet-runtime-library/src/LocalMemoryCacheSingleton.ts
@@ -7,9 +7,11 @@
 import * as nodeCache from 'node-cache';
 import { IAcquisitionWorkerContext } from "./Acquisition/IAcquisitionWorkerContext";
 import { CacheClearEvent, CacheGetEvent, CachePutEvent } from "./EventStream/EventStreamEvents";
+import { TelemetryUtilities } from './EventStream/TelemetryUtilities';
 import { CommandExecutor } from "./Utils/CommandExecutor";
 import { CommandExecutorCommand } from "./Utils/CommandExecutorCommand";
 import { CommandExecutorResult } from "./Utils/CommandExecutorResult";
+import { minimizeEnvironment } from './Utils/TypescriptUtilities';
 
 export interface CacheableCommand
 {
@@ -109,6 +111,11 @@ export class LocalMemoryCacheSingleton
             if (k === 'dotnetInstallToolCacheTtlMs')
             {
                 return undefined;
+            }
+            else if (k === 'env')
+            {
+                return `${minimizeEnvironment(v)}-${TelemetryUtilities.HashData(JSON.stringify(v))}`; // Each command with a unique env must be cached uniquely -- it's helpful to see what the important vars are in the log.
+                // But we don't want to store the entire env/log it.
             }
             return v;
         })}`;
--- a/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
@@ -435,8 +435,8 @@ ${stderr}`));
         else
         {
             const { env, ...optionsWithoutEnv } = options;
-            this.context?.eventStream.post(new CommandExecutionEvent(`Executing command ${fullCommandString}
-with options ${JSON.stringify(options.env !== null && options.env !== undefined ? { env: minimizeEnvironment(env), ...optionsWithoutEnv } : options)}.`));
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            this.context?.eventStream.post(new CommandExecutionEvent(`Executing command ${fullCommandString} with options ${JSON.stringify(options.env !== null && options.env !== undefined ? { env: minimizeEnvironment(env), ...optionsWithoutEnv } : options)}.`));
 
             if (command.runUnderSudo)
             {
--- a/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
@@ -57,7 +57,7 @@ import { ICommandExecutor } from './ICommandExecutor';
 import { IFileUtilities } from './IFileUtilities';
 import { IUtilityContext } from './IUtilityContext';
 import { LockUsedByThisInstanceSingleton } from './LockUsedByThisInstanceSingleton';
-import { executeWithLock, isRunningUnderWSL, loopWithTimeoutOnCond } from './TypescriptUtilities';
+import { executeWithLock, isRunningUnderWSL, loopWithTimeoutOnCond, minimizeEnvironment } from './TypescriptUtilities';
 
 export class CommandExecutor extends ICommandExecutor
 {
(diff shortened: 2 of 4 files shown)


vscode-dotnet-runtime-library/src/LocalMemoryCacheSingleton.ts
```diff
@@ -110,6 +112,11 @@ export class LocalMemoryCacheSingleton
             {
                 return undefined;
             }
+            else if (k === 'env')
+            {
+                return `${minimizeEnvironment(v)}-${TelemetryUtilities.HashData(JSON.stringify(v))}`; // Each command with a unique env must be cached uniquely -- it's helpful to see what the important vars are in the log.
```
Forgind, line 117: Does this hash values separately or together? Meaning would you be able to differentiate between:
{ "a": "b", "c": "d" } and {"c": "d", "a": "b"}
?

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -434,8 +434,9 @@ ${stderr}`));
         }
         else
         {
-            this.context?.eventStream.post(new CommandExecutionEvent(`Executing command ${fullCommandString}
-with options ${JSON.stringify(options)}.`));
+            const { env, ...optionsWithoutEnv } = options;
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            this.context?.eventStream.post(new CommandExecutionEvent(`Executing command ${fullCommandString} with options ${JSON.stringify(options.env !== null && options.env !== undefined ? { env: minimizeEnvironment(env), ...optionsWithoutEnv } : options)}.`));
```
Forgind, line 439: tiny nit:
```suggestion
            this.context?.eventStream.post(new CommandExecutionEvent(`Executing command ${fullCommandString} with options ${JSON.stringify(options.env !== null && options.env !== undefined ? { env: minimizeEnvironment(env), ...optionsWithoutEnv } : optionsWithoutEnv)}.`));
```
? May as well not send an extra undefined if present. Wondering if you should do the same if env is empty

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
```diff
@@ -157,6 +157,71 @@ export async function executeWithLock<A extends any[], R>(eventStream: IEventStr
     return returnResult;
 }
 
+const possiblyUsefulUpperCaseEnvVars = new Set<string>([ // This is a local variable instead of in the function so it doesn't get recreated every time the function is called. I looked at compiled JS and didn't see it get optimized.
+    'COMMONPROGRAMFILES',
+    'COMMONPROGRAMFILES(x86)',
+    'PATH',
```
Forgind, line 163: I can imagine users wanting this to be hidden

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
```diff
@@ -157,6 +157,71 @@ export async function executeWithLock<A extends any[], R>(eventStream: IEventStr
...
+    'ALLUSERSPROFILE',
+    'COMSPEC',
+    'DOTNET_MULTILEVEL_LOOKUP',
+    'ELECTRON_RUN_AS_NODE',
+    'LANG',
+    'HOME',
+    'PATHEXT',
+    'SHELL',
+    'TERM',
+    'PWD',
```
Forgind, line 187: Or this. I generally think that if it isn't specific to us, it might be safer to not grab it

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
```diff
@@ -157,6 +157,71 @@ export async function executeWithLock<A extends any[], R>(eventStream: IEventStr
...
+
+/*
+* @remarks Not intended for telemetry redaction -- PII must be handled correctly elsewhere by a telemetry system.
+* This is intended for logging to the console or a log file where we don't want to write out sensitive information, but we do want information that is useful for debugging weird user situations.
+*/
+export function minimizeEnvironment(pathEnv: NodeJS.ProcessEnv): string
+{
+    let pathEnvString = ``;
+
+    for (const key in pathEnv)
```
Forgind, line 215: Do you have to say pathEnv.keys or something?

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
```diff
@@ -157,6 +157,71 @@ export async function executeWithLock<A extends any[], R>(eventStream: IEventStr
...
+    {
+        if (possiblyUsefulUpperCaseEnvVars.has(key.toUpperCase()))
```
Forgind, line 217: This isn't valid on linux; environment variables are case-sensitive:
image

Forgind (approved):
This seems like a good direction 🙂

vscode-dotnet-runtime-library/src/LocalMemoryCacheSingleton.ts
nagilson: Those would be differentiated. It hashes the entire object. Interesting to note that if the ordering of the keys in the object are different that may make it get cached separately/twice, but I don't think that that would happen.

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
nagilson: I'd actually be interested in when the env is undefined or not, but that's a thoughtful suggestion.

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
nagilson: Do you mean PATH? I agree, though this is not for telemetry, only for the log. This isn't very helpful without the PATH, as we see a ton of bugs related to that, so I don't want to remove it.

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
nagilson: The shell/term has a lot of influential factors on whether our code works or not, so I don't want to remove that, but I see your point.

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
nagilson: Nope! :)

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
nagilson: That's an excellent point. Most of the correct environment variables on Linux are uppercase... in terms of the outcome of this:

- We would collect path and PATH, since path.touppercase is PATH. 
- The logging and cache key would contain both, since it doesn't uppercase it when it records it. That's fine.
I think that is useful, actually. Because we could tell if someone made the mistake of editing path and not PATH.

vscode-dotnet-runtime-library/src/LocalMemoryCacheSingleton.ts
nagilson: Ah, that's a good point ;) That's why I included the minimized environment at the front so we can see the things we need. The included hash is to make sure we don't cache things that are actually different environments because we minimized away all of the differences. The approach you mention would be helpful in differentiating variables that we don't need to collect but might have an unintended impact, however. What do you think? 

Thanks for taking a look.

Issue #2201, opened 2025-03-27
Invalid URL - Cannot Install .NET


Philofallever:
### Describe the bug.

![Image](https://github.com/user-attachments/assets/aa5dc064-f8d6-4cbe-8bc7-d389fdc58979)

### Please attach log file(s) here if you have any.

[DotNetAcquisition-ms-dotnettools.vscode-dotnet-runtime-1743042233796.txt](https://github.com/user-attachments/files/19478259/DotNetAcquisition-ms-dotnettools.vscode-dotnet-runtime-1743042233796.txt)
The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

1

### Expected Behavior

1

### .NET Install Tool Version

2.3.0

### Your operating system? Please provide more detail, such as the version and or distro above.

N/A

a35506322:
https://dotnet.microsoft.com/en-us/download/dotnet/thank-you/sdk-9.0.202-windows-x64-installer

To update to the latest .net 9

stann1:
Also broken for me, on Ubuntu 24.04, latest VS  code 1.98.2. Try to roll back the C# extension to version **2.63.32** and .NET instal tool extension to **2.2.8**. This fixed it for me

mwgustin:
> Also broken for me, on Ubuntu 24.04, latest VS code 1.98.2. Try to roll back the C# extension to version **2.63.32** and .NET instal tool extension to **2.2.8**. This fixed it for me

Same - roll back to these versions fixed it for me. 

This was in both my normal environment and a fresh dotnet devcontainer.

noamalul:
Broke for me as well

nagilson:
Investigating this. If anyone has a log file, that'd be helpful since I haven't been able to reproduce it yet.
To do so: Use the command pallette (ctrl + shift + p) and type open extensions log folder, and find the vscode-dotnet-runtime folder, inside of there will be the long.

We added a web cache to this version which I'm guessing is what is causing this issue.

nagilson:
Workaround for now: Downgrade to 2.2.8

It seems like there is a small chance that this axios code will fail: https://github.com/axios/axios/blob/5a9134e0995b08cc0817548ed44e79c420c620b7/lib/helpers/fromDataURI.js#L33 when calling .get() with the url 'https://builds.dotnet.microsoft.com/dotnet/release-metadata/releases-index.json'. That is a valid URL. I wish that the error had some more information. 

[...]

PR #2205, opened 2025-03-28
Bump Axios Cache to Fix Issue with URL & Use Fetch as Backup

nagilson:
In this issue, it appears we never got to the end of the constructor of the web worker since it never has web request sent. I was never able to repro. Seems to happen for about 5% of users now. Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2201

The version of Axios we upgraded to caused issues in https://github.com/arthurfiorette/axios-cache-interceptor/issues/953.

We weren't using the Cache before, but now that we are, it seems to have caused some issues given 1.6.0 of the cache interceptor is not compatible with the version of axios we were using.

Axios upgrades have had enough silent breaking changes that happen only on occasion or in rare contexts, + with the advent of numerous interceptor libraries that may all fail, I added a backup option for this. Almost every time it is in a minor update, unexpected, and really hard to track down :( Fetch is the native library which I would expect to be a reliable alternative when axios doesn't work. It's not the default since it is not as easily feature rich.


vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -56,100 +56,68 @@ export class WebRequestWorkerSingleton
...
+            uncachedAxiosClient.interceptors.response.use(res =>
             {
                 // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
                 (res as any).startTime = (res.config as any).startTime;
                 // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
                 (res as any).finalTime = process.hrtime.bigint();
                 return res;
-            })
+            },
+                res => // triggered when status code is not 2XX
```
Forgind, line 98: I don't think I understand the significance of what you changed here. These blocks of code look the same?

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -206,16 +170,53 @@ export class WebRequestWorkerSingleton
...
+        {
+            response = await this.getFetchResponse(url, ctx, useFetchDownload);
+            clearTimeout(timeout);
+        }
 
         // Response
         return response;
     }
 
+    private async getFetchResponse(url: string, ctx: IAcquisitionWorkerContext, returnDownloadStream = false): Promise<any>
```
Forgind, line 191: This is work the client is supposed to do, right? I'm a bit confused as to why we're making a temp client rather than just setting this.client to the alt client then making the web request as normal?

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -337,8 +340,8 @@ export class WebRequestWorkerSingleton
 
         if (this.proxySettingConfiguredManually(ctx) || discoveredProxy)
         {
-            this.proxyAgent = new HttpsProxyAgent(ctx.proxyUrl ?? discoveredProxy);
-            return this.proxyAgent;
+            const proxyAgent = new HttpsProxyAgent(ctx.proxyUrl ?? discoveredProxy);
```
Forgind, line 343: I think this means proxyAgent is never cached, right? Can you explain why that's positive?

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -427,34 +431,24 @@ export class WebRequestWorkerSingleton
...
-                }
-                catch (error: any)
-                {
-
-                }
-            }
-            return response?.data;
+            const response = await this.axiosGet(url, ctx, { ...options });
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            return this.client !== null ? response?.data : response;
```
Forgind, line 442: Why do we care if the client is null here? I might've expected more like `return response?.data ?? response;` If we do care about client being null, can we skip the previous line?

Forgind (approved):
I wish the lock-related files wouldn't auto-generate

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
nagilson: It's wrapped in a try block now.
Axios has a client system but fetch does not.
This creates the client if it can but if Axios fails unexpectedly (like in the linked invalid url bug) we dont make a client BC we can't.

Then instead we will use fetch and record the error diagnostics for why the axios client failed. We also do this fetch request as a backup if axios doesnt work but has a client.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
nagilson: I hope my comment above clarifies. This isnt a temp client, it is an entirely different web lib (native fetch) vs axios if axios does not work

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
nagilson: Correct. That is a positive bc the previous case is a bug. This class uses the singleton pattern which allows it to share a web cache amongst all of the code, and such that there is only one global web worker in the entire code. However, some code does not want to run under the proxy. Each request may have its own proxy setting, although unlikely, (so maybe we don't really want to do this perf hit for a bug fix that seems hard to actually manifest) you could configure the API to work as such. So, it is best to recreate this each time.

nagilson: That's a great question, btw :+1:

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
nagilson: Hm, I think the above code is misleading since axiosGet will call fetch instead of axios if client is null. I should rename the function so that it's more clear and possibly make it always output the same data format.

nagilson: I've improved this!

PR #2209, opened 2025-03-31
Skip Parsing Empty Dotnet Executable

nagilson:
Callers may try to find --list-runtimes with an empty string, especially if the existingPath setting is empty, which can return '""' instead of ''. This is a performance improvement.

In addition, it looks like a recent change I did break some of the logic to find a working command. options?.at() is not valid when options is not provided as an array because it is  an object type. [] indexing does the same thing and is safer, as indexing on an object will just return undefined, unless it has a '0' key, but that is never expected here.

### DIFF
--- a/sample/src/extension.ts
+++ b/sample/src/extension.ts
@@ -375,7 +375,7 @@ ${JSON.stringify(result) ?? 'undefined'}`);
         try
         {
             const result: IDotnetListVersionsResult | undefined = await vscode.commands.executeCommand('dotnet.recommendedVersion', { listRuntimes: getRuntimes });
-            vscode.window.showInformationMessage(`Recommended SDK Version to Install: ${result?.at(0)?.version}`);
+            vscode.window.showInformationMessage(`Recommended SDK Version to Install: ${result?.[0]?.version}`);
         }
         catch (error)
         {
--- a/vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
@@ -29,22 +29,28 @@ export class DotnetConditionValidator implements IDotnetConditionValidator
     {
         const hostArch = await this.getHostArchitecture(dotnetExecutablePath, requirement);
 
-        if (requirement.acquireContext.mode === 'sdk') {
+        if (requirement.acquireContext.mode === 'sdk')
+        {
             const availableSDKs = await this.getSDKs(dotnetExecutablePath);
-            if (availableSDKs.some((sdk) => {
+            if (availableSDKs.some((sdk) =>
+            {
                 return this.stringArchitectureMeetsRequirement(hostArch, requirement.acquireContext.architecture) &&
                     this.stringVersionMeetsRequirement(sdk.version, requirement.acquireContext.version, requirement) && this.allowPreview(sdk.version, requirement);
-            })) {
+            }))
+            {
                 return true;
             }
         }
-        else {
+        else
+        {
             // No need to consider SDKs when looking for runtimes as all the runtimes installed with the SDKs will be included in the runtimes list.
             const availableRuntimes = await this.getRuntimes(dotnetExecutablePath);
-            if (availableRuntimes.some((runtime) => {
+            if (availableRuntimes.some((runtime) =>
+            {
                 return runtime.mode === requirement.acquireContext.mode && this.stringArchitectureMeetsRequirement(hostArch, requirement.acquireContext.architecture) &&
                     this.stringVersionMeetsRequirement(runtime.version, requirement.acquireContext.version, requirement) && this.allowPreview(runtime.version, requirement);
-            })) {
+            }))
+            {
                 return true;
             }
         }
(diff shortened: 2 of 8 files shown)


vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
```diff
@@ -29,22 +29,28 @@ export class DotnetConditionValidator implements IDotnetConditionValidator
     {
         const hostArch = await this.getHostArchitecture(dotnetExecutablePath, requirement);
 
-        if (requirement.acquireContext.mode === 'sdk') {
+        if (requirement.acquireContext.mode === 'sdk')
+        {
             const availableSDKs = await this.getSDKs(dotnetExecutablePath);
-            if (availableSDKs.some((sdk) => {
+            if (availableSDKs.some((sdk) =>
+            {
```
edvilme, line 36: Nit: Same line, and maybe replace `(a) => { return statement }` with `(a) => statement`?

vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
```diff
@@ -29,22 +29,28 @@ export class DotnetConditionValidator implements IDotnetConditionValidator
...
             }
         }
-        else {
+        else
+        {
             // No need to consider SDKs when looking for runtimes as all the runtimes installed with the SDKs will be included in the runtimes list.
             const availableRuntimes = await this.getRuntimes(dotnetExecutablePath);
-            if (availableRuntimes.some((runtime) => {
+            if (availableRuntimes.some((runtime) =>
+            {
```
edvilme, line 49: Same as above

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -436,7 +436,7 @@ ${stderr}`));
         {
             const { env, ...optionsWithoutEnv } = options;
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-            this.context?.eventStream.post(new CommandExecutionEvent(`Executing command ${fullCommandString} with options ${JSON.stringify(options.env !== null && options.env !== undefined ? { env: minimizeEnvironment(env), ...optionsWithoutEnv } : options)}.`));
+            this.context?.eventStream.post(new CommandExecutionEvent(`Executing command ${fullCommandString} with options ${JSON.stringify(options?.env !== null && options?.env !== undefined ? { env: minimizeEnvironment(env), ...optionsWithoutEnv } : options)}.`));
```
Forgind, line 439: What does null.env normally return? I'd imagine ts would avoid erroring and make null.env return null anyway, no?

Forgind (approved):
At some point, can you have a single PR that does all the little style tweaks you might need without changing any functionality? I find these sorts of PRs much harder to read with nonfunctional changes mixed in.

sample/src/extension.ts
```diff
@@ -375,7 +375,7 @@ ${JSON.stringify(result) ?? 'undefined'}`);
         try
         {
             const result: IDotnetListVersionsResult | undefined = await vscode.commands.executeCommand('dotnet.recommendedVersion', { listRuntimes: getRuntimes });
-            vscode.window.showInformationMessage(`Recommended SDK Version to Install: ${result?.at(0)?.version}`);
+            vscode.window.showInformationMessage(`Recommended SDK Version to Install: ${result?.[0]?.version}`);
```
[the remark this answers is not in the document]
joeloff: Also, I'm curious to know why the recommended version would by null. Can that actually happen and if so, should that be an error or use a different message?

joeloff, line 378: Also, I'm curious to know why the recommended version would by null. Can that actually happen and if so, should that be an error or use a different message?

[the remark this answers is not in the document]
nagilson: Yep, that's because of the `?`, it's essentially expanding the object out if its not undefined but just doing ?[] is not valid.

vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
```diff
@@ -71,6 +77,11 @@ export class DotnetConditionValidator implements IDotnetConditionValidator
         // dotnet --info is not machine-readable and subject to breaking changes. See https://github.com/dotnet/sdk/issues/33697 and https://github.com/dotnet/runtime/issues/98735/
         // Unfortunately even with a new API, that might not go in until .NET 10 and beyond, so we have to rely on dotnet --info for now.*/
 
+        if (!hostPath || hostPath === '""')
+        {
+            return '';
+        }
+
```
edvilme, line 84:
```suggestion
        if (!hostPath || hostPath === '""') return '';
```

Note: I am not familiar with the specific style guidelines here, perhaps your approach makes more sense here

nagilson:
> At some point, can you have a single PR that does all the little style tweaks you might need without changing any functionality? I find these sorts of PRs much harder to read with nonfunctional changes mixed in.

You're right, thank you for pointing it out. I have been trying to be better about making the PRs smaller but I have a tendency to want to include other things inside of it, which is not the best for many reasons, not just PR review but also commit history and backing out changes 👍

edvilme (approved):
Left a couple of style comments (again, I am not very familiar with MS's TS conventions), but looks good to me :)

sample/src/extension.ts
nagilson: Good question - that is if there are no dotnet packages on the distro feed, and or it could not connect to the internet to find any version to recommend to the user. In the rare case their internet comes back after the UI popup, then they can still type in a version and it will work

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
nagilson: Exactly right, it would fail if you try null.env. ?. makes it become undefined by skipping evaluation of the rest of the statement

Issue #2212, opened 2025-04-01
[NETE2ESDK][Ubuntu24.04]DonetGlobalSDKAcquisitionError in output when installing .NET SDK


NicoleWang001:
### Describe the bug.

On Ubuntu 24.04, Error in output - Failed to download .NET 8.0.114-global~x64 with error DonetGlobalSDKAcquisitionError, but in fact .net SDK is installed successfully.

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

[Logs.zip](https://github.com/user-attachments/files/19547376/Logs.zip)

### Reproduction Steps

1. On Ubuntu 24.04, run the sample command ‘Install .NET SDK Globally via .NET Install Tool (Former Runtime Extension)‘, enter the specific version like 8.0.114.
2. Observe

### Expected Behavior

.NET SDK 8.0.114 should be installed successfully without any errors

### Exceptions & Errors (if any)

Error in output - Failed to download .NET 8.0.114-global~x64 with error DonetGlobalSDKAcquisitionError, 
    1) If re-run the command, it shows repairing packages and SDK is installed
 ![Image](https://github.com/user-attachments/assets/83f7392d-a1cb-474c-abd3-ae12624ec5e3)
    2) If run dotnet --list-sdks in Terminal then re-run the command in VSCode, it shows the SDK has been installed.

![Image](https://github.com/user-attachments/assets/903dc9fe-3739-4843-9810-10aef059592f)

![Image](https://github.com/user-attachments/assets/4a774906-92e7-43ba-9ba8-c41412396e9f)

### .NET Install Tool Version

v2.3.1

### Your operating system? Please provide more detail, such as the version and or distro above.

Linux

### VS Code Version

1.98.2

NicoleWang001:
If specific the version to 8.0.113, the same error occurs, but in fact SDK 8.0.114 is installed.

![Image](https://github.com/user-attachments/assets/8fa1bb0d-540b-45e7-bce4-6d7b266048a0)

nagilson:
This is an issue in 2.3.0 as well, I guess I should update the test plan for 24.04 as it seems to have different behavior from other versions of Ubuntu, since all of the older ones work. There is a quirky behavior with the VS Code Input Box where it auto closes and does not return the placeholder value on Ubuntu 24.04 that I observed.

NicoleWang001:
Fixed on the new build.

[...]

PR #2213, opened 2025-04-01
Minor Fixes for 2.3.1 - Fix File Deletion Check & Performance Improvement

nagilson:
Several fixes (sorry, trying to act quickly) for the next release are included:

- `In` checks whether the type has the object as a key in it (such as {key: value}, `.includes` should be used in the file deletion code. This fixes https://github.com/dotnet/vscode-dotnet-runtime/issues/2212 because it was using the old stdout files to think the apt-get install was completed, even though it was not, which would cause validation to run before apt-get install was finished and then fail, even when the install was successful. This was a regression in 2.3.0. Not a problem in 2.2.8 since it is sync blocked.
- Makes log file include MS for perf investigation team.
- Updates cache duration to be extended per request of C#DK Team and D. Kean
- Makes PATH logging happen async (may not be in order) to prevent startup perf delay
- Ubuntu 24.04 VS Code Input Box will sometimes auto-terminate and close, so adds logic to fill in version in this instance.


vscode-dotnet-runtime-library/src/Utils/FileUtilities.ts
```diff
@@ -81,7 +81,7 @@ export class FileUtilities extends IFileUtilities
             try
             {
                 eventStream?.post(new FileToWipe(`The file ${f} may be deleted.`))
-                if (!fileExtensionsToDelete || path.extname(f).toLocaleLowerCase() in fileExtensionsToDelete)
+                if (!fileExtensionsToDelete || fileExtensionsToDelete?.includes(path.extname(f).toLocaleLowerCase()) && !(f?.includes('lock')))
```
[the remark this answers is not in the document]
nagilson: Nope, this is actually correct.

vscode-dotnet-runtime-library/src/Utils/FileUtilities.ts
```diff
@@ -81,7 +81,7 @@ export class FileUtilities extends IFileUtilities
             try
             {
                 eventStream?.post(new FileToWipe(`The file ${f} may be deleted.`))
```
joeloff, line 83: It's logging the may-be-deleted message before it evaluates the file extension, in which case it may not be delete. The wording just seems a bit ambiguous since it sounds like it already determined that the deletion can proceed.

Would it make more sense to have an `else` statement that executes after the extension check and instead log when a file is not being deleted

vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
```diff
@@ -6,31 +6,38 @@ import * as fs from 'fs';
...
+{
     private log: string[] = [];
 
     constructor(private readonly logFilePath: string) {}
 
-    public post(event: IEvent): void {
-        this.writeLine(`${ new Date().toLocaleString() } ${ event.eventName }`);
+    public post(event: IEvent): void
+    {
+        this.writeLine(`${new Date().toLocaleString()} ${new Date().getMilliseconds()} ${event.eventName}`);
```
joeloff, line 17: Is this only including the MS count? Can you show a sample of what the entry looks like. Just from Wouldn't that just write

```
4/1/2025 0001
...
4/1/2025 0001 (this happened 5 seconds later and MS wrapped).
```

Also, does the output pad the MS count with leading 0s so you have fixed output width or does that not matter?

joeloff (approved):

vscode-dotnet-runtime-library/src/EventStream/LoggingObserver.ts
nagilson:
```
4/1/2025, 1:42:28 PM 327 DotnetCommandSucceeded
uninstall
```
This is what it will look like in the current state. It's not the prettiest, though I did it in a rush since the performance investigation team added this as a last-minute request, since their setup makes it hard to test CDK performance with a test build of ours.

I think I will also add improving this to a list of things to do after this is merged. The output is not padded, though this is currently only read by humans, including during the perf investigation CDK is undergoing.

PR #2215, opened 2025-04-02
Fix Proxy Lookup

nagilson:
The old lookup was as follows:
`ctx?.proxyUrl ?? configuredProxy`.

But `??` does not evaluate truthfulness. It evaluates nullness. '' is false but it is not null. The default empty value option for getting the string setting is '' or '""'. Setting this to the proxy agent will make it so it never finds the user setting and it will also make it fail: https://github.com/search?q=repo%3ATooTallNate%2Fproxy-agents%20invalid%20url&type=code

And it wasn't in the try block, as this is a best effort approach it should be.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
@@ -335,7 +335,7 @@ export class WebRequestWorkerSingleton
 
             if (this.proxySettingConfiguredManually(ctx) || discoveredProxy)
             {
-                const finalProxy = ctx?.proxyUrl && ctx?.proxyUrl !== '""' ? ctx.proxyUrl : discoveredProxy;
+                const finalProxy = ctx?.proxyUrl && ctx?.proxyUrl !== '""' && ctx?.proxyUrl !== '' ? ctx.proxyUrl : discoveredProxy;
                 ctx.eventStream.post(new ProxyUsed(`Utilizing the Proxy : Manual ? ${ctx?.proxyUrl}, Automatic: ${discoveredProxy}, Decision : ${finalProxy}`))
                 const proxyAgent = new HttpsProxyAgent(finalProxy);
                 return proxyAgent;
--- a/vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
@@ -37,6 +37,7 @@ import
     EventBasedError,
     EventCancellationError,
     OfflineDetectionLogicTriggered,
+    ProxyUsed,
     SuppressedAcquisitionError,
     WebRequestCachedTime,
     WebRequestError,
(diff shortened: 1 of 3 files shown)


vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -331,16 +332,18 @@ export class WebRequestWorkerSingleton
                     discoveredProxy = autoDetectProxies.http.toString();
                 }
             }
-            catch (error: any)
+
+            if (this.proxySettingConfiguredManually(ctx) || discoveredProxy)
             {
-                ctx.eventStream.post(new SuppressedAcquisitionError(error, `The proxy lookup failed, most likely due to limited registry access. Skipping automatic proxy lookup.`));
+                const finalProxy = ctx?.proxyUrl && ctx?.proxyUrl !== '""' ? ctx.proxyUrl : discoveredProxy;
```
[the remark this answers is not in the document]
nagilson: This is not necessary, but it may be good to be explicit. So Ok, sure.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -331,16 +332,18 @@ export class WebRequestWorkerSingleton
                     discoveredProxy = autoDetectProxies.http.toString();
                 }
             }
-            catch (error: any)
+
+            if (this.proxySettingConfiguredManually(ctx) || discoveredProxy)
```
Forgind, line 336: If you want to skip this check, you could put an "else return null" block inside the above if/else if block

Or I think you can simplify the whole thing if you just initialize discoveredProxy to what ctx?.proxyUrl then check if it's what's needed and configure it automatically if not. Then you can skip line 338.

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -331,16 +332,18 @@ export class WebRequestWorkerSingleton
...
             {
-                ctx.eventStream.post(new SuppressedAcquisitionError(error, `The proxy lookup failed, most likely due to limited registry access. Skipping automatic proxy lookup.`));
+                const finalProxy = ctx?.proxyUrl && ctx?.proxyUrl !== '""' && ctx?.proxyUrl !== '' ? ctx.proxyUrl : discoveredProxy;
```
Forgind, line 338: nit:
I think ctx?.proxyUrl !== '' is covered by ctx?.proxyUrl
?

Forgind (approved):
Looks good 🙂

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
nagilson: It is. CoPilot suggested that to make the code more clear 🤷

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -479,7 +482,7 @@ If you're on a proxy and disable registry access, you must set the proxy in our
 
     private proxySettingConfiguredManually(ctx: IAcquisitionWorkerContext): boolean
     {
-        return ctx?.proxyUrl ? true : false;
+        return ctx?.proxyUrl ? ctx?.proxyUrl !== '""' : false;
```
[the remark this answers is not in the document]
nagilson: Thankfully no, it could be '' but that evaluates to false

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
nagilson: That's a great point, @/joeloff had some suggestions too for the last PR, so ill combine those improvements in there.

PR #2217, opened 2025-04-03
Implement Node Mutex

nagilson:
# Summary

Adds the class `NodeIPCMutex`. It is designed in such a way that it could become a portable library.
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2221

This is a class that provides a mutex (mutual exclusion) lock using IPC (Inter-Process Communication) on Node.js.
 * It will work across multiple processes (aka multiple vs code windows) and async code in the same process.
 * Thus, node.js uses a named pipe on Windows and a file descriptor on Linux and OS X to create the lock, so that the lock can be shared across processes, and is resilient to pkill.

This approach was suggested to me after discussing with @joaomoreno, who owns windows-mutex and much of vscode. He was very helpful to talk to.

# Why are you writing all of this when it seems like this should already exist somewhere out there in node.js land?
 ### Hint: It Doesn't ( sadness, or I couldn't find it )

 * Many locking mechanisms such as lockfile and proper-lockfile run into issues when processes are killed or die unexpectedly, because they rely on the file system.
 * windows-mutex is another library, which is a wrapper around a C mutex, however it is Windows only.
 * async-mutex is another node.js library, but it stores a list of owners in memory, which may cause issues if multiple processes try to acquire the same lock.
 * Named pipes on windows die when processes die. There is no equivalent check on Linux or OS X, but IPC allows us to check whether the process is still alive or not so we don't deadlock

# How can I be reassured this wont be potentially catastrophic / have unexpected corner cases

 * Much of this code is inspired by VSCode's IPC code, which is a well tested approach. However, their approach focuses on permitting only one main VS Code Process to run at a time.
 * https://github.com/microsoft/vscode/blob/main/extensions/git/src/ipc/ipcServer.ts#L15
 * https://github.com/microsoft/vscode/blob/main/src/vs/code/electron-main/main.ts#L318

As much as I would love to have a backup option of the old code, we cannot have some code trying a backup option if it does not work, because then everything will be broken if 2 different locking mechanisms are utilized. It could be an environment flag, but I don't trust people will remember for all of the instances of vscode to set it correctly.

# Changes:

-> Rewrote the code to simply call around this wrapper without having to change any other logic.

-> The VS Code, (code) is well tested as it is run by all of the vscode instances. There is obviously an inherent risk that this may cause unexpected issues, but with some thorough testing and backup from the existence that this previous logic is good, I hope that we can trust this code.

# Examples:

In this image, I added a breakpoint to force vscode instance A, process A (right) to acquire the lock, to make sure that Instance A, process B, and Instance B is able to connect to process A, and wait. I verified that they ran in order and B waited 18 times for A to be released.
image

In this image, I have 2 vscode instances on Windows. 
image The unshown process acquired both locks, run under admin. The left one was run as user local. I killed the admin one while the local one had connected to it and made sure it worked and detected the pkill.

I demonstrated the same on Unix here:
image

And my tests also validate various scenarios. I will also have the vendors run these tests with insiders/regular vscode.

# Caveats

The nodejs library / OS does have issues with certain character limits for paths - if the user has a custom /temp/ folder on Mac, especially, that is too long, they may run into issues. However, vscode runs the same logic and has worked ok. Users can set the XGD runtime dir and it is usually assigned correctly by the OS. I add the environment-variable VSCODE_DOTNET_RUNTIME_DISABLE_MUTEX=1 to disable locking which is a concerning thing to set, but would enable us to help solve things for users in this odd case while suggesting they don't use multiple vscode instances in parallel.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
@@ -160,13 +160,7 @@ export class InstallTrackerSingleton
 
     private getLockFilePathForKey(provider: IInstallationDirectoryProvider, dataKey: string): string
     {
-        const providerPath = provider.getInstallDir('foo');
-        const hasNumber = /\d/;
-
-        const dir = hasNumber.test(providerPath) ? path.dirname(providerPath) : providerPath; // bad codependency I dont want to untangle atm : sdk local itnstall providers dont create a separate folder, runtime ones do
-        // but installing state should be at the root folder (not trustable from __dirname and can't access vscode storage folder) ... so if the folder is like 8.0 or 8.0~aspnetcore~x64 e go up to root
-
-        return path.join(dir, `${dataKey}.lock`); // install object required but we call the dir of it
+        return `${dataKey}.lock`;
     }
 
     /**
--- a/vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
@@ -160,7 +160,7 @@ export class InstallTrackerSingleton
 
     private getLockFilePathForKey(provider: IInstallationDirectoryProvider, dataKey: string): string
     {
-        return `${dataKey}.lock`;
+        return `${dataKey}Lk`;
     }
 
     /**
--- a/vscode-dotnet-runtime-library/src/Acquisition/StringConstants.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/StringConstants.ts
@@ -14,12 +14,12 @@ import { IInstallationDirectoryProvider } from './IInstallationDirectoryProvider
 */
 export function GLOBAL_INSTALL_STATE_MODIFIER_LOCK(directoryProvider: IInstallationDirectoryProvider, install: DotnetInstall): string
 {
-    return 'modifyingGlobalState.lock';
+    return 'vscdGlLk';
 }
 
 export function RUN_UNDER_SUDO_LOCK(sudoDirectory: string): string
 {
-    return 'runUnderSudo.lock';
+    return 'vscdSudoLk';
 }
 
 export const UNABLE_TO_ACQUIRE_GLOBAL_LOCK_ERR = '898998';
--- a/vscode-dotnet-runtime-library/src/Acquisition/StringConstants.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/StringConstants.ts
@@ -3,7 +3,6 @@
 *  The .NET Foundation licenses this file to you under the MIT license.
 *--------------------------------------------------------------------------------------------*/
 
-import * as path from 'path';
 import { DotnetInstall } from './DotnetInstall';
 import { IInstallationDirectoryProvider } from './IInstallationDirectoryProvider';
 
@@ -15,12 +14,12 @@ import { IInstallationDirectoryProvider } from './IInstallationDirectoryProvider
 */
 export function GLOBAL_INSTALL_STATE_MODIFIER_LOCK(directoryProvider: IInstallationDirectoryProvider, install: DotnetInstall): string
 {
-    return path.join(path.resolve(directoryProvider.getInstallDir(install.installId)), 'modifyingGlobalState.lock');
+    return 'modifyingGlobalState.lock';
 }
 
 export function RUN_UNDER_SUDO_LOCK(sudoDirectory: string): string
 {
-    return path.join(sudoDirectory, 'runUnderSudo.lock');
+    return 'runUnderSudo.lock';
 }
 
 export const UNABLE_TO_ACQUIRE_GLOBAL_LOCK_ERR = '898998';
--- a/vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
@@ -40,6 +40,8 @@ export class NodeIPCMutex
     /**
      *
      * @param lockId - The ID of the lock. This should be a unique identifier for the lock being created.
+     * The lockId must be less than about 90 characters as it will get truncated down to 107 characters -- if it is longer than the temp folder may allow on OS X, it may fail.
+     * // https://nodejs.org/api/net.html#:~:text=other%20operating%20systems.-,Identifying%20paths%20for%20IPC%20connections,-%23
      * @param logger - A custom logger for your own logging events. By default, this will log to the console.
      */
     constructor(lockId: string, readonly logger: INodeIPCMutexLogger = new INodeIPCMutexLogger())
@@ -49,27 +51,27 @@ export class NodeIPCMutex
 
     private getIPCHandlePath(id: string): string
     {
-        // The windows file system default length may cause us to fail to create the handle if we don't truncate it.
-        if (id.length > (256 - `\\\\.\\pipe\\vscode-dotnet-install-tool-`.length))
+        const lengthLimit = os.platform() === 'win32' ? 256 : 107;
+        if (id.length > (lengthLimit - `\\\\.\\pipe\\vscd-`.length))
         {
-            id = id.substring(0, 255);
+            id = id.substring(0, lengthLimit - 1);
         }
 
         if (process.platform === 'win32')
         {
             // Special File System to get a Named Pipe on windows (File Descriptors won't work) : https://nodejs.org/docs/latest/api/net.html#ipc-support:~:text=On%20Windows%2C%20the,owning%20process%20exits.
-            return `\\\\.\\pipe\\vscode-dotnet-install-tool-${id}-sock`;
+            return `\\\\.\\pipe\\vscd-${id}-sock`;
         }
 
         if (process.platform !== 'darwin' && process.env.XDG_RUNTIME_DIR)
         {
             // The user or system told us to use this as our applications temporary directory, so this this instead of /temp/
-            return path.join(process.env.XDG_RUNTIME_DIR as string, `vscode-dotnet-install-tool-${id}.sock`);
+            return path.join(process.env.XDG_RUNTIME_DIR as string, `vscd-${id}.sock`);
         }
 
         // A file descriptor in /temp/ is a good option to hold this sock.
         // Access to /tmp/ may be restricted, but all processes must use the same directory, we can't really condition on this.
-        return path.join(os.tmpdir(), `vscode-dotnet-install-tool-${id}.sock`);
+        return path.join(os.tmpdir(), `vscd-${id}.sock`);
     }
 
     /**
--- a/vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
@@ -35,7 +35,7 @@ export class INodeIPCMutexLogger
  */
 export class NodeIPCMutex
 {
-    private lockPath: string;
+    private readonly lockPath: string;
     private server?: Server;
 
     /**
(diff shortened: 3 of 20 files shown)


joaomoreno:
Cool. Forgot to inform you this also uses the same trick: https://github.com/joaomoreno/deemon

We use deemon for our local dev setup in VS Code as well.

nagilson:
> Cool! Forgot to inform you this also uses the same trick: https://github.com/joaomoreno/deemon
> 
> We use deemon for our local dev setup in VS Code as well.

Oh, that looks quite useful :) I think there are some first party language extensions that would benefit from utilizing that library, including us, considering we spawn a lot of processes. Thank you!

vscode-dotnet-runtime-library/src/Acquisition/StringConstants.ts
```diff
@@ -15,12 +14,12 @@ import { IInstallationDirectoryProvider } from './IInstallationDirectoryProvider
 */
 export function GLOBAL_INSTALL_STATE_MODIFIER_LOCK(directoryProvider: IInstallationDirectoryProvider, install: DotnetInstall): string
 {
-    return path.join(path.resolve(directoryProvider.getInstallDir(install.installId)), 'modifyingGlobalState.lock');
+    return 'vscdGlLk';
```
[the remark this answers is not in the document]
nagilson: That is horrible and would ruin the whole point of doing any of this.

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -160,13 +160,7 @@ export class InstallTrackerSingleton
...
     private getLockFilePathForKey(provider: IInstallationDirectoryProvider, dataKey: string): string
     {
-        const providerPath = provider.getInstallDir('foo');
-        const hasNumber = /\d/;
-
-        const dir = hasNumber.test(providerPath) ? path.dirname(providerPath) : providerPath; // bad codependency I dont want to untangle atm : sdk local itnstall providers dont create a separate folder, runtime ones do
-        // but installing state should be at the root folder (not trustable from __dirname and can't access vscode storage folder) ... so if the folder is like 8.0 or 8.0~aspnetcore~x64 e go up to root
-
-        return path.join(dir, `${dataKey}.lock`); // install object required but we call the dir of it
+        return `${dataKey}.lock`;
```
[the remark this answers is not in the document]
nagilson: This is a terrible idea.

nagilson:
@webreidi @lifenglu @dibarbet FYI on this reimplementation of mutexes in the .net runtime extension. If there is a bug here, it would impact acquire and all of the related acquisition code to not function at all. The latest logic after migrating our extension from sync -> async blocking methods introduced many race conditions. 

I initially used proper-lockfile as a mutex but this did not work and caused issues across several instances of code running at the same time so I had to implement a mutex myself. I talked to the vscode team who helped with this and did lots of validation but figured it was worth creating awareness. It might be worth having another eye on it.

I think the main issue this could hit is if /temp/ is restricted but most people should set XDG as mentioned above in the caveats section.

vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
```diff
@@ -0,0 +1,286 @@
...
+            return `${ipcPathDir}vscd-${id}-sock`;
+        }
+        else
+        {
+            return path.join(ipcPathDir, `vscd-${id}.sock`);
+        }
+    }
+
+    /**
+     * @remarks This function will try to hold a lock to prevent both other processes and other async code in this processes from running simultaneously..
```
MiYanni, line 83: nit:
```suggestion
     * @remarks This function will try to hold a lock to prevent both other processes and other async code in this processes from running simultaneously.
```

vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
```diff
@@ -0,0 +1,286 @@
...
+     * @param fn - The function to run while we have the lock. This should be a promise, and will be awaited with its value returned.
+     * Use the () => {} syntax to ensure your scope is correctly bound to the function so it can access the variables you declare.
+     * @param retryDelayMs The number of milliseconds to wait before retrying to acquire the lock if it is already held by another process.
+     * @param timeoutTimeMs The total amount of time to try to acquire the lock before giving up. This is the maximum time to wait for the lock to be released.
+     * @param actionId The action ID to use for logging and debugging purposes. This should be a unique identifier for the action being performed.
+     * @returns The awaited value returned by the function passed in as fn.
+     */
+    public async acquire<T>(fn: () => Promise<T>, retryDelayMs = 100, timeoutTimeMs = 1000, actionId: string): Promise<T>
+    {
+        const maxRetryCountToEndAtRoughlyTimeoutTime = Math.ceil(timeoutTimeMs / retryDelayMs) + 1;
```
MiYanni, line 97: `maxRetryCountToEndAtRoughlyTimeoutTime` 😅

vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
```diff
@@ -0,0 +1,286 @@
...
+                    if (retries >= maxRetryCountToEndAtRoughlyTimeoutTime)
+                    {
+                        throw new Error(`Action: ${actionId} Failed to acquire lock after ${retries} retries out of ${maxRetryCountToEndAtRoughlyTimeoutTime} total available attempts.
+The lock: ${this.lockPath} may be held by another process or instance of vscode. Try restarting your machine, deleting the lock, and or increasing the timeout time in the extension settings.
+
+Increase your OS path length limit to at least 256 characters.
+On Linux, you can set XDG_RUNTIME_DIR to be a writeable directory by your user.
+
+If you still face issues, set VSCODE_DOTNET_RUNTIME_DISABLE_MUTEX=true in the environment.
+Report this issue to our vscode-dotnet-runtime GitHub for help.`);
```
MiYanni, line 120: I think I've asked this before, but we don't do localization for this extension, right?

vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    private async tryAcquire<T>(actionId: string, fn: () => Promise<T>): Promise<T>
+    {
+        return new Promise<T>((resolve, reject) =>
+        {
+            this.server = createServer();
+
+            this.server.on('error', reject);
+            // The listeningListener interface is designed to return void, but we need to return the result of running f while holding the handle.
```
MiYanni, line 156: nit:
```suggestion
            // The listeningListener interface is designed to return void, but we need to return the result of running fn while holding the handle.
```

vscode-dotnet-runtime-library/src/Utils/TypescriptUtilities.ts
```diff
@@ -65,95 +59,40 @@ export async function isRunningUnderWSL(acquisitionContext: IAcquisitionWorkerCo
...
-    {
-        throw codeFailureAndNotLockFailure;
-    }
+        const logger = new NodeIPCMutexLoggerWrapper(eventStream);
+        const mutex = new NodeIPCMutex(lockId, logger);
 
-    return returnResult;
+        const result = await mutex.acquire(async () =>
+        {
+            // await must be used to make the linter allow fn to be async, which it must be.
```
MiYanni, line 90: nit:
```suggestion
            // await must be used to make the linter allow f to be async, which it must be.
```

vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
```diff
@@ -0,0 +1,286 @@
...
+     * @param logger - A custom logger for your own logging events. By default, this will log to the console.
+     */
+    constructor(lockId: string, readonly logger: INodeIPCMutexLogger = new INodeIPCMutexLogger())
+    {
+        this.lockPath = this.getIPCHandlePath(lockId);
+    }
+
+    private getIPCHandlePath(id: string): string
+    {
+        const lengthLimit = os.platform() === 'win32' ? 256 : 107;
```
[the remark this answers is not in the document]
Forgind: https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars

I think this is was nagilson was going for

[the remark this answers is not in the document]
nagilson: Exactly correct 😉 Good question. This is also documented in the code here: https://github.com/dotnet/vscode-dotnet-runtime/pull/2217/files/4a29eb8e30ecb0c9547ae4775b07cac880775a7b#diff-7690e47eb74728aa1dbdda008966d4f60d844282ecd6cfa4070d3bb4c7bedb22R44 

via the node.js docs: https://nodejs.org/api/net.html#:~:text=other%20operating%20systems.-,Identifying%20paths%20for%20IPC%20connections,-%23

vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
nagilson: That's correct, I need to add that eventually, though not in this PR. This error message should probably be customizable if it were to be an external library I ported out as well for other tools to have their own error.

This error message is quite long, but the problem is complex and there are many possible solutions/issues. I would be happy to have any suggestions to make it shorter/better.

vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
nagilson: My variable name here is definitely amusing 😄 
It doesn't exactly end at the timeout time since it's not a time-based timeout, and instead a number of retries based timeout. I could improve this, though I would rather not for the scope of this PR.

vscode-dotnet-runtime-library/src/Acquisition/StringConstants.ts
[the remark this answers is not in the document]
nagilson: I should be nicer to copilot, sorry copilot.

vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
nagilson: Note: this could theoretically be a bigger issue, but all of the callers to this are already wrapped in their own custom timeouts with timeout logic.

vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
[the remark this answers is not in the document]
nagilson: This post was quite interesting to read.
I didn't know [Mac OS X 10.9](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man4/unix.4.html) limited the length to 104 characters. 😵 That's a very old version, though FreeBSD also limits it to that length. I don't want to break this for FreeBSD people, so maybe I should change this to 104...

nagilson, line 54:
```suggestion
        const lengthLimit = os.platform() === 'win32' ? 256 : 104; // Mac 10.9 and FreeBSD have their own length limit.
```

nagilson:
With respect to the new test failures, it looks like they updated the CI machines to a new version of Ubuntu that has some changes for dotnet support, so I will have to fix that in another PR, but not going to block this one on it.

nagilson:
> I definitely don't have a technical understanding of the changes. But I didn't see anything that seemed glaringly problematic. 👍 I'm guessing this took a lot of testing and investigation. Great job! ⭐

Thank you, I'm consulting with some others to see if we can get a review from someone who is more familiar with this logic as well.

vscode-dotnet-runtime-library/src/Utils/NodeIPCMutex.ts
```diff
@@ -0,0 +1,282 @@
...
+        this.supplementalErrorInfo = lockFailureErrorMessage ?? '';
+        this.lockPath = this.getIPCHandlePath(lockId);
+    }
+
+    private getIPCHandlePath(id: string): string
+    {
+        const lengthLimit = os.platform() === 'win32' ? 256 : 104; // Mac 10.9 and FreeBSD have their own length limit.
+
+        // On Unix, A file descriptor in /temp/ is a good option to hold this sock.
+        // In Linux, The user or system may set XDG_RUNTIME_DIR to set our applications temporary directory, so this this instead of /temp/
```
AArnott, line 59: "this this"

AArnott (approved):
I'm impressed that all this can be done with mere javascript. I certainly am not intimately familiar with handling errors from these APIs with javascript, and the nuances of per-OS behavior can be relevant, but it appears you account for everything I know of.

PR #2226, opened 2025-04-08
Fix Linux Tests By Centralizing Distro Version Info

nagilson:
The tests are conditioned to check the support status of dotnet, etc. Last night the CI runners got updated to Ubuntu 24.04. 24.04 does not have .NET 9 support built-in, which means the versions of dotnet either went up or down.

The old logic was somewhat spread out.
This is still not ideal but this at least reduces the points of contact that need updated. Maybe this is a bit too extensive of code cleanup for fixing a breaking pipeline, though I'm the only one who will be blocked by this for now, so I think it's ok.

Some of the strings that got updated were just placeholder data.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/StringConstants.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/StringConstants.ts
@@ -24,3 +24,7 @@ export function RUN_UNDER_SUDO_LOCK(sudoDirectory: string): string
 }
 
 export const UNABLE_TO_ACQUIRE_GLOBAL_LOCK_ERR = '898998';
+
+export const UBUNTU_DISTRO_INFO_KEY = 'Ubuntu';
+export const RED_HAT_DISTRO_INFO_KEY = 'Red Hat Enterprise Linux';
+export const DEBIAN_DISTRO_INFO_KEY = 'Debian GNU/Linux';
--- a/vscode-dotnet-runtime-library/src/test/unit/DebianDistroTests.test.ts
+++ b/vscode-dotnet-runtime-library/src/test/unit/DebianDistroTests.test.ts
@@ -5,10 +5,10 @@
  * ------------------------------------------------------------------------------------------ */
 import * as chai from 'chai';
 import * as os from 'os';
+import { MockCommandExecutor } from '../mocks/MockObjects';
 import { DebianDistroSDKProvider } from '../../Acquisition/DebianDistroSDKProvider';
 import { DotnetInstallMode } from '../../Acquisition/DotnetInstallMode';
 import { DistroVersionPair, DotnetDistroSupportStatus, LinuxVersionResolver } from '../../Acquisition/LinuxVersionResolver';
-import { MockCommandExecutor } from '../mocks/MockObjects';
 import { getDistroInfo, getMockAcquisitionContext, getMockUtilityContext } from './TestUtility';
 import { DEBIAN_DISTRO_INFO_KEY } from '../../Acquisition/StringConstants';
 const assert = chai.assert;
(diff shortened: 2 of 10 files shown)


vscode-dotnet-runtime-library/src/test/unit/TestUtility.ts
```diff
@@ -58,29 +59,78 @@ export function getMockAcquisitionWorkerContext(acquireContext : IDotnetAcquireC
...
+
+    if (distroInfo.distro === UBUNTU_DISTRO_INFO_KEY)
+    {
+        if (distroInfo.version < '22.04')
+        {
+            return '6.0.100';
+        }
+        if (distroInfo.version < '24.04')
+        {
+            return '7.0.100';
```
nagilson, line 112:
```suggestion
            return '9.0.100';

```

vscode-dotnet-runtime-library/src/Acquisition/StringConstants.ts
```diff
@@ -24,3 +24,7 @@ export function RUN_UNDER_SUDO_LOCK(sudoDirectory: string): string
 }
 
 export const UNABLE_TO_ACQUIRE_GLOBAL_LOCK_ERR = '898998';
+
+export const UBUNTU_DISTRO_INFO_KEY = 'Ubuntu';
```
Forgind, line 28: 👍

vscode-dotnet-runtime-library/src/test/unit/DebianDistroTests.test.ts
```diff
@@ -5,30 +5,36 @@
...
+    const distroInfo = await getDistroInfo(acquisitionContext);
+    return os.platform() === 'linux' && distroInfo.distro === DEBIAN_DISTRO_INFO_KEY;
+}
+
 suite('Debian Distro Logic Unit Tests', () =>
 {
     test('Recommends Correct Version', async () =>
     {
-        if (shouldRun)
+        if (await shouldRun())
```
Forgind, line 37: So disabling a bunch of tests on non-Debian? Edit: these are the DebianDistroTests, so that actually makes sense 😆 Could any of them be moved to the linux tests?

'await' in a test feels weird to me. I can't point to a real problem with it; it just makes me concerned about hangs

vscode-dotnet-runtime-library/src/test/unit/TestUtility.ts
```diff
@@ -58,29 +59,79 @@ export function getMockAcquisitionWorkerContext(acquireContext : IDotnetAcquireC
...
+    if (distroInfo.distro === UBUNTU_DISTRO_INFO_KEY)
+    {
+        if (distroInfo.version < '22.04')
+        {
+            return '6.0.100';
+        }
+        if (distroInfo.version < '24.04')
+        {
+            return '9.0.100';
+
```
Forgind, line 113: nit: extra newline

Forgind (approved):
🙂

vscode-dotnet-runtime-library/src/test/unit/DebianDistroTests.test.ts
nagilson: Yeah, each distro has its own tests that can be run locally to make sure it works properly. The linux tests also test the same logic, just not the distro specific logic. There is certainly room for improvement. Mocha as a test framework times out every test with our current options, so it should be ok, but I think that's a good point.

vscode-dotnet-runtime-library/src/test/unit/TestUtility.ts
nagilson: Ah, good catch, this got added in the above commit.

vscode-dotnet-runtime-library/src/test/unit/DebianDistroTests.test.ts
Forgind: Alternatively, could we check shouldRun() once then just access it? You aren't going to start a test run on Debian and finish on Ubuntu

Issue #2224, opened 2025-04-08
[NETE2ESDK][Ubuntu24.04] Error occurs when re-install runtime or SDK after re-launch VSCode which killed before


NicoleWang001:
### Describe the bug.

Error occurs when re-install runtime or SDK after re-launch VSCode which killed before

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

1. On Ubuntu 24.04, installed 8.0.114 with commands 
    ```
    sudo apt-get update && \
      sudo apt-get install -y dotnet-sdk-8.0
    ```
2. Launch VSCode and run  a command ‘.NET Install Tool: Install .NET SDK System-Wide' to install the .NET SDK and enter the newest version 8.0.114
    While this command is running, run the command ‘Sample: Acquire ASPNET Runtime’
    While this command is running, run the command ‘Sample” concurrently acquire 2.x runtimes’
3. Kill the process code
4. Re-launch VSCode and re-run the commands on step2
5. Observe

### Expected Behavior

All SDK and runtimes should be installed normally

### Exceptions & Errors (if any)

Error occurs when re-install runtime or SDK

![Image](https://github.com/user-attachments/assets/8f347941-4f0a-4398-8b08-24a5d3ffc24a)

### .NET Install Tool Version

v2.3.2

### Your operating system? Please provide more detail, such as the version and or distro above.

Linux

### VS Code Version

1.99.0

PR #2229, opened 2025-04-09
Fix Linux Mutex Utilization

nagilson:
Now that the mutex is implemented correctly, I can check that the code which uses the mutex is all correct.
An issue I discovered, is that on Linux, we use a sudo process folder to exchange files with a master sudo process. 
This folder is shared across all instances of vscode. If you kill vscode, the sudo process does not go away.

When a new instance checks if it should spawn a master sudo process, it tries to communicate with the old one. But the old one may be busy, and the old logic only waited 1 second. The sudo process may be busy for multiple minutes. This can result in race conditions where the old sudo process and new one are doing some of the same tasks. Now, I've given each vscode instance its own folder to run sudo commands under.

The script that runs these commands is still in the same protected location. It will only run authorized and approved commands by us, so spawning a new directory is OK.

Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2224

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
@@ -133,7 +133,7 @@ export class DotnetPathFinder implements IDotnetPathFinder
 
         else
         {
-            this.executor?.execute(CommandExecutor.makeCommand('env', []), { dotnetInstallToolCacheTtlMs: DOTNET_INFORMATION_CACHE_DURATION_MS }, false)
+            this.executor?.execute(CommandExecutor.makeCommand('echo', ['PATH']), { dotnetInstallToolCacheTtlMs: DOTNET_INFORMATION_CACHE_DURATION_MS }, false)
                 .then((result) =>
                 {
                     // Log the default shell state
@@ -143,7 +143,7 @@ export class DotnetPathFinder implements IDotnetPathFinder
 
             if (!(options.shell === '/bin/bash'))
             {
-                this.executor?.execute(CommandExecutor.makeCommand('env', []), { shell: 'bin/bash', dotnetInstallToolCacheTtlMs: SYS_CMD_SEARCH_CACHE_DURATION_MS }, false)
+                this.executor?.execute(CommandExecutor.makeCommand('echo', ['PATH']), { shell: 'bin/bash', dotnetInstallToolCacheTtlMs: SYS_CMD_SEARCH_CACHE_DURATION_MS }, false)
                     .then((result) =>
                     {
                         this.workerContext.eventStream.post(new DotnetFindPathLookupPATH(`Execution Path (Unix Bash): ${result?.stdout}`));
--- a/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
@@ -99,10 +99,12 @@ export class CommandExecutor extends ICommandExecutor
         catch (error: any)
         {
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-            error.message = error.message + `\nFailed to create ${this.sudoProcessCommunicationDir}. Please check your permissions or install dotnet manually.`;
+            error.message = error?.message + `\nFailed to create ${this.sudoProcessCommunicationDir}. Please check your permissions or install dotnet manually.`;
             this.context?.eventStream.post(new SudoDirCreationFailed(`The command ${fullCommandString} failed, as no directory could be made: ${JSON.stringify(error)}`));
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
             if (error?.code !== 'EEXIST')
             {
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
                 if (error?.code === 'EPERM' || error?.code === 'EACCES')
                 {
                     this.sudoProcessCommunicationDir = path.dirname(this.sudoProcessScript);
--- a/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
@@ -145,7 +145,7 @@ Please install the .NET SDK manually by following https://learn.microsoft.com/en
         }
         else
         {
-            if (await this.sudoProcIsLive(false, fullCommandString, 2000)) // If the sudo process was spawned by another instance of code, we do not want to have 2 at once but also do not waste a lot of time checking
+            if (await this.sudoProcIsLive(false, fullCommandString, 500)) // If the sudo process was spawned by another instance of code, we do not want to have 2 at once but also do not waste a lot of time checking
             // As it should not be in the middle of an operation which may cause it to take a while, unless it was pkilled.
             {
                 return '0';
--- a/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
@@ -281,11 +281,12 @@ ${stderr}`));
 
 
         const waitTimeMs = this.context?.timeoutSeconds ? (Math.max(this.context?.timeoutSeconds * 1000, 1000)) : 600000;
-        await loopWithTimeoutOnCond(100, waitTimeMs,
+        const sampleRateMs = 100;
+        await loopWithTimeoutOnCond(sampleRateMs, waitTimeMs,
             function ProcessFinishedExecutingAndWroteOutput(): boolean { return fs.existsSync(outputFile) },
             function doNothing(): void { ; },
             this.context.eventStream,
-            new SudoProcCommandExchangePing(`Ping : Waiting. ${new Date().toISOString()}`)
+            new SudoProcCommandExchangePing(`Ping : Waiting, at rate ${sampleRateMs} with timeout ${waitTimeMs} ${new Date().toISOString()}`)
         )
             .catch(error =>
             {
--- a/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
+++ b/vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
@@ -31,6 +31,7 @@ import
     DotnetWSLSecurityError,
     EventBasedError,
     EventCancellationError,
+    FailedToRunSudoCommand,
     SudoProcAliveCheckBegin,
     SudoProcAliveCheckEnd,
     SudoProcCommandExchangeBegin,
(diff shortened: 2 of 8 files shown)


vscode-dotnet-runtime-library/src/Utils/FileUtilities.ts
```diff
@@ -96,8 +96,16 @@ export class FileUtilities extends IFileUtilities
...
+        try
+        {
+            const output = await fs.promises.readFile(filePath, 'utf8');
+            return output;
+        }
+        catch (error: any)
+        {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            return `File ${filePath} does not exist or is not readable : ${error?.message}`;
+        }
```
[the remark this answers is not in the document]
nagilson: That's a good point, it was intentional because this will get returned up and spit out as an error soon after, but also, this is not ideal and hard to follow, so I like this change.

vscode-dotnet-runtime-library/src/Acquisition/DotnetPathFinder.ts
```diff
@@ -143,7 +143,7 @@ export class DotnetPathFinder implements IDotnetPathFinder
 
             if (!(options.shell === '/bin/bash'))
             {
-                this.executor?.execute(CommandExecutor.makeCommand('env', []), { shell: 'bin/bash', dotnetInstallToolCacheTtlMs: SYS_CMD_SEARCH_CACHE_DURATION_MS }, false)
+                this.executor?.execute(CommandExecutor.makeCommand('echo', ['PATH']), { shell: 'bin/bash', dotnetInstallToolCacheTtlMs: SYS_CMD_SEARCH_CACHE_DURATION_MS }, false)
```
nagilson, line 146:
```suggestion
                this.executor?.execute(CommandExecutor.makeCommand('echo', ['$PATH']), { shell: 'bin/bash', dotnetInstallToolCacheTtlMs: SYS_CMD_SEARCH_CACHE_DURATION_MS }, false)

```

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -85,7 +90,18 @@ export class CommandExecutor extends ICommandExecutor
...
+        const shellScript = this.sudoProcessScript;
+
+        try
+        {
+            await fs.promises.mkdir(this.sudoProcessCommunicationDir, { recursive: true });
+        }
+        catch (error: any)
+        {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            error.message = error.message + `\nFailed to create ${this.sudoProcessCommunicationDir}. Please check your permissions or install dotnet manually.`;
```
nagilson, line 102: This may happen in the rare event of a colission where the dir exists, so its ok to keep going here.

Forgind: Can you have fallback dirs? Like if dir exists, check dir1, then dir2, etc.? I don't feel great about catching any exception and not even verifying that it failed for the reason you mentioned. You could also have sudoProcessCommunication dir is unexpectedly null; that would be a bug, and we should care, I think

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -130,8 +146,8 @@ Please install the .NET SDK manually by following https://learn.microsoft.com/en
         }
         else
         {
-            if (await this.sudoProcIsLive(false, fullCommandString, 1000)) // If the sudo process was spawned by another instance of code, we do not want to have 2 at once but also do not waste a lot of time checking
-            // As it should not be in the middle of an operation which may cause it to take a while.
+            if (await this.sudoProcIsLive(false, fullCommandString, 500)) // If the sudo process was spawned by another instance of code, we do not want to have 2 at once but also do not waste a lot of time checking
```
Forgind, line 149: I'm not sure users would really notice the difference between 500ms and 1000ms. I don't know how long this is expected to take (IPC, right? I think speed then depends on OS), but I could imagine it taking longer than 500ms on a heavily loaded system

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -187,7 +203,7 @@ ${stderr}`));
     private async sudoProcIsLive(errorIfDead: boolean, fullCommandString: string, maxTimeoutTimeMs?: number, runCommand = false): Promise<boolean | CommandExecutorResult>
     {
         const processAliveOkSentinelFile = path.join(this.sudoProcessCommunicationDir, 'ok.txt');
-        const waitForLockTimeMs = maxTimeoutTimeMs ? maxTimeoutTimeMs : this.context?.timeoutSeconds ? (this.context?.timeoutSeconds * 1000 / 5) : 180000;
+        const waitForLockTimeMs = maxTimeoutTimeMs ? maxTimeoutTimeMs : (this.context?.timeoutSeconds !== undefined ? (Math.max(this.context.timeoutSeconds * 1000 / 5, 100)) : 180000);
```
Forgind, line 206: So if this.context is null, what is it supposed to do?

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -279,22 +295,24 @@ ${stderr}`));
...
+                this.context?.eventStream.post(new FailedToRunSudoCommand(`The command ${commandToExecuteString} failed to run: ${JSON.stringify(error ?? '')}.`));
                 // Let the rejected promise get handled below. This is required to not make an error from the checking if this promise is alive
             });
 
         commandOutputJson = {
-            stdout: (fs.readFileSync(stdoutFile, 'utf8')).trim(),
-            stderr: (fs.readFileSync(stderrFile, 'utf8')).trim(),
-            status: (fs.readFileSync(statusFile, 'utf8')).trim()
+            stdout: (await (this.fileUtil as FileUtilities).read(stdoutFile)).trim(),
+            stderr: (await (this.fileUtil as FileUtilities).read(stderrFile)).trim(),
```
Forgind, line 314: This defaults to utf8? I'm also wondering if you can have these three going in parallel then just 'join' them afterwards.

vscode-dotnet-runtime-library/src/Utils/FileUtilities.ts
```diff
@@ -96,8 +96,16 @@ export class FileUtilities extends IFileUtilities
...
-        return output;
+        try
+        {
+            const output = await fs.promises.readFile(filePath, 'utf8');
+            return output;
+        }
+        catch (error: any)
+        {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            throw new Error(`Failed to read file ${filePath}: ${error?.message}`);
```
Forgind, line 107: You don't expect filePath to be part of the message? I think this would kill your stack

vscode-dotnet-runtime-library/src/test/unit/TestUtility.ts
```diff
@@ -149,6 +149,10 @@ export async function getLinuxSupportedDotnetSDKVersion(context: IAcquisitionWor
         {
             return '6.0.100';
         }
+        if (distroInfo.version < '22.06')
```
Forgind, line 152: This is a very narrow range. It seems strange to me that you'd want 9.0.100 (the later version) on 22.04 (the earlier version) compared with 8.0.100 on 22.06. Why is that?

Forgind (commented):
Don't think any of these are too serious. I'm a little concerned about the undefined thing, but that might just be my lack of js knowledge

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -347,6 +365,7 @@ ${stderr}`));
         await executeWithLock(this.context.eventStream, false, RUN_UNDER_SUDO_LOCK(this.sudoProcessCommunicationDir), SUDO_LOCK_PING_DURATION_MS, this.context.timeoutSeconds * 1000 / 5,
             async () =>
             {
+                await (this.fileUtil as FileUtilities).wipeDirectory(this.sudoProcessCommunicationDir, this.context?.eventStream, ['.txt']);
```
MiYanni, line 368: What's the `.txt` referring to here? I'm a bit confused about what is exactly being written to disk.

MiYanni (approved):

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
nagilson: That's true. This check only exists for when there is an existing sudo process working in same the individualized folder.

That should only happen when we think we have not spawned a sudo process ourselves that is currently alive, or the sudo process spawned unsuccessfully while throwing an error. This check can probably be removed in its entirety.

vscode-dotnet-runtime-library/src/Utils/FileUtilities.ts
nagilson: Good point, in that I don't really think this is necessary anymore. Although, the error here is created within the same stack, but avoids including the nodejs internal stack, since it is created within our own code instead of within the node js library, so I actually prefer that, as it often gets truncated 🤔

vscode-dotnet-runtime-library/src/test/unit/TestUtility.ts
nagilson: Thank you for looking at this so thoroughly -- I hope this helps. The dotnet support for these minor versions seems to be different, as is the dotnet thats installed on these CI machines. When we get a different CI machine, either in dneng internal or public, it was failing with 22.05 because that expects .NET 9,  but the 22.04 machines expect .NET 8. If I fix that version but not 22.04, it will make the other machines fail.

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
nagilson: The `.txt` files are there to communicate with the sudo process. We can't use fork to talk to it, because the sudo library does not provide a method to do so. Probably it would have been easier at this point to go edit the vscode sudo library, but alas

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
nagilson: Yes! The function does default to utf8. I like the parallelization idea, though I want to avoid adding that complexity here

nagilson:
> Can you have fallback dirs? Like if dir exists, check dir1, then dir2, etc.? I don't feel great about catching any exception and not even verifying that it failed for the reason you mentioned. You could also have sudoProcessCommunication dir is unexpectedly null; that would be a bug, and we should care, I think

That's a good point. If we cant make the first dir, I doubt we can make the 2nd. Perhaps we could use the existing directory as a fallback, though that would introduce the race condition again if it happens multiple times. I could condition it to not fail on EEXIST. Though, I just don't know what other types of errors the kernel could throw at us :/

nagilson:
@Forgind, I think I resolved the 2 issues you put at the top by throwing the error in certain cases if we cant use mkdir or using a backupdir, and then by removing the extra 500 ms timeout check. Thanks for looking and let me know what you think :)

Forgind (approved):
I'm happy; thanks!

Issue #2059, opened 2024-11-30
.NET installation error


pamelafox:
### Describe the bug.

Every time I open a new VS Code window, an error pops up about .NET being unable to be installed. It's triggered from the Azure Copilot agent, as it requires .NET.

### Please attach log file(s) here if you have any.

11/29/2024, 4:43:17 PM DotnetWSLCheckEvent
true
Checking if system is WSL. OS: darwin

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.bat for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.clojure for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.coffeescript for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.configuration-editing for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.cpp for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.csharp for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.css for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.css-language-features for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.dart for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.debug-auto-launch for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.debug-server-ready for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.diff for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.docker for .NET installation requests

11/29/2024, 4:43:17 PM DotnetVSCodeExtensionFound
true
Checking extension vscode.emmet for .NET installation requests

[...]

PR #2106, opened 2025-01-28
Fix handler when install script makes an incomplete install

nagilson:
For https://github.com/dotnet/vscode-dotnet-runtime/issues/2059

PR #2235, opened 2025-04-11
Handle Corrupt Install Script Installs

nagilson:
The dotnet install script has a bug where it wont do the install if the folder is not empty. Sometimes it will leave a partially done install, like if killed, in which case we get to a bad state.

This fixes that by cleaning up the directory, and or checking if dotnet.exe exists, whether it works. dotnet --list-runtimes is a fast way to do this since it should only invoke the host and not need to invoke the SDK. doing just 'dotnet' will not work as that does not return 0.

Also, gets rid of the 'lock' exception since thats from the proper-lockfile lib which we don't use anymore.

Resolves: https://github.com/dotnet/vscode-dotnet-runtime/issues/2059
First scrapped attempt: https://github.com/dotnet/vscode-dotnet-runtime/pull/2106

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
@@ -29,7 +29,7 @@ import { FileUtilities } from '../Utils/FileUtilities';
 import { InstallScriptAcquisitionWorker } from './InstallScriptAcquisitionWorker';
 
 import { IUtilityContext } from '../Utils/IUtilityContext';
-import { executeWithLock } from '../Utils/TypescriptUtilities';
+import { executeWithLock, getDotnetExecutable } from '../Utils/TypescriptUtilities';
 import { WebRequestWorkerSingleton } from '../Utils/WebRequestWorkerSingleton';
 import { LOCAL_LOCK_PING_DURATION_MS } from './CacheTimeConstants';
 import { DotnetInstall } from './DotnetInstall';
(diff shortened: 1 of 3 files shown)


Forgind (commented):
I'm not really clear as to why your `dotnet --list-runtimes` check really tests whether the installation is valid or not. Any part could be corrupted, right? And --list-runtimes skips a lot of that, so shouldn't we install or not install based entirely on user input, i.e., user says it's broken --> reinstall?

nagilson:
> I'm not really clear as to why your `dotnet --list-runtimes` check really tests whether the installation is valid or not. Any part could be corrupted, right? And --list-runtimes skips a lot of that, so shouldn't we install or not install based entirely on user input, i.e., user says it's broken --> reinstall?

That's a great point. I've changed it to validate that the install includes a runtime that is correct and matches the correct version/arch.

vscode-dotnet-sdk-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -304,49 +304,6 @@ suite('DotnetCoreAcquisitionExtension End to End', function ()
       await vscode.commands.executeCommand('dotnet-sdk.uninstallAll');
     }).timeout(standardTimeoutTime);
 
-    test('Uninstall Command', async () =>
```
MiYanni, line 307: Is this just an old test or is it no longer valid (or both)?

MiYanni (approved):

vscode-dotnet-sdk-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
nagilson: Both!

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
```diff
@@ -76,6 +79,30 @@ You will need to restart VS Code after these changes. If PowerShell is still not
                             windowsFullCommand = windowsFullCommand.replace('powershell.exe', powershellReference);
                         }
 
+                        // The install script can leave behind a directory in an invalid install state. Make sure the executable is present at the very least.
+                        if (await this.fileUtilities.exists(installContext.installDir))
+                        {
+                            const dotnetPath = path.join(installContext.installDir, getDotnetExecutable());
+                            if (await this.fileUtilities.exists(dotnetPath))
+                            {
+                                const validator = new DotnetConditionValidator(this.workerContext, this.utilityContext);
```
Forgind, line 88: I'm a bit concerned this could throw, and we might want to catch that.

Forgind (approved):
Seems like this more directly tests what you need

vscode-dotnet-runtime-library/src/Acquisition/AcquisitionInvoker.ts
nagilson: I double checked, and this code should never throw, but I understand the concern and good to check!

Issue #2175, opened 2025-03-12
SDK Installation on RHEL Should Rely on DNF Over YUM


nagilson:
### Describe the bug.

Users may see Unable to get latest .NET Installation Data: Please update the .NET Install Tool to version 2.0.2 or later." after clicking the button 'Install .NET SDK' in VS Code from DevKit in RHEL 9.5. 

This is why it is failing:
`rhel-9-for-x86_64-baseos-rhui-rpms': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried` when we try to use `yum`.

In recent RHEL versions, DNF should be preferred over Yum.

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

Install SDK via vscode

### Expected Behavior

 .

### .NET Install Tool Version

latest

### Your operating system? Please provide more detail, such as the version and or distro above.

N/A

PR #2252, opened 2025-04-29
RHEL: Prefer DNF Over Yum

nagilson:
Please see: https://github.com/dotnet/vscode-dotnet-runtime/issues/2175
Resolves: https://devdiv.visualstudio.com/DevDiv/_queries/query/?tempQueryId=b79dee94-a770-46a2-9d77-6beba92ee01a

https://cyberpanel.net/blog/dnf-vs-yum#:~:text=DNF%20is%20the%20next%2Dgeneration,of%20Red%20Hat%2Dbased%20systems.

We mainly want to support RHEL 8, 9 and both already use DNF. We dont support RHEL 7 or below.
Note: I noticed RHEL does not support dotnet until it is registered. But that is not related to this issue.
Confirmed working on RHEL 9.5 VM:
image

### DIFF
--- a/vscode-dotnet-runtime-library/distro-data/distro-support.json
+++ b/vscode-dotnet-runtime-library/distro-data/distro-support.json
@@ -799,7 +799,7 @@
         "searchCommand": [
             {
                 "runUnderSudo": false,
-                "commandRoot": "yum",
+                "commandRoot": "dnf",
                 "commandParts": [
                     "search",
                     "{packageName}",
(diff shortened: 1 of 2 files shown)


Forgind (approved):
Is there a way to try dnf then, if it fails, try yum? I don't see why it has to be one or the other.

richlander:
Small context: Azure Linux will be moving to `dnf` in a later version, so preferring `dnf` here would be aligned with that.

nagilson:
> Is there a way to try dnf then, if it fails, try yum? I don't see why it has to be one or the other.

That's a good question, and it's good to be safe. Though in RHEL 8, yum is actually just an alias for dnf, so there shouldnt be a case where yum works but dnf does not. Then RHEL 9 is even more on the dnf train and yum is not advised.

PR #2284, opened 2025-05-22
Quick Attempt To Preemptively Load Cache

nagilson:
Calling dotnet --info, --list-runtimes etc is time consuming. We can do this at activation instead of later making assumptions about what most callers will request to speed this up. This does cause some extra work to be done, so this is an experiment.

We can more intelligently find the common case in the future.


nagilson:
Local Machine Testing, 3x:
 
New Project Without Cache Init: 3.47s Avg
New Project With Cache Init: 2.76s Avg
(Values were +- .20s)
 
Load Time (VS Code Initialization, -> VS Code Runtime Extension Start to Found Path Finish) 
This includes --info, and also all of the other work that may happen while our extension thread is running, though it only includes the time from which it gets the thread to when it finishes the API request during initialization
With Cache Init: 6745-> 7266 -> 5322 ms
Without Cache Init: 7696 -> 5304 -> 1995 ms

nagilson:
Exploring other methods to improve NPD as --info is 1 second of the time consumed according to other testers and this may not predict the cache correctly as it is invoked 2ce

PR #2288, opened 2025-05-22
Skip checking install requests for known extensions 

nagilson:
We check the package.json file for each extension which is an expensive wholistic file read that happens async during startup. This is slow and we can skip that for many extensions that we know will never need to ask us to install or find dotnet.

What I did to generate this list:
- Look at the extensions.all output on an empty (no custom extensions) code
- Look at the marketplace and add all first party / business partner related extensions I saw on the page that had the highest number of downloads

This is not an endorsement of these extensions or their codebases.
There is not a way to filter these or get anything besides .all of the extensions at this time.

cc @adrianvmsft @JakeRadMSFT @lifengl


vscode-dotnet-runtime-library/src/VSCodeExtensionContext.ts
```diff
@@ -30,17 +30,18 @@ export class VSCodeExtensionContext extends IVSCodeExtensionContext
...
+        {
             f(...(args));
         })
     }
 
-    public getExtensions() : readonly vscode.Extension<any>[]
+    public getExtensions(): readonly vscode.Extension<any>[]
     {
-        return vscode.extensions.all;
+        return vscode.extensions.all;;
```
nagilson, line 41:
```suggestion
        return vscode.extensions.all;
```

nagilson:
Set is indeed implemented typically using an o(1) data structure such as a hashtable in JS based on my research. So, it may be slightly memory intensive, but considering we do the operation a great number of times, it's likely worth this cost. Can run through with the perf review team.

vscode-dotnet-runtime-library/src/Acquisition/JsonInstaller.ts
```diff
@@ -13,6 +13,131 @@ import { IJsonInstaller } from "./IJsonInstaller";
 
 export class JsonInstaller extends IJsonInstaller
 {
+    private readonly knownCommonExtensionIdsNotUsingDotnet: Set<string> = new Set<string>([
```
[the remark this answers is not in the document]
nagilson: Hm, do you mean inclusive instead of exclusive? I think that's a good idea, but I worry it would harm the adoption/usefulness of our API by only allowlisting certain extensions to use it. Worst case, I could see that coming up in an anti-competitive practice lawsuit, which concerns me.

Now, maybe we could add some other way for extensions to specify they want us to do something. However, this needs to be done at activation before other extensions can run. We could ask them to ship with an additional manifest file, perhaps... though I'm not sure if we would be privy to that logic, and that risks things like protected files / bad read access, so I think I prefer this approach. Thanks for taking a look 👀

Documentation/commands.md
```diff
@@ -47,4 +47,7 @@ lookup the PATH first to see if there's a matching install, etc.
     "version": "8.0",
     "mode": "aspnetcore"
 }
-```
\ No newline at end of file
+```
+
+Note: If you are a developer for a highly used extension or an extension that is provided by Microsoft, you may be in our 'skip' list to enhance performance.
+Please check the list if this is not working for you and ask us to exclude you: [Skipped Extensions](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/Acquisition/JsonInstaller.ts).
```
MiYanni, line 53: nit: Generally, markdown documents within the same repo should always use relative paths to those documents. That way, this link from any branch always points to the document version within that branch.
```suggestion
Please check the list if this is not working for you and ask us to exclude you: [Skipped Extensions](vscode-dotnet-runtime-library/src/Acquisition/JsonInstaller.ts).
```

MiYanni (approved):

Documentation/commands.md
nagilson: That's a great point. Thank you, fixed.

PR #2296, opened 2025-06-02
Skip calling `dotnet --info` by using `--list-x --arch` when available 

nagilson:
`dotnet --info` resolves in the sdk (C# code) instead of the host (C++) code. It also resolves workloads and a slew of other JIT-ed components which makes it very 'slow' in comparison to host commands.

We only call --info to get the architecture of an existing install. Thus, after some discussion in my issue here: https://github.com/dotnet/runtime/issues/108503, the runtime team https://github.com/dotnet/runtime/pull/116078 added support for `dotnet --list-runtimes --arch` and the same for `--list-sdks`.

What this flag does is show the installs that match the specified architecture given in arch. 
The previous behavior ignored other flags and also only printed what matched the host architecture (unknown, which is what we got from info.)

I've utilized this new API to improve performance by skipping `dotnet --info` when applicable. We still need to call it for some .NET hosts before the .NET 10 hosts that have this functionality. And the easiest way to check is by seeing if --arch invalid fails or not (in which case, it supports `--arch`.) This is somewhat unfortunate as it is slower, but the best we came up with. (See discussion in the aforementioned PR.) I added some rudimentary logic to skip this --arch support check, if the host is already known to be too old (e.g. younger than .NET 10) to prevent this from slowing down the existing scenario too much (most people are on .NET 8.) I also time-gated it to always check after a certain point, as mentioned in the comments. Somewhat hacky, but probably sufficient.


vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
```diff
@@ -145,6 +151,22 @@ Please set the PATH to a dotnet host that matches the architecture ${requirement
         return sdkInfo;
     }
 
+    private async hostSupportsArchFlag(dotnetExecutablePath: string, listDotnetInstallsStdout: string): Promise<boolean>
+    {
+        // https://github.com/dotnet/runtime/pull/116078 --arch was not added until .NET 10 to allow us to skip calling dotnet --info because that is slow, as it is not native code.
+        // However, --arch gets ignored if the host does not support it. The output is also identical with or without --arch.
+        // After discussion with the runtime team, the best way to determine if the host supports --arch is to call it with an invalid arch to see if it fails, because that only happens when --arch is supported.
```
MiYanni, line 158: This is more of an aloud note to myself, but it is obviously a problem that we have no clear delineation between native-based and non-native-based CLI operations. And the other problem is that the perf between those is so vastly different. I know AOT was discussed, but there must be a non-AOT solution and a solution that doesn't require pushing more functionality into the host itself (native code). 🤔

PR #2302, opened 2025-06-10
Don't find Locally Managed Runtimes

nagilson:
If we change this, everyone who calls the findPath API will stop getting runtime updates, because they call `findPath` before `acquire`. We need to update the runtimes we install (the users own managed runtimes are different) so we can't do this yet. Let's take this out but leave in the feature so we can just revert the lookup and then revert this part later after we implement the update story.

That will cause a big breaking change because it will change the surface area of the acquire API to no longer update by default but instead update over time periodically.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
+++ b/vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
@@ -475,8 +475,6 @@ suite('DotnetCoreAcquisitionExtension End to End', function ()
             // Then verify we can find the extension-managed runtime
             const result = await vscode.commands.executeCommand<IDotnetAcquireResult>('dotnet.findPath', findPathContext);
             assert.notExists(result, 'Should find a runtime');
-            assert.notExists(result!.dotnetPath, 'Should find a runtime path');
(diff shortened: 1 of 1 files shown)


vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -474,10 +474,11 @@ suite('DotnetCoreAcquisitionExtension End to End', function ()
 
             // Then verify we can find the extension-managed runtime
             const result = await vscode.commands.executeCommand<IDotnetAcquireResult>('dotnet.findPath', findPathContext);
-            assert.exists(result, 'Should find a runtime');
-            assert.exists(result!.dotnetPath, 'Should find a runtime path');
-            assert.equal(result!.dotnetPath.toLowerCase(), runtimePath.toLowerCase(), 'Should find the correct runtime path');
-        } finally
+            assert.notExists(result, 'Should find a runtime');
+            assert.notExists(result!.dotnetPath, 'Should find a runtime path');
+            assert.notEqual(result!.dotnetPath.toLowerCase(), runtimePath.toLowerCase(), 'Should find the correct runtime path');
```
nagilson:
```suggestion

```
Don't need these checks for now if the inner part is null, just revert it all

MiYanni (approved):

PR #2305, opened 2025-06-12
Don't block the world writing a log file

nagilson:
This is a silly point of slowness we don't need. I will test this in a bit.


nagilson:
In my testing (thread is shared, keep in mind, though this op blocks our api), the api now returned at 3:52:16 PM 169. The log file write finished at 309 ms. (confirmed by using wmic https://superuser.com/questions/937380/get-creation-time-of-file-in-milliseconds), so this saved 140 ms.

PR #2306, opened 2025-06-13
Turn off shell by default

nagilson:


nagilson:
Ah so only spawn has boolen|string for shell, not exec.... facepalm

vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
```diff
@@ -143,7 +143,7 @@ Please set the PATH to a dotnet host that matches the architecture ${requirement
             return [];
         }
 
-        const findSDKsCommand = CommandExecutor.makeCommand(`"${existingPath}"`, ['--list-sdks', '--arch', requestedArchitecture]);
```
nagilson, line 146: Does this break Linux/Mac scenarios in the shell? I think we need to look at this from a broader context.

vscode-dotnet-runtime-library/src/Utils/CommandExecutor.ts
```diff
@@ -440,7 +439,11 @@ ${stderr}`));
             options.cwd ??= path.resolve(__dirname);
 
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-            options.shell ??= true;
+            options.shell ??= os.platform() === 'win32' ? false : true;
+            // ^ Windows systemcalls (node.js process library uses process_wrap which uses process.cc which makes system calls)
+            // Windows seems to resolve the PATH by default in a system call. Unix system calls do not.
```
nagilson, line 444: Can we just respect shell and let the shell do this?

PR #2309, opened 2025-06-17
Use ELF/COFF/PE/MACHO Header to get Executable Architecture

nagilson:
We need to verify a user's install matches the architecture that the code application / extension wants to run under.

We used to do this with dotnet --info which is a fallback. Now, we can also do this with dotnet --list-runtimes --arch when supported. However, checking if it is supported requires an extra process call which is slower than reading the file header.

Once we hit dotnet 11, we will know that --arch is supported so we can skip this work of reading the file header. For now, this will allow us to gain a perf boost in the cases where this works.

Documented in the code is how this logic works. Additionally, I've added test executables (real .net 9 ones) to show it works. Finally, I added the copyfiles module to copy the test assets into the dist folder, as otherwise they dont get copied.

This is low risk because if this cannot determine the architecture the existing logic will fallback in. Also, it is easy to test with customers if they give us the binary headers of their file and this code does not work for some reason (the log will indicate --info etc are called)


vscode-dotnet-runtime-library/src/test/unit/DotnetConditionValidator.test.ts
```diff
@@ -372,4 +371,32 @@ suite('DotnetConditionValidator Unit Tests', function ()
         assert.isNull(sdks[0].architecture, 'Architecture should be null when --arch is not supported');
     }).timeout(defaultTimeoutTimeMs);
 
+    test('It does not call info or list-runtimes for known invalid architectures', async () =>
```
nagilson:
```suggestion
    test('It does not call info or list-runtimes for known architectures', async () =>
```

vscode-dotnet-runtime-library/src/Utils/ExecutableArchitectureDetector.ts
```diff
@@ -0,0 +1,202 @@
+/*---------------------------------------------------------------------------------------------
+*  Licensed to the .NET Foundation under one or more agreements.
+*  The .NET Foundation licenses this file to you under the MIT license.
+*--------------------------------------------------------------------------------------------*/
+
+import { readFileSync } from 'fs';
+
+export type Architecture = 'x86' | 'x64' | 'arm64' | 'other' | null;
```
MiYanni, line 8: What's `'other'` vs `null` in this context?

vscode-dotnet-runtime-library/src/Utils/ExecutableArchitectureDetector.ts
```diff
@@ -0,0 +1,202 @@
...
+            fd.copy(headerBuffer, 0, 0, 64);
+
+            // Check for valid file size
+            if (headerBuffer.length < 64)
+            {
+                return null;
+            }
+
+            // Detect file format based on magic numbers
+            if (this.isElfFormat(headerBuffer))
```
MiYanni, line 65: So, you're using magic to detect if it is elf format? 🧝 Is this code from Tolkien?

vscode-dotnet-runtime-library/src/Utils/ExecutableArchitectureDetector.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    public static IsKnownArchitecture(architecture: Architecture): boolean
+    {
+        return architecture !== null && architecture !== 'other';
+    }
+
+    /**
+     * Checks if the file is in ELF format
+     * Magic number: 0x7F 'ELF'
+     * Reference: https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header
```
MiYanni, line 94: Not as fun as a format as I was hoping. 😭

vscode-dotnet-runtime-library/src/Utils/ExecutableArchitectureDetector.ts
```diff
@@ -0,0 +1,202 @@
...
+            case 0x03: return 'x86';      // EM_386
+            case 0x3E: return 'x64';      // EM_X86_64
+            case 0xB7: return 'arm64';    // EM_AARCH64
+            default: return 'other';
+        }
+    }
+
+    /**
+     * Parse Mach-O format architecture
+     * Reference: https://github.com/apple-oss-distributions/xnu/blob/main/EXTERNAL_HEADERS/mach-o/loader.h
```
MiYanni, line 150: Reminded me of...
image

It is just anime mecha Batman, basically.

MiYanni (approved):

joeloff (approved):

vscode-dotnet-runtime-library/src/Utils/ExecutableArchitectureDetector.ts
nagilson: Checking the wiki page, there are dozens of architecture values possible. There is a test that shows this, with arm32, which we produce an arm32 dotnet. We dont need to worry about those in the context of dotnet/code because it is the intersection of what platforms both support, which is very limited to these values above.

vscode-dotnet-runtime-library/src/Utils/ExecutableArchitectureDetector.ts
nagilson: 😁  This is a comment from CoPilot when I asked it to provide documentation and sources for the file formats. I will clean this up..

vscode-dotnet-runtime-library/src/Utils/ExecutableArchitectureDetector.ts
```diff
@@ -0,0 +1,202 @@
...
+                F: 0x46
+            },
+            // Mach-O Magic: 0xCF 0xFA 0xED 0xFE (Little-endian format)
+            MACHO: {
+                CF: 0xCF,
+                FA: 0xFA,
+                ED: 0xED,
+                FE: 0xFE
+            },
+            // DOS/PE Magic: 'MZ' in ASCII (Mark Zebinski, format creator)
```
nagilson: It's Zbikowski, actually. Lol

vscode-dotnet-runtime-library/src/Utils/ExecutableArchitectureDetector.ts
nagilson: It... totally looks like it. Glad you had fun reviewing this one 🤣

PR #2317, opened 2025-06-23
Update Dependencies, Migrate to Node 22

nagilson:
VS Code 1.100.0 migrated from node 20 to node 22. We should stay aligned with VS Code. Because our binaries actually run under whichever version node in Code is. This can cause extreme issues such as a complete failure to load if the compiler or language feature changes, and we would not know for some time.

Switching the linting from Mac to Windows because I cant get permission to push native linux packages.


nagilson:
I think I needed to push a new update to run lint to the internal feed.

vscode-dotnet-runtime-extension/CHANGELOG.md
```diff
@@ -7,9 +7,26 @@ and this project adheres to [Semantic Versioning].
...
+
+Breaking change: No longer download the latest runtime per every single request with `acquire`.
+Instead, update periodically, to prevent slow down on startup when new tooling must be acquired to be secure.
+
+## [2.3.7] - 2025-7
+
+Improve performance with shell utilization.
+Improve error messages and output minimization.
+Add API surface to get dotnet --list-sdks --arch or dotnet --list-runtimes --arch output.
+
```
nagilson:
```suggestion
Improve error messages and output minimization.
Add API surface to get dotnet --list-sdks --arch or dotnet --list-runtimes --arch output.
```

nagilson: de-prioritizing the shell change as it is not as impactful

vscode-dotnet-runtime-extension/CHANGELOG.md
```diff
@@ -7,9 +7,22 @@ and this project adheres to [Semantic Versioning].
...
 
-## [2.3.6] - 2025-6
+## [3.0.0] - 2025-7
+
+Breaking change: No longer download the latest runtime per every single request with `acquire`.
+Instead, update periodically, to prevent slow down on startup when new tooling must be acquired to be secure.
+
+Improve error messages and output minimization.
+Add API surface to get dotnet --list-sdks --arch or dotnet --list-runtimes --arch output.
+## [2.3.6] - 2025-6-23
```
nagilson:
```suggestion
Improve error messages and output minimization.
Add API surface to get dotnet --list-sdks --arch or dotnet --list-runtimes --arch output.

## [2.3.6] - 2025-6-23

```

1es-azure-pipeline.yml
```diff
@@ -42,7 +42,7 @@ parameters:
     os: windows
     emoji: 🪟
   - name: NetCore1ESPool-Internal
-    image: 1es-ubuntu-2204
+    image: ubuntu-latest
```
MiYanni: These are the only images available to that agent pool:
image

`ubuntu-latest` isn't a valid image name.

1es-azure-pipeline.yml
```diff
@@ -95,8 +95,8 @@ extends:
         parameters:
           pool:
             name: Azure Pipelines
-            image: macOS-latest
-            os: macOS
+            image: windows-latest
+            os: windows
```
MiYanni: This, absolutely, should not be done. We should *never* use external Windows images. We use external macOS images because we don't have internal macOS agents.

1pr-azure-pipeline.yml
```diff
@@ -59,8 +59,8 @@ stages:
   - template: pipeline-templates/lint.yaml
     parameters:
       pool:
-        vmImage: macOS-latest
-        os: macOS
+        vmImage: windows-latest
+        os: windows
```
MiYanni: Same comment as above.

1es-azure-pipeline.yml
nagilson: There is further logic that was already utilizing this, I don't recall how/why that is the case. I looked at the 1es migration PR and some of it may have been a migration tooling issue/pre-existing issue, and some of it may be due to the fact that `name` was not passed to the parameter pool object, thus causing the azure pipelines pool to be used. I think I've fixed all of those now, good call

1es-azure-pipeline.yml
nagilson: It looks like `name` did not get propagated to the pool object which caused the default pool provider of azure pipelines to be used, so this was a valid image name, but it should not have been using that pool provider. I think this has been fixed now.

nagilson: I found the doc link to the images. I wish that it was better explained which section is for which pool, since there is no section for the netcore public pool, only the external test pooling in general 🤔

MiYanni (approved):

nagilson:
Thank you ✨

vscode-dotnet-runtime-extension/CHANGELOG.md
jmaa4: > ```suggestion
> Improve error messages and output minimization.
> Add API surface to get dotnet --list-sdks --arch or dotnet --list-runtimes --arch output.
> ```

# - []()

jmaa4: > de-prioritizing the shell change as it is not as impactful

[]()

Issue #2027, opened 2024-11-13
Option to disable annoying "Using configured .NET path" output


brokenbeta:
### Is your feature request related to a problem? Please describe.

Every single time I start VS Code, the .NET install tool spams out "**Using configured .NET path**"... to the console.

I already know, and it's just spam I have to close before I start working.

Also, the "Do Not Disturb" setting does not seem to work to stop this message from coming through.

### A clear and concise description of what you want to happen. Include any alternative solutions you've considered.

I would like a way to disable this message, so I don't have to close a pointless panel every time I start work. It's annoying.

### Describe the solution you'd like

Maybe some kind of config setting that disables this message or maybe the system just remembers the previous configuration and doesn't show it if it's the same as the last run.

lattaai8:
Hello, I tried to solve the issue.

This is what I did:

Add a new configuration option to suppress the 'Using configured .NET path' message and modify the OutputChannelObserver to respect this setting.

You can review changes in this commit: https://github.com/lattaai8/dotnet-vscode-dotnet-runtime-2027/commit/35623161b980b0640873922d5d3b9bc1e772761b.


> [!CAUTION]
> Disclaimer: The commit was created by Latta AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---
This issue was tried to solve for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

nagilson:
I would lean towards a boolean setting on our extension which allows you to suppress this message.

wes22-max:
Create a new hyper link deleting the pop window from your code write. Or you may be able to find your url settings and see if you can disable the function. Or Boolean function[settings/spam](url)

PR #2332, opened 2025-07-01
Add setting to disable messages in the output window

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2027

The output window in vscode, under the '.NET Install Tool' tab may be annoying/unnecessary to the user. This adds a setting to disable that output. We will want to do that for the automatic update, so this prepares for that as well.

### DIFF
--- a/vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
@@ -186,7 +186,7 @@ export class OutputChannelObserver implements IEventStreamObserver
 
     private startDownloadIndicator()
     {
-        this.downloadProgressInterval = setInterval(() => this.outputChannel.append('.'), 1000);
+        this.downloadProgressInterval = setInterval(() => this.outputChannel.append('.'), 3000);
     }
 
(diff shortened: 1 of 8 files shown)


vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
```diff
@@ -181,7 +186,7 @@ export class OutputChannelObserver implements IEventStreamObserver
 
     private startDownloadIndicator()
     {
-        this.downloadProgressInterval = setInterval(() => this.outputChannel.append('.'), 1000);
+        this.downloadProgressInterval = setInterval(() => this.outputChannel.append('.'), 3000);
```
MiYanni, line 189: Is this going from 1 to 3 seconds?

MiYanni (approved):

vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
nagilson: Yeah, we have an actual UX person now ✨ so we will revisit this UX soon.

PR #2333, opened 2025-07-01
Find Locally Managed Runtimes

nagilson:
This was implemented but then backed out for release in https://github.com/dotnet/vscode-dotnet-runtime/pull/2302.

This includes one fix which is to respect DOTNET_ROOT over the locally installed runtimes. Users who set DOTNET_ROOT typically have an explicit scenario that they want.

This was backed out because we need to do auto update of the local installed runtimes for this feature to be compatible with the API shape of keeping customers secure. We will do that in this release.


nagilson:
Note: This needs to be merged with a breaking change so I would pause to merge this until that is ready. But it is still ready for review.

MiYanni (approved):

PR #2339, opened 2025-07-10
Fix Bundling of Install Scripts

nagilson:
The .NET Install Scripts are not bundled correctly into the extension anymore. It looks like this is a relic of the CDN changes and the url endpoints no longer work. This is reflected in lower success rates across the telemetry. We should fix that and clean up the related code.

The script is bundled at build time based on the latest version. The script is only used if the online version cannot be reached. This is an existing behavior. Please contact me internally for any other concerns w.r.t this.

> ❗ Tip: Disable whitespace diff for this PR review.


vscode-dotnet-runtime-library/src/Acquisition/InstallScriptAcquisitionWorker.ts
```diff
@@ -27,26 +26,35 @@ export class InstallScriptAcquisitionWorker implements IInstallScriptAcquisition
...
-        this.scriptFilePath = path.join(__dirname, 'install scripts', `${scriptFileName}.${this.scriptFileEnding}`);
+        this.scriptFilePath = path.join(__dirname, 'install scripts', `${this.scriptFileName}.${this.scriptFileEnding}`);
         this.webWorker = WebRequestWorkerSingleton.getInstance();
         this.fileUtilities = new FileUtilities();
     }
 
+    private async getFallbackScript(): Promise<string>
+    {
+        const fallbackPath = this.getFallbackScriptPath();
+        if ((await this.fileUtilities.exists(fallbackPath)))
```
MiYanni, line 41: Does the extra `()` wrapping do something special? (Not sure of `await` logic in TS)

MiYanni (approved):

vscode-dotnet-runtime-library/src/Acquisition/InstallScriptAcquisitionWorker.ts
nagilson: Sometimes parans are needed depending on the order of operations with `await` but in this case they are not needed, that's a good point 🤔

Issue #2337, opened 2025-07-04
SDK install bug


LowJack187:
### Describe the bug.

Make sure your program can skin a cat multiple ways! Look for it in both places and new users will be able to actually install the thing.

The file /etc/os-release takes precedence over /usr/lib/os-release. Applications should check for the former, and exclusively use its data if it exists, and only fall back to /usr/lib/os-release if it is missing. Applications should not read data from both files at the same time. /usr/lib/os-release is the recommended place to store OS release information as part of vendor trees. /etc/os-release should be a relative symlink to /usr/lib/os-release, to provide compatibility with applications only looking at /etc.

Sauce: https://bbs.archlinux.org/viewtopic.php?id=251724

### Please attach log file(s) here if you have any.

Gen X doesn't need log files. ;-)

### Reproduction Steps

Remove os-release file @ /etc/os-release and/or break sim link to /usr/lib/os-release. This is to make it act like this info didn't get placed correctly.

### Expected Behavior

Microsoft employees to test their code!

### .NET Install Tool Version

2.3.6

### Your operating system? Please provide more detail, such as the version and or distro above.

Linux

### VS Code Version

Arch

nagilson:
`/etc/os-release` is the only file that is scanned currently and it is a good suggestion to scan `/usr/lib/os-release`. Systems running on systemd should include this file. The logic here detects whether we support the running distro. The running distros which are supported include Ubuntu and RHEL, and Debian which is community maintained. You are right that we should test when users delete a system file and have a fallback because it does happen😉 (or the OS may not include it, based on your specification above Arch does not, necessarily, though the UI does not support it ATM.)  

https://unix.stackexchange.com/questions/351557/on-what-linux-distributions-can-i-rely-on-the-presence-of-etc-os-release

Anyway this is very easy to do. We just need to fallback to `/usr/lib/os-release` if the other file does not exist in the `LinuxVersionResolver.ts`.

PR #2345, opened 2025-07-15
Add fallback for /usr/lib/os-release to detect distro

nagilson:
More context at https://github.com/dotnet/vscode-dotnet-runtime/issues/2337, but TLDR is in the code comments which is that this is recommended as an official fallback. Note that there are still distros which do not adhere to this specification but the ones we support 1st party do, and if the distro cannot be detected, the unsupported message is still given.

This also changes the detection, I believe it was written using 'cat' because a different approach from etc/os-release was first attempted to try to work across further platforms. Anyway, its very expensive to run cat for this so let's not do that.

There may be a string parsing error, I will see if the tests pass before I spend the time to go try this on a VM.


andrijzyn:
Sorry. but that Merge didn't fix Issue. This the latests - dotnet / .NET Install Tool / VS Code

All details about my configuration were written here - #2352 

image

image

PR #2353, opened 2025-07-18
Add flag for verbosity to disable messages by default

nagilson:
Some of these messages dont need to be shown.

### DIFF
--- a/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
+++ b/vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
@@ -331,16 +331,19 @@ abstract class DotnetAcquisitionFinalErrorBase extends DotnetAcquisitionError
 export class DotnetGlobalSDKAcquisitionError extends DotnetAcquisitionFinalErrorBase
 {
     public eventName = 'DotnetGlobalSDKAcquisitionError';
+    public verboseOutputOnly = true;
 }
 
 export class DotnetRuntimeFinalAcquisitionError extends DotnetAcquisitionFinalErrorBase
 {
     public eventName = 'DotnetRuntimeFinalAcquisitionError';
+    public verboseOutputOnly = true;
 }
 
 export class DotnetASPNetRuntimeFinalAcquisitionError extends DotnetAcquisitionFinalErrorBase
 {
     public eventName = 'DotnetASPNetRuntimeFinalAcquisitionError';
+    public verboseOutputOnly = true;
 }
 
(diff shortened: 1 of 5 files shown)


vscode-dotnet-runtime-library/src/EventStream/OutputChannelObserver.ts
```diff
@@ -27,14 +27,15 @@ export class OutputChannelObserver implements IEventStreamObserver
...
+    constructor(private readonly outputChannel: IOutputChannel, private readonly suppressOutput: boolean = false, private readonly highVerbosity: boolean = false)
     {
     }
 
 
     public post(event: IEvent): void
     {
-        if (this.suppressOutput)
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        if (this.suppressOutput || ((event as any)?.verboseOutputOnly && this.highVerbosity))
```
nagilson:
```suggestion
        if (this.suppressOutput || ((event as any)?.verboseOutputOnly && !this.highVerbosity))
```

vscode-dotnet-runtime-library/src/EventStream/EventStreamEvents.ts
```diff
@@ -1673,9 +1677,17 @@ export class DotnetAcquisitionPartialInstallation extends DotnetAcquisitionMessa
...
 {
     public readonly type = EventType.DotnetAcquisitionInProgress;
+    public verboseOutputOnly = true;
 
     public readonly eventName = 'DotnetAcquisitionInProgress';
-    constructor(public readonly install: DotnetInstall, public readonly requestingExtensionId: string | null) { super(); }
+    constructor(public readonly install: DotnetInstall, public readonly requestingExtensionId: string | null)
+    {
+        super();
+        if (requestingExtensionId === 'user')
```
joeloff, line 1686: so the high verbosity doesn't always apply? If you want to reduce output, shouldn't it be `false` by default?

nagilson: It doesn't - it's basically implemented as a 'tag' to a property where it should be only shown in high verbosity mode. Some text we still want to show regardless of the verbosity mode (at least in my opinion)

joeloff: Ah, got it. Thanks for clarifying

joeloff (approved):
LGTM. The other option might be to have a logging API that can take the verbosity as a parameter and if that's less than the allowed verbosity it simply does nothing. Similar to how the logging extensions in C# works.

PR #2355, opened 2025-07-18
Prep 2.3.7 Pre-Release, Try to make prerelease depending on condition

nagilson:
This version should be a pre-release, and we want to add logic to automate that and do it more often. Code insiders uses pre-release builds by default and we want to use that to ensure safety with updates before a broader audience.

### DIFF
--- a/1es-azure-pipeline.yml
+++ b/1es-azure-pipeline.yml
@@ -58,9 +58,6 @@ parameters:
   displayName: Prerelease
   type: boolean
   default: false
-  values:
-  - true
-  - false
 
 extends:
   template: v1/1ES.Official.PipelineTemplate.yml@1esPipelines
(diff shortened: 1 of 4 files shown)


1es-azure-pipeline.yml
```diff
@@ -54,6 +54,13 @@ parameters:
   type: string
   default: Test
   values: [ 'Test', 'Real' ]
+- name: PRERELEASE
+  displayName: Prerelease
+  type: boolean
+  default: false
+  values:
+  - true
+  - false
```
MiYanni: When type is `boolean`, I don't know if `values` does anything here...

MiYanni (approved):

1es-azure-pipeline.yml
nagilson:
```suggestion
```

nagilson: so trueee, fixed, thanks

Issue #2315, opened 2025-06-20
Provide API to get ListRuntimes and ListSDKs output


nagilson:
### Is your feature request related to a problem? Please describe.

The debugger would like to reuse our logic to get the architecture and installations so they don't use `dotnet --info`. We can share this info via an API available in version `2.3.7` as follows:

Command:
```ts
`dotnet.availableInstalls` 
```
Takes:
```ts
export interface IDotnetSearchContext :
{
mode : DotnetInstallMode // (`runtime` | `aspnetcore` | `sdk`) The 2 runtime modes behave the same and return both runtime types.
dotnetExecutablePath : string, // A full path is preferred, as 'dotnet' will fail if 'which' or 'where' is corrupted.
architecture : string | null // (accepts 'x64', 'x86', 'arm64'). Will default to the executable architecture if detectable, else the `os.arch()`.
}
```

Returns: 
`IDotnetListInfo[], empty if none found`
```ts
export interface IDotnetListInfo
 { mode: DotnetInstallMode, 
version: string, 
directory : string, 
architecture: string | null }; // Architecture is null if it cannot be determined, which should not happen except in cases such as: a dotnet executable built for arm32 or custom architecture such as `Silicon Graphics SVx`, or a corrupt dotnet executable.
```

### A clear and concise description of what you want to happen. Include any alternative solutions you've considered.

Utilize our cache and architecture API to reduce load.

### Describe the solution you'd like

see above.

JoeRobich:
Can we make `dotnetExecutablePath` optional? Also, it might be useful to include the path to the dotnet CLI as part of the results.

WardenGnaw:
+1 with making `dotnetExecutablePath` optional. 

The debugger is just looking for the sdk in the PATH used to determine what built the debuggee to figure out what kind of architecture it needs to debug.

nagilson:
In this case, we would utilize the logic for finding the path to deduce the path I assume? I like this idea, and happy to include that as well.

We could call both `dotnetPath : string (|undefined on the context)` to match the existing APIs. Perhaps we should, though it might be confusing to users who dont know the distinction between the `host`/`executable` vs a runtime or SDK.

nagilson:
> +1 with making `dotnetExecutablePath` optional.
> 
> The debugger is just looking for the sdk in the PATH used to determine what built the debuggee to figure out what kind of architecture it needs to debug.

[...]

PR #2366, opened 2025-07-23
Create API to get output of --list-runtimes and --list-sdks with arch

nagilson:
The motivation to do this is described in https://github.com/dotnet/vscode-dotnet-runtime/issues/2315 as well as the spec.

This will enable the caching and performance related code to be shared and was mainly implemented for the debugger so they can remove their call to dotnet --info.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/extension.ts
+++ b/vscode-dotnet-runtime-extension/src/extension.ts
@@ -53,6 +53,7 @@ import
     IDotnetConditionValidator,
     IDotnetEnsureDependenciesContext,
     IDotnetFindPathContext,
+    IDotnetListInfo,
     IDotnetListVersionsContext,
     IDotnetListVersionsResult,
     IDotnetUninstallContext,
@@ -116,6 +117,8 @@ namespace commandKeys
     export const ensureDotnetDependencies = 'ensureDotnetDependencies';
     export const reportIssue = 'reportIssue';
     export const resetData = 'resetData';
+
+    export const availableInstalls = 'availableInstalls';
 }
 
 const commandPrefix = 'dotnet';
@@ -400,6 +403,36 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
         }, getIssueContext(existingPathConfigWorker)(commandContext.errorConfiguration, 'acquireStatus'));
         return pathResult;
     });
+
+
+    const dotnetAvailableInstallsRegistration = vscode.commands.registerCommand(`${commandPrefix}.${commandKeys.availableInstalls}`,
+        async (commandContext: IDotnetSearchContext): Promise<IDotnetSearchResult[]> =>
+        {
+            if (commandContext.mode === undefined || commandContext.requestingExtensionId === undefined)
+            {
+                throw new EventCancellationError('BadContextualAvailbleInstallsError', `The dotnet.availableInstalls API request was missing either a mode or requestingExtensionId. Please provide this.`);
+            }
+
+            const installs = await callWithErrorHandling(async () =>
+            {
+                const workerContext = getAcquisitionWorkerContext(commandContext.mode, commandContext);
+                const dotnetResolver = new DotnetConditionValidator(workerContext, utilContext);
+                const installsInListForm: IDotnetListInfo[] = await dotnetResolver.getDotnetInstalls(commandContext.dotnetExecutablePath ?? 'dotnet', commandContext.mode, commandContext.architecture);
+
+                return installsInListForm.map((installInfo: IDotnetListInfo) =>
+                {
+                    return {
+                        mode: installInfo.mode,
+                        version: installInfo.version,
+                        directory: installInfo.directory,
+                        architecture: installInfo.architecture ?? DotnetCoreAcquisitionWorker.defaultArchitecture(),
+                    } as IDotnetSearchResult;
+                });
+            }, getIssueContext(existingPathConfigWorker)(commandContext?.errorConfiguration, commandKeys.availableInstalls));
+
+            return installs ?? [];
+        });
+
 
     const resetDataPublicRegistration = vscode.commands.registerCommand(`${commandPrefix}.${commandKeys.resetData}`, async () =>
     {
--- a/vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/DotnetConditionValidator.ts
@@ -12,6 +12,7 @@ import { ICommandExecutor } from '../Utils/ICommandExecutor';
 import { IUtilityContext } from '../Utils/IUtilityContext';
 import { DOTNET_INFORMATION_CACHE_DURATION_MS } from './CacheTimeConstants';
 import { DotnetCoreAcquisitionWorker } from './DotnetCoreAcquisitionWorker';
+import { DotnetInstallMode } from './DotnetInstallMode';
 import { IAcquisitionWorkerContext } from './IAcquisitionWorkerContext';
 import { IDotnetConditionValidator } from './IDotnetConditionValidator';
 import { IDotnetListInfo } from './IDotnetListInfo';
@@ -28,7 +29,7 @@ export class DotnetConditionValidator implements IDotnetConditionValidator
         this.executor ??= new CommandExecutor(this.workerContext, this.utilityContext);
     }
 
-    public async dotnetMeetsRequirement(dotnetExecutablePath: string, requirement: IDotnetFindPathContext): Promise<boolean>
+    public async getDotnetInfo(dotnetExecutablePath: string, mode: DotnetInstallMode, requestedArchitecture: string | undefined | null): Promise<IDotnetListInfo[]>
     {
         const hostArch = new ExecutableArchitectureDetector().getExecutableArchitecture(dotnetExecutablePath);
         const oldLookup = process.env.DOTNET_MULTILEVEL_LOOKUP;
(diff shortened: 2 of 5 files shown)


nagilson:
FYI @WardenGnaw, hoping to merge this into our next version

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
```diff
@@ -0,0 +1,223 @@
...
+    {
+        // https://github.com/dotnet/runtime/pull/116078 --arch was not added until .NET 10 to allow us to skip calling dotnet --info because that is slow, as it is not native code.
+        // However, --arch gets ignored if the host does not support it. The output is also identical with or without --arch.
+        // After discussion with the runtime team, the best way to determine if the host supports --arch is to call it with an invalid arch to see if it fails, because that only happens when --arch is supported.
+
+        // The --arch flag was added in the middle of .NET 10, so we can assume it is supported if the version is 10.0 or later.
+        // We don't want to slow down the current common case for people without .NET 10 by adding another process spawn check.
+        // We don't check that the version is 10.0 or later after 2026 when .NET 11 starts rolling out, as It will be slower to check all of the numbers in the output for versions >= 10.
+        const hostMaySupportArchFlag = listDotnetInstallsStdout.includes("10.0") || listDotnetInstallsStdout.includes("11.0") || Date.now() >= new Date('2026-03-01').getTime();
```
WardenGnaw, line 183: Is there a scenario where the user only has an older version of .NET but this will now always return `true` if we are past March 1st, 2026?

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
```diff
@@ -0,0 +1,223 @@
...
+    public async getSDKs(existingPath: string, requestedArchitecture: string, knownArchitecture: boolean): Promise<IDotnetListInfo[]>
+    {
+        if (!existingPath || existingPath === '""')
+        {
+            return [];
+        }
+
+        const findSDKsCommand = CommandExecutor.makeCommand(`"${existingPath}"`, ['--list-sdks', '--arch', requestedArchitecture]);
+
+        const sdkInfo = await (this.executor!).execute(findSDKsCommand, { dotnetInstallToolCacheTtlMs: DOTNET_INFORMATION_CACHE_DURATION_MS }, false).then(async (result) =>
```
WardenGnaw: Unless I'm missing something, it looks like `execute` returns a promise. Is there a reason why you are mixing `await` and `.then(...)`?

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
```diff
@@ -0,0 +1,223 @@
...
+        {
+            if (result.status !== '0')
+            {
+                return [];
+            }
+
+            const architectureKnown = knownArchitecture ? true : await this.hostSupportsArchFlag(existingPath, result.stdout);
+            const sdks = result.stdout.split('\n').map((line) => line.trim()).filter((line) => (line?.length ?? 0) > 0);
```
WardenGnaw: Since this is a machine specific output, do you want to split on `os.EOL`?

vscode-dotnet-runtime-library/src/Acquisition/IDotnetResolver.ts
```diff
@@ -0,0 +1,14 @@
...
+*  Licensed to the .NET Foundation under one or more agreements.
+*  The .NET Foundation licenses this file to you under the MIT license.
+*--------------------------------------------------------------------------------------------*/
+
+import { DotnetInstallMode } from './DotnetInstallMode';
+import { IDotnetListInfo } from './IDotnetListInfo';
+
+export interface IDotnetResolver
+{
+    getDotnetInstalls(dotnetExecutablePath: string, mode: DotnetInstallMode, requestedArchitecture: string | undefined | null): Promise<IDotnetListInfo[]>
```
WardenGnaw, line 11: Should jsdoc comments be added here to clarify what these methods do?

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
```diff
@@ -0,0 +1,223 @@
...
+            const sdkInfos: IDotnetListInfo[] = sdks.map((sdk) =>
+            {
+                const parts = sdk.split(' ', 2);
```
WardenGnaw: Is there a reason why you are limiting this `split` but doing the `split` again on line 152 to join the remainders?

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
WardenGnaw: A fun way to clear out your `falsey` values is to just do `.filter(Boolean)` instead of `.filter((line) => (line?.length ?? 0) > 0)`

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
nagilson: That's a great question. To explain the assumptions of this code, which exists in prod today:
- This code only runs to check if `--arch` is supported or not, which has a performance hit 
- We only need to run `--arch` in rare cases, when the executable architecture cannot be read from the binary header. This is typically only true for a corrupt binary, or a binary made for an unexpected architecture 
- We know --arch will always be supported in .NET 11 and after
- Instead of checking whether there is a number in the string >= 10, we assume by the release date window of .NET 11 that the typical user will have .net 10 or .net 11, so we always run the check to see if --arch is supported, so we don't need to account for .net 12, 13, etc etc. It would be slower to check for any number in the string vs for "10.0" or "11.0"

In short, this should never cause an issue, besides a perf regression after that date, but only for edge case scenarios, and only for when users are running on older versions of .NET, which by this point, CDK likely wont run on anyway

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
nagilson: That's a great point... I thought it would be cleaner to return the correct value by processing the return result of the promise in the then chain, but that seems like an anti pattern. Fixed.

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
nagilson: That's a great point. I've never had this be an issue but this is great. That must be because of the parsing that we do later, which would remove the /r on win

vscode-dotnet-runtime-library/src/Acquisition/IDotnetResolver.ts
nagilson: Great idea, added. Thanks 😄

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
nagilson: That's... awesome and my code will forever be better from you telling me that, thank you ✨

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
nagilson: Yes, we can't split the path based on spaces, it may have spaces in it. So we do the next split to get the path by finding whats between the [ ]

nagilson: We could use regex to avoid doing 2 .splits. I think it makes the code much harder to read... but maybe that's my issue

nagilson: Im not even sure if thats faster. I would probbaly have to just... test it and see for this scenario 🤔  I think this is great feedback and maybe worth making an issue on. But there are things I can spend time on atm that would impact perf significantly more, so I likely won't prioritize this one for now

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
WardenGnaw: Sounds good to me!

sample/src/extension.ts
```diff
@@ -264,6 +264,42 @@ ${stderr}`);
...
+
+        let searchMode = await vscode.window.showInputBox({
+            placeHolder: 'runtime',
+            value: 'runtime',
+            prompt: 'look for an sdk, runtime, aspnetcore runtime, etc',
+        });
+
+        try
+        {
+            const result = await vscode.commands.executeCommand<IDotnetAcquireResult>('dotnet.availableInstalls', { architecture: arch, requestingExtensionId: requestingExtensionId, mode: searchMode, dotnetExecutablePath: dotnetPath });
```
MiYanni, line 293: If `dotnetPath` is undefined, what happens here?

sample/src/extension.ts
```diff
@@ -264,6 +264,42 @@ ${stderr}`);
...
+            vscode.window.showInformationMessage(`.NET Discovered:\n
+${JSON.stringify(result) ?? 'undefined'}`);
```
MiYanni, line 295: Does `stringify` return a potential value that works with `??`?

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
```diff
@@ -0,0 +1,239 @@
...
+    * Thus, any runtime or sdk that it prints out will be the same architecture as the host.
+    * This information is not always accurate as dotnet info is subject to change.
+    *
+    * @remarks Will return '' if the architecture cannot be determined for some peculiar reason (e.g. dotnet --info is broken or changed).
+    */
+    // eslint-disable-next-line @typescript-eslint/require-await
+    private async getHostArchitectureViaInfo(hostPath: string, expectedArchitecture?: string | undefined | null): Promise<string>
+    {
+        // dotnet --info is not machine-readable and subject to breaking changes. See https://github.com/dotnet/sdk/issues/33697 and https://github.com/dotnet/runtime/issues/98735/
+        // Unfortunately even with a new API, that might not go in until .NET 10 and beyond, so we have to rely on dotnet --info for now.*/
```
MiYanni: nit: stray `*/`?

```suggestion
        // Unfortunately even with a new API, that might not go in until .NET 10 and beyond, so we have to rely on dotnet --info for now.
```

MiYanni (approved):

sample/src/extension.ts
nagilson: Yes!

sample/src/extension.ts
nagilson: 'dotnet' is used instead by the API; good question. We need to support that based on the IDotnetSearchContext specification.

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
nagilson: Actually it looks like this could be a breaking change and cause some failures due to the comment Ive added below, so I undid this.

WardenGnaw: No worries, I wonder if the `dotnet` executable only emits newline outputs so this split is fine.

nagilson: It produces both, but os.EOL may be wrong. When we trim, we get rid of the /r on windows, so splitting by /n is fine

PR #2370, opened 2025-07-28
Enable Automatic Update

nagilson:


nagilson:
That's right, and great point - this needs to be updated to do updates daily.

Issue #1052, opened 2023-07-14
VSCode extension customers can acquire and interact with global .NET installs


nagilson:
# EPIC - Add Global .NET SDK Installs for DevKit, & Others

See https://github.com/dotnet/vscode-dotnet-runtime/pull/1046 for an overview of major changes made and our plan.
cc
@marcpopMSFT
@baronfel @leslierichardson95  @webreidi 
@JL03-Yue 

## Basic API 
- [x] Change the API to match the specifications requested
https://github.com/dotnet/vscode-dotnet-runtime/issues/763
https://github.com/dotnet/vscode-dotnet-runtime/pull/1081
- [x] Create an API to list the newest available SDKs and Runtimes/
https://github.com/dotnet/vscode-dotnet-runtime/issues/764
- [x] Create an API to suggest what version of the .NET SDK or Runtime to Install
https://github.com/dotnet/vscode-dotnet-runtime/pull/791
- [x] Create a resolver that figures out what installer you need to install based on your request

## Windows and Mac
- [x] Create the installer that installs the .NET SDK globally for windows.
https://github.com/dotnet/vscode-dotnet-runtime/issues/763
https://github.com/dotnet/vscode-dotnet-runtime/pull/1081
- [x] Add logic to read the registry to prevent overhead of installing when it is not needed.
- [x] Create the installer that installs the .NET SDK globally for Mac.
- [x] Perform and improve error handling that comes out from the existing installer.

## Existing Code Improvement
- [x] Migrate to a new cache for web-requests so the cache does not get stale.
https://github.com/dotnet/vscode-dotnet-runtime/issues/416
https://github.com/dotnet/vscode-dotnet-runtime/pull/799

## Global Runtimes
- [ ] Create the ability to manage global runtimes.
https://github.com/dotnet/vscode-dotnet-runtime/issues/1055
- [ ] Add an API to list the installed global runtimes on the machine.
https://github.com/dotnet/vscode-dotnet-runtime/issues/1055

## Threat Model
- [x] Create a threat model for our global extension installs.
https://github.com/dotnet/vscode-dotnet-runtime/issues/1053
- [x] Add sign-checking to .NET installers that we download from the interwebs. 
- [x] Investigate if/how we can sign and validate signatures in typescript.

## Linux
- [x] Add the ability to manage global installs on Linux
https://github.com/dotnet/vscode-dotnet-runtime/issues/763
[...]

PR #2373, opened 2025-08-01
Simplify Mac Installation Process

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/1052 (That story is technically already resolved but this closes it out)

The mac installer currently always requires user input to go through the UI dialog prompt. This gets missed by a large number of users. Worse yet, we use 'open' so we don't parse the user password, but open doesn't allow us to get the exit code of the installer, so we don't know if someone cancelled, closed out the installer, left their pc, or rejected the password prompt, and that gets counted as a failure on our side.

This updates the method to use 'installer' and the sudo prompt to elevate and install. The sudo prompt must be used with caution with gui applications, as the files it touches can become owned by 'root.' Luckily in this case, the installer is meant to be executed with admin privileges.

An alternative is to use osascript -e 'do shell script "..." with administrator privileges' but this doesn't seem to be able to parse properly in this context.

In the event users have issues with 'installer' such as it being disabled by a policy, we fallback to the old method. However, now we collect data in the fallback to allow us to better count the cases when a user leaves/denies the request.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/InstallationValidator.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/InstallationValidator.ts
@@ -114,7 +114,7 @@ export class InstallationValidator extends IInstallationValidator
         } catch (error)
         {
             return this.handleValidationError(
-                `${baseErrorMessage} Unable to verify that "${dotnetPath}" is a file: ${error}`,
+                `${baseErrorMessage} Unable to verify that "${dotnetPath}" is a file: ${JSON.stringify(error ?? '')}`,
                 context,
                 failOnErr
             );
@@ -155,7 +155,7 @@ export class InstallationValidator extends IInstallationValidator
         catch (error)
         {
             return this.handleValidationError(
-                `${baseErrorMessage} Unable to read dotnet directory "${dotnetPath}": ${error}`,
+                `${baseErrorMessage} Unable to read dotnet directory "${dotnetPath}": ${JSON.stringify(error ?? '')}`,
                 context,
                 failOnErr
             );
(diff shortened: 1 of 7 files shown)


vscode-dotnet-runtime-library/src/Acquisition/InstallationValidator.ts
```diff
@@ -3,69 +3,209 @@
...
 
-            this.assertOrThrowError(failOnErr, fs.lstatSync(dotnetPath).isFile(),
-                `${dotnetValidationFailed} Expected executable file exists but is not a file: "${dotnetPath}"`, install, dotnetPath);
+        // Check if parent directory exists
+        if (!this.validateCondition(
+            fs.existsSync(parentDirectory),
+            `${baseErrorMessage} Expected installation folder ${parentDirectory} does not exist.`,
+            context,
+            failOnErr
+        ))
```
MiYanni, line 86: nit: I notice multiple if-statements below have this `))` format, which I find a bit weird. I understand it, but it just looks out-of-place. It would probably look a little less weird if both `))` and `{` were on the same line like `)) {` because parens usually signify the end of an expression an squiggly brackets are the start/end of a statement scope (which is why they are separate lines in C#, but not normally in TS, but you have your own formatting for this repo).

MiYanni (approved):

vscode-dotnet-runtime-library/src/Acquisition/InstallationValidator.ts
nagilson: I agree that's odd. I'll look into seeing if there's something I can do with the auto-formatter to prevent this from happening.

PR #2375, opened 2025-08-01
[Proof of Concept] Enable Code GitHub CoPilot to Install, Uninstall the .NET SDK / Runtimes

nagilson:
Related: https://github.com/microsoft/vscode-dotnettools/issues/2090#issuecomment-3070467289
https://github.com/microsoft/vscode/issues/255860

# Warning - This is only a proof of concept

This code is very draft (lots done with CoPilot), so it needs further review, such as checking the policy for whether we can install an MCP server or not and whether that's been disabled, validating that the install works and doesn't break other versions of code, and does not consider whether we should ship an MCP server within an extension. In addition, I haven't checked that the js file for the MCP server bundles properly, and currently its just a hardcoded source file. Finally, this assumes node is available on the dev machine, which is not always the case. We would want to ship a platform agnostic binary which would likely need to be signed and sign validated. Also, the PATH update doesn't seem to always trigger in this scenario. All of that would add a lot of overhead, this is more of a Proof of Concept.

# What it does 
- The VS Code chat agent is not aware or able to call commands through the command palette.
- This creates a 2 step process to make that happen. It's a bit hacky. 
- 1. Launches a process from within the extension that listens on a special socket and runs our specific vs code commands
- 2. Registers a node.js file that communicates with the above process and is callable via any terminal session via `node`
- 3. Registers an MCP Server on startup that is aware of the above node.js process, so that the chat can use it.
- 4. Our API surface already protects from 'injection' and unsanitized strings.

# Observations

It seems to be able to install the SDK, which is awesome. 
It also seems to be able to parse our informational and error diagnostics and provide further information, such as when I ask to re-install a version that's already on the box.

image

image

{E7EF3255-4B20-4420-83D4-2485A6F32A3B}

### DIFF
--- a/sample/yarn.lock
+++ b/sample/yarn.lock
@@ -86,23 +86,23 @@
     tslib "^2.6.2"
 
 "@azure/msal-browser@^4.2.0":
-  version "4.18.0"
-  resolved "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@azure/msal-browser/-/msal-browser-4.18.0.tgz"
-  integrity sha1-6hETdmmRb2rKZ0SKQk4ClBf8md4=
+  version "4.15.0"
+  resolved "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@azure/msal-browser/-/msal-browser-4.15.0.tgz"
+  integrity sha1-BZ0nASdVDCUCoNBgkKvADCx4Iu8=
   dependencies:
-    "@azure/msal-common" "15.9.0"
+    "@azure/msal-common" "15.8.1"
 
-"@azure/msal-common@15.9.0":
-  version "15.9.0"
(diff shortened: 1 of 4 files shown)


nagilson:
This reminds me of @ellahathaway's MCP - hers provides a lot of intelligent functionality. It would be cool to tie these together, but this only works in VS Code. I think part of the plan of the 'bootstrapper' project would be to enable their MCP to do the install and updates based on the things like CVE lists.

ellahathaway:
cc'ing some other folks for visibility: @richlander, @MichaelSimons, @mthalman

nagilson:
Note that https://github.com/microsoft/vscode/issues/255860 would likely remove the need to do this work, this is simply a workaround of the fact that the agent isn't aware of the extension contributions at the moment. I think to 'productify' this PR it would take enough time that this issue may be resolved by the time is got resolved. (We should need to sign a binary, go through security review, etc.) I think we should see what functionality that code issue enables and whether we still need work to do after the fact to improve the experience.

PR #2385, opened 2025-08-08
Fix VSIX signing and packaging process

JoeRobich:
The new signing process requires signing a signature files created by generating a manifest from the VSIX package. All three files are necessary for publishing the VSIX package correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,6 @@
 	},
 	"dependencies": {
 		"@typescript-eslint/eslint-plugin-tslint": "^7.0.2",
-		"@vscode/vsce": "^2.26.1",
 		"eslint-import-resolver-typescript": "^3.6.3",
 		"eslint-plugin-import": "^2.30.0",
 		"eslint-plugin-jsdoc": "^50.2.2",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -28,6 +27,7 @@
 		"@types/source-map-support": "^0.5.6",
 		"@typescript-eslint/eslint-plugin": "^8.0.0",
 		"@typescript-eslint/parser": "^8.0.0",
+		"@vscode/vsce": "^3.6.0",
```
JoeRobich, line 30: Updated to bring in the generate-manifest command

pipeline-templates/package-vsix.yaml
```diff
@@ -45,7 +45,10 @@ jobs:
       else
         vsce package -o $(package-name)-$(GetVersion.version).vsix --ignoreFile ../.vscodeignore --yarn
       fi
+      vsce generate-manifest -i $(package-name)-$(GetVersion.version).vsix -o $(package-name)-$(GetVersion.version).manifest
       cp $(package-name)-$(GetVersion.version).vsix ../packages/$(package-name)-$(GetVersion.version).vsix
+      cp $(package-name)-$(GetVersion.version).manifest ../packages/$(package-name)-$(GetVersion.version).manifest
+      cp $(package-name)-$(GetVersion.version).manifest ../packages/$(package-name)-$(GetVersion.version).signature.p7s
```
JoeRobich, line 51: The manifest is copied twice. The second time as the signature file which will be signed.

JoeRobich:
> Thanks for this, though It looks like this got reverted in the internal vs-green repo. I wonder if that's due to the new signing requirement announced Wednesday on top of this one? Blergh.
 
@nagilson  It got reverted because I was still signing the VSIX instead of the signature file. 🤦 The same issue this PR attempts to address for this repo.

Issue #2380, opened 2025-08-05
[NETE2ESDK] The parameter .NET runtime architecture in the command 'Sample: Find the .NET Installs on the PATH' does not work.


NicoleWang001:
### Describe the bug.

On Windows Arm64, .NET with arm64 architecture is discovered even 'x64' architecture is required.

### Please attach log file(s) here if you have any.

[logs.zip](https://github.com/user-attachments/files/21594265/logs.zip)
The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

1. On Windows Arm64 OS, install .NET SDK e.g. 8.0.412
2. Run 'Sample: Find the .NET Installs on the PATH' command, provide the .NET Host path to Scan as 'undefined', the .NET runtime architecture as '**x64**' rather than 'arm64' and the look for as 'runtime' or 'sdk'
3. Observe

### Expected Behavior

.NET with arm64 architecture should not be discovered.

### Exceptions & Errors (if any)

.NET with arm64 architecture is discovered even 'x64' architecture is required.

<img width="896" height="837" alt="Image" src="https://github.com/user-attachments/assets/ccdae304-b472-4532-bcde-9f0dae7ba64b" />

### .NET Install Tool Version

v2.3.8

### Your operating system? Please provide more detail, such as the version and or distro above.

N/A

### VS Code Version

1.102.3

PR #2391, opened 2025-08-13
Fix New Architecture Refactor in Unreleased 2.3.8

nagilson:
Fixes https://github.com/dotnet/vscode-dotnet-runtime/issues/2380


vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
```diff
@@ -82,6 +95,117 @@ export class DotnetResolver implements IDotnetResolver
...
+                {
+                    // The .NET install layout is a well known structure on all platforms.
+                    // See https://github.com/dotnet/designs/blob/main/accepted/2020/install-locations.md#net-core-install-layout
+                    //
+                    // Therefore we know that the runtime path is always in <install root>/shared/<runtime name>
+                    // and the dotnet executable is always at <install root>/dotnet(.exe).
+                    //
+                    // Since dotnet --list-runtimes will always use the real assembly path to output the runtime folder (no symlinks!)
+                    // we know the dotnet executable will be two folders up in the install root.
+                    const truePath = path.join(path.dirname(path.dirname(runtimeInfo[0].directory)), getDotnetExecutable());
```
joeloff, line 172: Do you care if the path exists, or is this just to determine what the expected path should be?

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
```diff
@@ -82,6 +95,117 @@ export class DotnetResolver implements IDotnetResolver
...
+                    if (os.platform() === 'win32')
+                    {
+                        LocalMemoryCacheSingleton.getInstance().aliasCommandAsAnotherCommandRoot(`"${truePath}"`, `"${tentativePath}"`, this.workerContext.eventStream);
+                    }
+                }
+                else
+                {
+                    this.workerContext.eventStream.post(new DotnetFindPathNoRuntimesOnHost(`The host: ${tentativePath} does not contain a .NET runtime installation.`));
+
+                }
```
joeloff:
```suggestion
                    this.workerContext.eventStream.post(new DotnetFindPathNoRuntimesOnHost(`The host: ${tentativePath} does not contain a .NET runtime installation.`));
                }
```

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
```diff
@@ -164,14 +286,14 @@ Please set the PATH to a dotnet host that matches the architecture. An incorrect
                 mode: 'sdk',
                 version: parts[0],
                 directory: sdk.split(' ').slice(1).join(' ').slice(1, -1), // need to remove the brackets from the path [path],
-                architecture: architectureKnown ? requestedArchitecture : null
+                architecture: null
```
joeloff, line 288: What does it mean to have the `architecture` set to `null`? Wouldn't the SDK in this case be one that matches `requestedArchitecture`?

nagilson: Great question - `null` means we dont know what it is for sure, which is true, and a change to this function to fix a bug. That's not a guarantee and we don't want to check that here, for A - performance reasons (This needs to work with 'dotnet' but we need the full path to 'dotnet' to get the architecture from the binary file. We dont want to have to resolve 'which' or 'where' or try to resolve the full path every time we call this function. And, sometimes we want to just use 'dotnet' - sometimes we dont even need a 'requestedArchitecture' because the API caller doesn't care, and wants to use whichever architecture is on the PATH.) and B - to remove the responsibility of this from having to. 

requestedArchitecture is the architecture the API caller wanted. It may be null. The SDK on the path or in a user setting may be the same architecture of vscode, or whichever debugger architecture is needed, or it may not be. Just because we pass --arch (requestedArchitecture) to --list-sdks does not mean that it actually worked, if the dotnet host is too old to support --arch, then it could be for a different architecture. This is a private function now which I think makes it easier to have this caveat.

vscode-dotnet-runtime-library/src/Acquisition/DotnetResolver.ts
nagilson: A Validator will run on the path later to make sure it exists. That's a separate responsibility. path.dirname will return '.' if it's an invalid directory, so we don't have to worry about an exception here. All this does is resolve the path to the actual path of the file. Example: snap -> dotnet.exe instead of snap.exe on the path

PR #2401, opened 2025-08-18
Add PME Service for Signing

nagilson:
Please read https://dev.azure.com/devdiv/DevDiv/_wiki/wikis/DevDiv.wiki/46279/Real-signing-with-PME-Enforcement for more information about this new signing enforcement. The PME Service Name is fine to put in public, it is already public. We can't do manual builds anymore either, so this changes the default standards and makes it so odd number releases are pre-release and even number releases are not. (This is so a x.0.0 release is an official release and matches some of our other patterns.) Since the next release is a pre-release, I bumped the version to produce the correct build.

### DIFF
--- a/pipeline-templates/prepare-signing.yaml
+++ b/pipeline-templates/prepare-signing.yaml
@@ -15,7 +15,8 @@ steps:
       zipSources: false
       feedSource: https://dnceng.pkgs.visualstudio.com/_packaging/MicroBuildToolset/nuget/v3/index.json
       azureSubscription: 'MicroBuild Signing Task (DevDiv)'
-      ConnectedPMEServiceName: c24de2a5-cc7a-493d-95e4-8e5ff5cad2bc
+      ${{ if eq(variables['Build.Reason'], 'IndividualCI') }}:
+        ConnectedPMEServiceName: c24de2a5-cc7a-493d-95e4-8e5ff5cad2bc
     env:
       SignType: ${{ parameters.SignType }}
       TeamName: DotNetCore


nagilson:
Ignore the last commit, I was trying to verify the PME got used, but we're only allowed to use it in CI. It looks like its getting picked up but its blocked if it's not from CI, so I think this should work. I can't really test further it until it's merged.

MiYanni:
> Ignore the last commit, I was trying to verify the PME got used, but we're only allowed to use it in CI. It looks like its getting picked up but its blocked if it's not from CI, so I think this should work. I can't really test further it until it's merged.

To test, you need to make an internal branch and run the build with these changes. Usually when doing infra work, I don't fork. I just make a public branch and an internal branch, and keep them synced as I'm making changes.

pipeline-templates/prepare-signing.yaml
```diff
@@ -14,6 +14,9 @@ steps:
       signType: ${{ parameters.SignType }}
       zipSources: false
       feedSource: https://dnceng.pkgs.visualstudio.com/_packaging/MicroBuildToolset/nuget/v3/index.json
+      azureSubscription: 'MicroBuild Signing Task (DevDiv)'
+      ${{ if eq(variables['Build.Reason'], 'IndividualCI') }}:
+        ConnectedPMEServiceName: c24de2a5-cc7a-493d-95e4-8e5ff5cad2bc
```
MiYanni, line 19: Is this the only change related to PME stuff?

pipeline-templates/package-vsix.yaml
```diff
@@ -30,8 +29,19 @@ jobs:
...
+      # Calculate if this should be prerelease based on patch version (odd = prerelease)
+      PATCH_VERSION="${VERSION##*.}"  # Gets everything after the last dot
+      IS_ODD=$((PATCH_VERSION % 2))
+      if [ $IS_ODD -eq 1 ]; then
+        echo "##vso[task.setvariable variable=isPrerelease;isOutput=true]true"
+        echo "Version $VERSION has odd patch number ($PATCH_VERSION) - will be marked as prerelease"
+      else
+        echo "##vso[task.setvariable variable=isPrerelease;isOutput=true]false"
+        echo "Version $VERSION has even patch number ($PATCH_VERSION) - will be marked as stable release"
+      fi
```
MiYanni, line 42: ~~So is the cadence always, even = release and odd = prerelease? What happens if there's a critical issue and needs an immediate release? Do you just skip the odd number?~~

I answered my own question by looking lower in the changes and seeing you skipped 2.3.8 so you could put out another prerelease. Got it.

nagilson:
Im blocked from changing the default branch, but that's a good idea

pipeline-templates/prepare-signing.yaml
nagilson: In a way, yes. On the same day PME is enforced, they are also enforcing no manual builds, so this is making the manual stuff not manual. It's required to always real sign on the PME pipeline so that is required for this pr to work as well, so this code change alone is not sufficient, but it is the main change.

PR #2409, opened 2025-08-20
Don't include windows desktop runtimes in output

nagilson:
We don't support windows desktop runtimes as a mode so we shouldnt return those.


joeloff (approved):
Just to be clear, you now have a test to ensure that the desktop runtime is not being returned, but the logic was already there since you only checked for runtime/ASP.NET Core and otherwise return the SDK?

nagilson:
The logic was already there, except it was a regression because we did not filter out the windows desktop runtimes

PR #2410, opened 2025-08-22
Fix PME Signing 

nagilson:
This adds to the requirements of PME signing. 

With this PR, https://github.com/dotnet/vscode-dotnet-runtime/pull/2401, some things were missing/incorrect:
- `signWithProd` was not enabled
- azureSubscription is not needed.
- The wrong Service Connection was used (Unix instead of Win)
- Signature verification was not conditioned to be per only official builds
- The AzDo Env did not have permission for the Service Connection

This PR also adds a step which prints the file structure which makes it easier to see where the sign file drops are when they get ingested.

### DIFF
--- a/pipeline-templates/list-file-structure.yaml
+++ b/pipeline-templates/list-file-structure.yaml
@@ -16,7 +16,6 @@ steps:
         Write-Host "====================== FILE STRUCTURE OVERVIEW ======================"
         Write-Host ""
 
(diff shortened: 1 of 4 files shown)


joeloff (approved):

Mind-Vault-AI (approved):

Mind-Vault-AI:
Thx

PR #2411, opened 2025-08-22
Migrate to Deployment Model

nagilson:
Also includes changes from https://github.com/dotnet/vscode-dotnet-runtime/pull/2410 until force pushed, as those are needed to work properly.

The timed task will cause the pipeline to fail. It may have a 'resume' feature we can check to see if it actually failed or not, but there is no way to pause the next task and 'pass' that 'resume' onto the next step, because the timed task is using a 'serverless' pool.

This method is more complicated but a better e2e approach.
This commit removes the line which does the publish so I can test that the deployment 'gate' in the environment actually works.


nagilson:
Instead of trying to do all of this in an isolated deployment environment, we can separate the release pipeline into its own component, which is a more elegant approach. Thank you to @JoeRobich for the inspiration and assistance 🥳 

https://github.com/dotnet/vscode-dotnet-runtime/pull/2515

Issue #2359, opened 2025-07-21
Automatic Update / Management of Runtimes - Breaking 3.0.0 NET Install Tool Change


nagilson:
# Scenario Today

The `dotnet.acquire` API for downloading a local .NET runtime is designed in such a way that it:
- Always downloads the latest runtime of the specified `major.minor`
- Will use existing local installs it's already created if they exist
- Shares installs across extensions
- Always checks for a runtime update before doing anything, which requires pinging the releases.json manifest

The time to ping the releases.json manifest varies and is generally short but for the worst case of users it can take many seconds. (7 to 36 seconds for the percentile of 5% worst,  but a median delay of 280ms) - This runtime is needed to resolve any C#DK dotnet logic for code written in C# like the language server processes.

We have decided to pivot for the new version.

# Breaking Change for 3.0.0

The `dotnet.acquire` API will still download the runtime of the `major.minor`, but it will not check for updates every single time. The .NET Install Tool will periodically try to update and replace that runtime.
We will provide the option `forceUpdate : boolean = false` in the context API object call to revert to the prior behavior.

The update check will now run sometime (Say, 5 minutes) after VS Code startup to prevent slowdown of other essential components, unless the force mode is used. The check will run based upon the update schedule of .NET and within the confines of the proper SDL update requirements, which we have discussed with the security team. A message should be displayed if the user has not restarted code within some time to receive the update.

The older runtimes will be uninstalled if they are no longer in use to prevent deleting the runtime from under other extensions, and only upon the next time when a newer runtime is installed. Typically, this would happen upon the next session.

The `dotnet.findPath` API will become aware of local .NET Runtimes. Those runtimes will also become managed and automatically update. This feature will be disable-able in the lookup via a flag `disableLocalLookup : boolean = false` in the API call's context object. We do not plan on giving extensions an option to `pin` a version where we would not try to auto-update it. Callers would not need to change to rely on this behavior if they are calling `findPath` unless they always want the latest runtime no matter what.

This design decision I believe was an artifact of the prototype.

cc @JakeRadMSFT @claudiaregio @baronfel @blowdart @lifengl @dibarbet

dibarbet:
> The dotnet.findPath API will become aware of local .NET Runtimes. Those runtimes will also become managed and automatically update. This feature will be disable-able in the lookup via a flag disableLocalLookup : boolean = false in the API call's context object. 

Local .NET runtimes being ones from the global.json path?  Or local in some other way?

PR #2420, opened 2025-09-08
Automatic Update Management of Archive Based Runtime Installs

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2359
See breaking change doc: (https://github.com/dotnet/docs/issues/49127)

# Motivation

Please read the issue for motivation. SDK installs are not impacted as we only manage system-level installs for the SDK and those updates are managed natively by the OS.

# Testing

9, 10, and 11 in the test plan defined here onenote:https://microsoft.sharepoint.com/teams/ManagedLanguagesIDE/Shared%20Documents/CLI/CLI%20OneNote/Test%20Plans.one#VSCode%20Runtime%20Extension%20Test%20Plan&section-id={A8CA4C1A-9FF7-4495-9836-4D46E9A58081}&page-id={607935D1-2904-490A-ABC2-4B7E2E52159C}&end is an easy way to test this out yourself.

# Changes

- The `acquire` command now returns an existing installation matching a requested `major.minor` instead of always checking the `releases.json` first and downloading a new version.

- An update check runs and tries to update the installed runtimes once daily. The update check runs after 5 minutes by default and is configurable in the new setting: `runtimeUpdateDelaySeconds`. Users could already disable the install mechanism, so this is not a 'new' power they have.

- Each 'runtime' install is categorized into a group of 'alike' installs. We install the latest version in that 'group' and uninstall all of the other versions that are not in use within that group. A group is defined by a combination of: `mode (runtime/aspnetcoreruntime/wdesktopruntime)` + `arch` + `major.minor`.

- `dotnet.uninstall` won't allow uninstalling if there are 'live' dependent extensions of the install besides the dependent extension asking to commit to the uninstall action. Expected impact as minimal as I couldn't find any largely adopted code in GH that uses this besides us. For a definition of 'live' please see https://github.com/dotnet/docs/issues/49127

- Adds options `forceUpdate` for other extensions to force trigger an update of the .NET runtime. 
- Adds option `disableLocalLookup` to `dotnet.findPath` to not find local .NET runtimes and only find (presumably) system-managed or user-managed installs.

# Code Context

- Previously, we required extensions to uninstall installs themselves and we kept ref counts.

- We have for a year+ kept track of 'dependents' or extensions which requested a specific install. We now keep track of whether we've returned the path to be used in any 'live' session of code in addition to this info. This requires a mutex as it must work across vscode, vscode insiders, and multiple versions/processes of code. Because we don't have a hook to reliably clear data when vscode is closed, the extension holds onto a permanent session mutex (without dead-waiting) to represent that it is still alive and using the returned installs.

- We also check the PATH and DOTNET_ROOT to prevent uninstalling something that's in use. We don't delete dotnet if the `dotnet.exe` is in use either which is a good protection mechanism. However, some dependents may use a `dll` inside the runtime and not just `dotnet.exe`, so that is not a sufficient check.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,7 +46,6 @@ install.*
 .vscode-test/
 *.vsix
 vscode-dotnet-runtime-library/install scripts/dotnet-install.*
-!/vscode-dotnet-runtime-library/src/test/mocks/MockMutexHolder.js
 vscode-dotnet-runtime-extension/LICENSE.txt
 vscode-dotnet-sdk-extension/LICENSE.txt
 !vscode-dotnet-runtime-library/src/test/mocks/MockMutexHolder.js
--- a/vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
@@ -88,7 +88,7 @@ export class InstallTrackerSingleton
             {
                 InstallTrackerSingleton.sessionMutexReleaser = resolved;
             }
-        }).catch(() => {}); // Assumption : We can ignore these errors because install/uninstall won't work if mutexes are unacquireable
+        }).catch(() => {}); // Assumption : We can ignore these errors because install/uninstall won't work if mutexes are unacquirable
     }
 
     /**
@@ -361,7 +361,7 @@ export class InstallTrackerSingleton
             {
                 if (typeof install === 'string')
                 {
-                    this.eventStream.post(new ConvertingLegacyInstallRecord(`Converting legacy install record ${install} to a null owner.Assuming:
+                    this.eventStream.post(new ConvertingLegacyInstallRecord(`Converting legacy install record ${install} to a null owner. Assuming:
                         ${JSON.stringify(InstallToStrings(getAssumedInstallInfo(install, null)))}`));
                     convertedInstalls.push(
                         {


vscode-dotnet-runtime-library/src/Acquisition/LocalInstallUpdateService.ts
```diff
@@ -0,0 +1,213 @@
...
+        this.eventStream.post(new UpdatingInstallGroups(`Automatic update triggered and detected the following install groups: ${JSON.stringify(installGroupEntries)}`))
+        for (const [group, installsInGroup] of installGroupEntries)
+        {
+            const acquireContext: IDotnetAcquireContext = {
+                version: group.majorMinor,
+                architecture: group.architecture,
+                mode: group.mode,
+                installType: 'local',
+                requestingExtensionId: 'dotnet-runtime-library',
+                errorConfiguration: AcquireErrorConfiguration.DisableErrorPopups, // todo : make this quiet as well ?
```
nagilson:
```suggestion
                errorConfiguration: AcquireErrorConfiguration.DisableErrorPopups
```

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -84,11 +288,42 @@ export class InstallTrackerSingleton
...
+    public async addOwners(install: DotnetInstall, ownersToAdd: (string | null)[], dirProvider: IInstallationDirectoryProvider): Promise<void>
+    {
+        await executeWithLock(this.eventStream, false, this.getLockFilePathForKey(dirProvider, 'installed'), 5, 200000,
+            async () =>
+            {
+                const existingInstalls = await this.getExistingInstalls(dirProvider, true);
+                const idx = existingInstalls.findIndex(x => IsEquivalentInstallation(x.dotnetInstall, install));
+                if (idx !== -1)
+                {
+                    const installRecord: InstallRecord = existingInstalls[idx];
```
nagilson, line 330: It might be faster to not use a Set since this is very small and only in memory as a set for a small period of time. I'll remark on improving this code later after this PR is merged so its more manageable to review.

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -26,16 +33,101 @@ import
...
+import { InstallOwner, InstallRecord, InstallRecordOrStr } from './InstallRecord';
 
 export type InstallState = 'installing' | 'installed';
 
+export type SessionsWithInUseExecutables = Map<string, Set<string>>;
+
 export class InstallTrackerSingleton
 {
     protected static instance: InstallTrackerSingleton;
+    protected static sessionId: string;
```
nagilson: It seems like we would want to not have this be static, since this bypasses the constructor which instantiates the id.

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -64,14 +250,32 @@ export class InstallTrackerSingleton
...
                 const installWasMadeByUserAndHasNoExtensionDependencies = (allowUninstallUserOnlyInstall &&
-                    installRecord[0]?.installingExtensions?.length === 1 && installRecord[0]?.installingExtensions?.includes('user'));
+                    (installRecord[0]?.installingExtensions?.length ?? 0) === 1 && installRecord[0]?.installingExtensions?.includes('user'));
 
-                return zeroInstalledRecordsLeft || installedRecordsLeftButNoOwnersRemain || installWasMadeByUserAndHasNoExtensionDependencies;
+                return zeroRelevantRecordsLeft || installedRecordsLeftButNoOwnersRemain || installWasMadeByUserAndHasNoExtensionDependencies;
             }, 'installed', dotnetInstall);
     }
 
+    public async reportSuccessfulUninstall(context: IAcquisitionWorkerContext, installIdObj: DotnetInstall, forceUninstall = false, alreadyHoldingLock = false)
```
nagilson: I don't think the already holding lock bit should be publicly accessible.

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
nagilson: While this is true, it's protected so I don't see this causing any issue.

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -548,6 +589,11 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
         return uninstall(commandContext);
     });
 
+    const dotnetForceUpdateRegistration = vscode.commands.registerCommand(`${commandPrefix}.forceUpdate`, async (commandContext: IDotnetAcquireContext): Promise<void> =>
+    {
+        return automaticUpdater.ManageInstalls(0).catch((e: any) => {});
```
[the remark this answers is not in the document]
nagilson: I'd rather not surface these details to bug the user. We should be tracking if this happens on our own and not relying on users to report a bug. This isn't meant to be a fully fledged scanner tool so this is not needed.

vscode-dotnet-runtime-library/src/Acquisition/LocalInstallUpdateService.ts
```diff
@@ -0,0 +1,215 @@
...
+                const uninstallContext: IDotnetAcquireContext = {
+                    version: install.dotnetInstall.version,
+                    architecture: install.dotnetInstall.architecture,
+                    mode: install.dotnetInstall.installMode,
+                    requestingExtensionId: 'dotnet-runtime-library',
+                    installType: 'local',
+                    errorConfiguration: AcquireErrorConfiguration.DisableErrorPopups
+                };
+
+                this.uninstallAction(uninstallContext, false, true).catch((e: any) => {});
```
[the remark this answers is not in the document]
nagilson: We already have uninstall failure reporting.

drewnoakes (approved):
LGTM. This is a great improvement for people who are seeing outlier performance issues with C# Dev Kit.

.gitignore
```diff
@@ -46,6 +46,8 @@ install.*
 .vscode-test/
 *.vsix
 vscode-dotnet-runtime-library/install scripts/dotnet-install.*
+!/vscode-dotnet-runtime-library/src/test/mocks/MockMutexHolder.js
 vscode-dotnet-runtime-extension/LICENSE.txt
 vscode-dotnet-sdk-extension/LICENSE.txt
+!vscode-dotnet-runtime-library/src/test/mocks/MockMutexHolder.js
```
MiYanni: If I'm reading this right, the first is from root with that exact path being excluded. Then what is the second for?

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -185,11 +189,35 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
...
 
 
+    const runtimeUpdateDirectoryProvider = directoryProviderFactory(
+        'runtime', vsCodeContext.globalStoragePath); // Assumption : aspnetcore and runtime directory provider use the same logic, otherwise updates would not be found
+    const automaticUpdater = new LocalInstallUpdateService(globalEventStream, vsCodeContext.globalState, runtimeUpdateDirectoryProvider,
+        acquireLocal,
+        uninstall
+    );
+
+    if (!(process.env.DOTNET_INSTALL_TOOL_UNDER_TEST === 'true')) // Don't try to update while testing - this would make tests fail randomly
```
MiYanni, line 199: Is there any other situations that may apply where we wouldn't want to run the update?

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -900,20 +947,20 @@ ${JSON.stringify(commandContext)}`));
...
+            return Promise.resolve(existingOfflinePath);
+        }
+        else
+        {
+            if (!(await WebRequestWorkerSingleton.getInstance().isOnline(timeoutValue ?? defaultTimeoutValue, globalEventStream)))
             {
-                globalEventStream.post(new DotnetOfflineWarning(`It looks like you may be offline (can you connect to www.microsoft.com?) and have no installations of .NET for VS Code.
-We will try to install .NET, but are unlikely to be able to connect to the server. Installation will timeout in ${timeoutValue} seconds.`))
+                globalEventStream.post(new DotnetOfflineWarning(`It looks like you may be offline (can you connect to www.microsoft.com?) and have no compatible installations of .NET ${workerContext.acquisitionContext.version} for ${workerContext.acquisitionContext.requestingExtensionId ?? 'user'}.
+Installation will timeout in ${timeoutValue} seconds.`))
```
MiYanni, line 963: nit: Maybe be more specific with the MS URL `https://www.microsoft.com` or maybe just remove the question?

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -26,16 +39,103 @@ import
...
+        mutex.acquireWithManualRelease(InstallTrackerSingleton.sessionId, InstallTrackerSingleton.SESSION_MUTEX_PING_DURATION, InstallTrackerSingleton.SESSION_MUTEX_ACQUIRE_TIMEOUT_MS).catch((error) =>
+        {
+            this.eventStream.post(new SessionMutexAcquisitionFailed(`Failed to acquire permanent mutex for session ${InstallTrackerSingleton.sessionId}: ${error}`));
+        }).then((resolved) =>
+        {
+            if (resolved)
+            {
+                InstallTrackerSingleton.sessionMutexReleaser = resolved;
+            }
+        }).catch(() => {}); // Assumption : We can ignore these errors because install/uninstall won't work if mutexes are unacquireable
```
MiYanni: nit: `unacquirable`

vscode-dotnet-runtime-library/src/Acquisition/InstallTrackerSingleton.ts
```diff
@@ -105,8 +361,8 @@ export class InstallTrackerSingleton
             {
                 if (typeof install === 'string')
                 {
-                    this.eventStream.post(new ConvertingLegacyInstallRecord(`Converting legacy install record ${install} to a null owner. Assuming:
-                    ${JSON.stringify(InstallToStrings(getAssumedInstallInfo(install, null)))}`));
+                    this.eventStream.post(new ConvertingLegacyInstallRecord(`Converting legacy install record ${install} to a null owner.Assuming:
+                        ${JSON.stringify(InstallToStrings(getAssumedInstallInfo(install, null)))}`));
```
MiYanni, line 365: nit: Lost a space before 'Assuming'.

.gitignore
nagilson: Looks like this was from a bad merge - good catch!

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: That's a good question - there's nothing I can think of to my knowledge. Customers could set the update time to a very long time, though I suppose there isn't a way to turn it off completely besides telling them to use `DOTNET_INSTALL_TOOL_UNDER_TEST` 😁

Someone may say this is concerning, but env variables are generally considered to be trusted.

vscode-dotnet-runtime-extension/src/extension.ts
nagilson: The specific website is there because that's one website we check if the user is offline. If microsoft.com goes down and they file an issue saying they weren't offline, then it would be helpful to know 😁

nagilson:
Merging - the linter is failing to install due to a new auth token, which was the only issue.

Issue #2460, opened 2025-10-29
[NETE2ESDK] Automatically update .NET runtimes and ASP.NET Core Runtimes does not work with the .NET Install Tool


yinjiehe01:
### Describe the bug.

Install the previous .NET runtime and ASP.NET Core runtime and then change the 'runtimeUpdatedelaySeconds' to 20 seconds. It cannot update to the latest version after waiting 20 seconds.

### Please attach log file(s) here if you have any.

[logs.zip](https://github.com/user-attachments/files/23204032/logs.zip)

The log file location should be copied to your clipboard if you came from VS Code.

### Reproduction Steps

1.Intall 8.0.20 runtime and ASP.NET Core runtime on Windows, Mac.
2.Set 'runtimeUpdateDelaySeconds' to 20 seconds and then restart VS Code.
3.Wait for a while and observe the .NET Install Tool output window.

### Expected Behavior

The .NET Install Tool output window shows the .NET runtime and ASP.NET Core runtimes update to the latest version, and uninstall the previous version.

### Exceptions & Errors (if any)

There is nothing happened in the .NET Install Tool output window.

### .NET Install Tool Version

2.4.1

### Your operating system? Please provide more detail, such as the version and or distro above.

N/A

### VS Code Version

1.105.1

yinjiehe01:
Not repro on the new version, so closed it.

PR #2472, opened 2025-11-03
Automatic Update - Fit & Finish Improvements

nagilson:
https://github.com/dotnet/vscode-dotnet-runtime/pull/2420 was merged which enabled automatic update and cleanup of vscode-based .NET runtime installs.

The vendor tests reported it as not working: https://github.com/dotnet/vscode-dotnet-runtime/issues/2460
I realized the vendor tests didn't really work - the daily check for updates only runs once. The vendor tests require starting vscode for the day and installing older versions to verify if they get cleaned up. So, I had to add a command that allows a tester to reset the update time, so they can see if the automatic updates trigger.
This command is `Sample: Reset the update timer for .NET`

Their tests also did not have any logs. I realized there was a race condition with the logs, and that we weren't always outputting to the log. This code enforces the installation updater to push out to the log when it's done running update checks. It also prevents a collision from occurring - the log basically keeps track of a bunch of messages in memory and gets dumped at once to avoid waiting on file writes. But when we have multiple of those happening at the same time, the file can become corrupt or empty. This moves the file writes into a move operation which is synchronously completed when we've finished writing to the temp file. There is room for improvement in this system but this seems good for now.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/extension.ts
+++ b/vscode-dotnet-runtime-extension/src/extension.ts
@@ -445,7 +445,6 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
             const existingOfflinePath = await getExistingInstallOffline(worker, workerContext);
             if (existingOfflinePath)
             {
-                // make sure this is marked
                 return Promise.resolve(existingOfflinePath);
             }
 


vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -443,6 +445,7 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
             const existingOfflinePath = await getExistingInstallOffline(worker, workerContext);
             if (existingOfflinePath)
             {
+                // make sure this is marked
```
nagilson:
```suggestion

```

MiYanni (approved):

Issue #2476, opened 2025-11-07
[NETE2ESDK] An error occurred when using a single number version to install SDK.


yinjiehe01:
### Describe the bug.

When we install the SDK with a single number version, an error occureed.
<img width="415" height="301" alt="Image" src="https://github.com/user-attachments/assets/b077ff8d-93e5-4d05-997a-0ab6655573e5" />

### Please attach log file(s) here if you have any.

The log file location should be copied to your clipboard if you came from VS Code.

[logs.zip](https://github.com/user-attachments/files/23410632/logs.zip)

### Reproduction Steps

1.Run command `Sample: Install .NET SDK Globally via .NET Install Tool (Former Runtime Extension)` or `.NET Install Tool - Install the .NET SDK System-Wide` and then input SDK version number like 7, 8, 9.

### Expected Behavior

It will install the latest SDK version.

### Exceptions & Errors (if any)

The SDK cannot be installed as an error occurred.

### .NET Install Tool Version

2.4.1

### Your operating system? Please provide more detail, such as the version and or distro above.

N/A

### VS Code Version

1.150.1

PR #2497, opened 2025-11-25
Only Check for Existing Installs if Offline on SDK Admin Installs

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2476

2 core issues are resolved in this change:

1: 
https://github.com/dotnet/vscode-dotnet-runtime/pulls?q=is%3Apr+is%3Aclosed made it so that we skip the offline check before reusing existing installs. The runtime install is automated and we update it automatically so it's ok to use an older runtime version for the acquire API if forceUpdate is not true. For the SDK, we don't update automatically, so we shouldn't use the existing install of the same major.minor unless we are offline and can't do anything. You could debate whether we'd want to do that either, but we've leaned towards minimizing friction.

2:
The check to get a major minor failed if the version was not fully quantified. We should be able to handle that better.

3:
Updates dependencies. Normally this would be a separate PR, but there is a lot of OOF during the holidays and it will be easier to get this through with 1 PR.

Also improves test coverage.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/extension.ts
+++ b/vscode-dotnet-runtime-extension/src/extension.ts
@@ -449,7 +449,8 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
 
             globalEventStream.post(new DotnetAcquisitionStatusRequested(commandContext.version, commandContext.requestingExtensionId));
 
-            // Caveat : acquireStatus expects only a major.minor, so fully specified versions wont be checked here
+            // Caveat : acquireStatus expects only a major.minor, so fully specified versions won't be checked here
+
             const existingOfflinePath = await getExistingInstallOffline(worker, workerContext);
             if (existingOfflinePath)
             {


vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -449,6 +449,7 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
 
             globalEventStream.post(new DotnetAcquisitionStatusRequested(commandContext.version, commandContext.requestingExtensionId));
 
+            // Caveat : acquireStatus expects only a major.minor, so fully specified versions wont be checked her
```
nagilson:
```suggestion
            // Caveat : acquireStatus expects only a major.minor, so fully specified versions wont be checked here
```

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -449,6 +449,7 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
 
             globalEventStream.post(new DotnetAcquisitionStatusRequested(commandContext.version, commandContext.requestingExtensionId));
 
+            // Caveat : acquireStatus expects only a major.minor, so fully specified versions wont be checked here
```
joeloff:
```suggestion
            // Caveat : acquireStatus expects only a major.minor, so fully specified versions won't be checked here

```

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -325,7 +325,7 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
 
             globalEventStream.post(new DotnetAcquisitionRequested(commandContext.version, commandContext.requestingExtensionId ?? 'notProvided', commandContext.mode!, commandContext.installType ?? 'global'));
 
-            const existingOfflinePath = await getExistingInstallOffline(worker, workerContext);
+            const existingOfflinePath = await getExistingInstallIfOffline(worker, workerContext);
```
joeloff, line 328: Is the old method still being used? I noticed the code for `getExistingInstallOffline` still exists

nagilson: Yes, great question - `getExistingInstallOffline` is meant to be able to run to find existing installs without using the internet, but returns the installs regardless of if we are offline or not. `getExistingInstallIfOffline` is meant to do the same, but to only run the logic if we are offline.

nagilson: Thanks for taking a look 😊

joeloff (approved):

PR #2515, opened 2025-12-04
Utilize a separate pipeline for release tasks

nagilson:
Succeeds https://github.com/dotnet/vscode-dotnet-runtime/pull/2411

To migrate to a standardized deployment approach, we'd like to move away from the legacy Azure DevOps release mechanisms. The best way I've seen to do that is to have a separate release pipeline which handles only release activities - the release pipeline must be quite restricted which causes issues even trying to run our security scanners and compliance tools such as SBOM, so trying to create a deployment off our existing pipeline served challenging.

This approach is based off of the vscode-csharp release pipeline. https://github.com/dotnet/vscode-csharp/blob/main/azure-pipelines/release.yml

An example run is here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2858298&view=results

Other updates to make this work:
- Signing is not done on non-CI workflows by default, so we don't run into failures trying to pack when I run test pipeline runs in internal
- We print the pipeline workspace since it's the standard approach for staging deployment artifacts
- We pass the artifacts and package lock files to enable `npm ci` install of vsce. This is because we are blocked from installing updates in a ci machine so it would fail anytime a package sub dependency got an update (which happens nearly daily)
- updates get version logic and other logic to work in a new deployment scenario
- verifies PAT token 
- removes custom SBOM logic as SBOM no longer fails as it used to 

Thank you to @/joerobich who was very helpful in explaining the vscode-csharp-release structure.

### DIFF
--- a/release.yml
+++ b/release.yml
@@ -136,7 +136,7 @@ extends:
                     }
                     Else {
                       Write-Host "##[command]npx vsce $publishArgsString"
-
+                      npx vsce @publishArgs
                       if ($LASTEXITCODE -ne 0) {
                         throw "vsce publish failed with exit code $LASTEXITCODE."
                       }


release.yml
```diff
@@ -0,0 +1,92 @@
...
+                    If ("${{ parameters.SignType }}" -ne "Real") {
+                      Write-Host "With a test-signed build, the command to publish is printed instead of run."
+                      Write-Host "##[command]npx vsce $publishArgs"
+
+                      Write-Host "🔒 Verify PAT."
+                      npx vsce verify-pat --azure-credential ms-dotnettools
+                    }
+                    Else {
+                      Write-Host "##[command]npx vsce $publishArgs"
+
```
nagilson: We need to make this run the vsce step once we're sure that everything is working.

release.yml
```diff
@@ -0,0 +1,121 @@
+trigger: none
+pr: none
+
+parameters:
+  - name: test
+    type: boolean
+    default: true
```
MiYanni: nit: For consistency, keep list elements, `- {prop}`, aligned with the parent property. Thus,

```suggestion
parameters:
- name: test
  type: boolean
  default: true
```

You did this with `variables:` and `repositories:` and other stuff below. It still works if it is indented extra, but it just adds to confusion about YAML if they're all not the same.

release.yml
```diff
@@ -0,0 +1,121 @@
...
+            pipeline: officialBuildCI
+            artifactName: package_files
+            destinationPath: $(Pipeline.Workspace)/package_files
+        strategy:
+          runOnce:
+            deploy:
+              steps:
+              - template: pipeline-templates/install-node.yaml@self
+              - template: pipeline-templates/list-file-structure.yaml@self
+              - bash: |
```
MiYanni: `bash:` is a shorthand for this task: https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/bash-v3?view=azure-pipelines

Any reason this script needs to be bash on Windows? It then requires the script to run under WSL. Copilot could probably directly convert this to PowerShell pretty easily. Then, this entire job just uses PowerShell.

release.yml
nagilson: Great consideration, thanks.

release.yml
```diff
@@ -0,0 +1,126 @@
...
+              - pwsh: |
+                  $packageFilesDir = Join-Path '$(Pipeline.Workspace)' 'package_files'
+                  if (-not (Test-Path $packageFilesDir)) {
+                    Write-Error "package_files artifact was not downloaded."
+                    exit 1
+                  }
+
+                  Push-Location $packageFilesDir
+                  try {
+                    npm ci
```
jaredpar, line 101: If this command fails the pipeline will still succeed (pretty sure). The default failure behavior for powershell is to emit an error, not throw an exception. Also not sure that the `try / finally` is needed given the current settings.

release.yml
nagilson: I've converted this to pwsh. Thanks ;)

release.yml
nagilson: I think when it tries to expand `.version` it would fail, but that's a great point. I've improved this logic. With the new settings, I think the try catch makes more sense.

release.yml
nagilson: @MiYanni Are you good with the changes I made since last review + the feedback response? Thanks again for taking a look.

MiYanni (approved):

Issue #2519, opened 2025-12-04
RHEL Update Required to Remark on .NET 10 Support, '[Distro] does not officially support dotnet version 10.0.100'


nagilson:
We need to update the distro-json file to include .NET 10 support now that .NET 10 packages have been released to the public for Linux. The installation works but it doesn't think that it is supported.

PR #2520, opened 2025-12-04
Add support for dotnet 10 and newer distro versions

nagilson:
This is our version support system which powers our warnings for when users try to install out of support versions or not yet released versions for linux distros.

Ideally, this would be automated based on the documentation, though this likely won't be bumped in priority.

To try to minimize impact in the future, I've added the future versions of distros.

Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2519


vscode-dotnet-runtime-library/distro-data/distro-support.json
```diff
@@ -935,6 +1078,12 @@
             },
             {
                 "version": "9.0"
+            },
+            {
+                "version": "10.0"
+            },
+            {
+                "version": "11.0"
```
joeloff, line 1086: 11.0?

nagilson: This is for RHEL 11, whenever that happens! Just like how I added 26.04 for ubuntu.

richlander: FYI: We already have 26.04 in runtime CI.

PR #2550, opened 2026-01-26
Add SDK version utilities and GlobalJson parsing

phenning:
Add new utilities for SDK version parsing and comparison:
- parseSdkVersion: Parse SDK version string into components
- isCompatibleSdkVersion: Check version compatibility with rollForward policy
- getCompatibleSdkVersions: Filter versions by compatibility
- compareSdkVersions: Compare two SDK versions
- isNewerSdkVersion: Check if one version is newer than another
- isValidNumber: Exported helper for number validation

Add GlobalJson parsing utilities:
- GlobalJson, GlobalJsonSdk, GlobalJsonRequirements interfaces
- parseGlobalJsonContent: Parse global.json content
- getRequirementsFromGlobalJson: Extract SDK requirements
- parseGlobalJsonRequirements: Combined parse and extract
- getEffectiveRollForward: Get effective roll-forward policy
- getEffectiveAllowPrerelease: Get effective prerelease setting

Register VS Code commands for external extension consumption:
- dotnet.parseSdkVersion
- dotnet.isCompatibleSdkVersion
- dotnet.getCompatibleSdkVersions
- dotnet.compareSdkVersions
- dotnet.isNewerSdkVersion
- dotnet.parseGlobalJson

Add unit tests for all new SDK version utilities.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/extension.ts
+++ b/vscode-dotnet-runtime-extension/src/extension.ts
@@ -13,6 +13,7 @@ import
     AcquisitionInvoker,
     callWithErrorHandling,
     CommandExecutor,
+    compareSdkVersions,
     directoryProviderFactory,
     DotnetAcquisitionMissingLinuxDependencies,
     DotnetAcquisitionRequested,
@@ -43,11 +44,15 @@ import
     ExistingPathResolver,
     ExtensionConfigurationWorker,
     formatIssueUrl,
+    getCompatibleSdkVersions,
     getInstallIdCustomArchitecture,
     getMajor,
     getMajorMinor,
+    getRequirementsFromGlobalJson,
     GlobalAcquisitionContextMenuOpened,
     GlobalInstallerResolver,
+    GlobalJson,
+    GlobalJsonRequirements,
     IAcquisitionWorkerContext,
     IDotnetAcquireContext,
     IDotnetAcquireResult,
@@ -67,13 +72,19 @@ import
     InstallationValidator,
     InstallRecord,
     InvalidUninstallRequest,
+    isCompatibleSdkVersion,
+    isNewerSdkVersion,
     IUtilityContext,
     JsonInstaller,
     LinuxVersionResolver,
     LocalInstallUpdateService,
     LocalMemoryCacheSingleton,
     NoExtensionIdProvided,
+    ParsedSdkVersion,
+    parseGlobalJsonContent,
+    parseSdkVersion,
     registerEventStream,
+    RollForwardPolicy,
     UninstallErrorConfiguration,
     UserManualInstallFailure,
     UserManualInstallRequested,
@@ -83,7 +94,7 @@ import
     VSCodeEnvironment,
     VSCodeExtensionContext,
     WebRequestWorkerSingleton,
-    WindowDisplayWorker
+    WindowDisplayWorker,
 } from 'vscode-dotnet-runtime-library';
 import { InstallTrackerSingleton } from 'vscode-dotnet-runtime-library/dist/Acquisition/InstallTrackerSingleton';
 import { dotnetCoreAcquisitionExtensionId } from './DotnetCoreAcquisitionId';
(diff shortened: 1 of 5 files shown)


vscode-dotnet-runtime-library/src/Acquisition/GlobalJson.ts
```diff
@@ -0,0 +1,147 @@
...
+ * @returns The effective roll-forward policy.
+ */
+export function getEffectiveRollForward(requirements: GlobalJsonRequirements): RollForwardPolicy
+{
+    if (requirements.rollForward)
+    {
+        return requirements.rollForward;
+    }
+    // Default is 'latestPatch' when version is specified, 'latestMajor' otherwise
+    return requirements.sdkVersion ? 'latestPatch' : 'latestMajor';
```
nagilson, line 130: Should we support more than latestPatch and latestMajor?

nagilson, line 130: I'm wondering if we should store these as static constant strings instead of having them coded here.

vscode-dotnet-runtime-library/src/test/unit/VersionUtilities.test.ts
```diff
@@ -146,4 +146,137 @@ suite('Version Utilities Unit Tests', function ()
         assert.equal(resolver.isValidLongFormVersionFormat(badSDKVersionPeriods, mockEventStream, mockCtx), false, 'It detects a version with a bad number of periods');
     });
 
+    test('parseSdkVersion parses SDK versions correctly', async () =>
+    {
+        const v1 = resolver.parseSdkVersion('8.0.308');
+        assert.equal(v1.major, 8);
```
nagilson, line 152: I like the conciseness of the tests here!

vscode-dotnet-runtime-library/src/Acquisition/VersionUtilities.ts
```diff
@@ -19,6 +19,180 @@ import { BAD_VERSION } from './StringConstants';
...
+ * Check if an installed SDK version is compatible with a required version based on the rollForward policy.
+ * This implements the global.json rollForward behavior as documented at:
+ * https://learn.microsoft.com/en-us/dotnet/core/tools/global-json
+ *
+ * @param installedVersion The installed SDK version string
+ * @param requiredVersion The required SDK version string (from global.json)
+ * @param rollForward The rollForward policy (defaults to 'latestPatch' if not specified)
+ * @returns true if the installed SDK satisfies the requirement
+ */
+export function isCompatibleSdkVersion(installedVersion: string, requiredVersion: string, rollForward: RollForwardPolicy = 'latestPatch'): boolean
```
nagilson, line 79: `stringVersionMeetsRequirement` already exists as a method to compare strings against rollforward values. Perhaps it should be moved here and utilized as such.

vscode-dotnet-runtime-library/src/Acquisition/VersionUtilities.ts
```diff
@@ -19,6 +19,180 @@ import { BAD_VERSION } from './StringConstants';
...
+    patchFull: number;
+    isPrerelease: boolean;
+    originalVersion: string;
+}
+
+/**
+ * The rollForward policy from global.json that determines SDK version selection behavior.
+ * Reference: https://learn.microsoft.com/en-us/dotnet/core/tools/global-json
+ */
+export type RollForwardPolicy = 'disable' | 'patch' | 'latestPatch' | 'feature' | 'latestFeature' | 'minor' | 'latestMinor' | 'major' | 'latestMajor';
```
nagilson, line 43: I like the docstring here and the name makes more sense to me, but these exist to some degree already in `simplifiedVersionSpec` - we should consolidate them.

vscode-dotnet-runtime-library/src/Acquisition/VersionUtilities.ts
```diff
@@ -19,6 +19,180 @@ import { BAD_VERSION } from './StringConstants';
...
+
+/**
+ * Parse SDK version string into its component parts.
+ * Handles versions like "8.0.308", "9.0.100-preview.1", etc.
+ * Uses the pure utility functions for consistency with other version parsing in this file.
+ *
+ * @param version The SDK version string
+ * @returns Parsed version components
+ */
+export function parseSdkVersion(version: string): ParsedSdkVersion
```
nagilson, line 53: I think this is a clean approach to have a class like this but it is not the most performant approach since it parses many parts multiple times. The logic that compares versions already in the code short-circuits the evaluations of the string and prevents duplicated splits.

vscode-dotnet-runtime-library/src/Acquisition/VersionUtilities.ts
```diff
@@ -260,19 +435,79 @@ export function isFullySpecifiedVersion(version: string, eventStream: IEventStre
...
+function isPreviewVersionSimple(version: string): boolean
+{
+    return version.includes('-');
+}
+
+/**
+ * Gets the full patch field from an SDK version string (e.g., 308 from "8.0.308").
+ * @returns The full patch number, or undefined if it cannot be extracted
+ */
+function getFullPatchFromVersionSimple(version: string): number | undefined
```
nagilson, line 470: This seems similar to `getSDKFeatureBandOrPatchFromFullySpecifiedVersion` (albeit a less verbose name) - I'm  thinking many of these functions are not necessary.

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -841,6 +858,34 @@ ${JSON.stringify(commandContext)}`));
         open(url).catch(() => {});
     });
 
+    const compareSdkVersionsRegistration = vscode.commands.registerCommand(`${commandPrefix}.${commandKeys.compareSdkVersions}`,
```
nagilson, line 861: The most important feedback I have here:

I think it's wise for this logic to live here, and we shouldn't reimplement what's done in `recommendedVersion` and `listVersions`. Although I think the overhead of the vscode command calls is quite high. The fact that the codebase parses versions as strings instead of a type is definitely a regret and hard to change from the initial design, which is why we haven't replaced it with something like semver. For new code, this makes me think the version parsing/compare could live in CDK instead if there was a smarter type around it, for the sake of not repeating a bad pattern.

If we continue with this, really the best approach would be for some of the version utilities ( and our types) to be put on npm as a library dependency, considering there's no shared caching/persistent data. I have avoided doing that as there's never been a big drive to do so - (although it hurts seeing others have to copy our type files instead of using @types/.)

I understand doing so would:
A - create a new work item 
B - create a new maintenance point (which we could hopefully automate the release of alongside new extension releases)
C - delay the progress of your feature 

I'm curious on your opinion w.r.t this - it may be best to have it live in CDK for now.

nagilson: https://github.com/dotnet/vscode-dotnet-runtime/pull/2556 for types exporting

PR #2573, opened 2026-02-18
WIP - Add opt-in global runtime and ASP.NET Core installation

JakeRadMSFT:
## Why

C# Dev Kit needs an explicit way on Windows to install a .NET runtime or ASP.NET Core runtime system-wide instead of creating a private extension-managed installation. The Install Tool currently exposes system-wide acquisition only for SDKs.

This change adds an opt-in, Windows-only API for that caller. It does not change `dotnet.acquire`, startup acquisition, JSON acquisition, SDK acquisition, or any existing default installation behavior. Local runtime installation remains the safer general-purpose path; this preview API is intentionally limited to the identified vendor scenario.

## What

- Adds `dotnet.acquireGlobalRuntime` for `runtime` and `aspnetcore` modes on Windows.
- Rejects the new command on non-Windows platforms before resolving or installing anything.
- Leaves the existing Linux and global-SDK interfaces, lock construction, path lookup, and install behavior intact; product-neutral worker calls use compatibility wrappers.
- Resolves Windows runtime and ASP.NET Core installers without applying SDK feature-band rules.
- Preserves the existing SDK registry/feature-band conflict comparison and applies product-filtered runtime version comparison only to runtime/ASP.NET Core.
- Verifies an installed global product by exact framework name and version using the expected global host rather than ambient PATH.
- Publishes product-specific modal lifecycle events for global SDK, runtime, and ASP.NET Core acquisition.
- Updates the sample extension with vendor-facing commands for both global runtime modes.
- Adds direct Install Tool E2E coverage for global runtime and ASP.NET Core acquisition on Windows, plus non-Windows rejection coverage.
- Leaves PATH updates to the system installer for runtime/ASP.NET Core and preserves existing global SDK PATH behavior.

## API example

```typescript
const result = await vscode.commands.executeCommand('dotnet.acquireGlobalRuntime', {
    version: '9.0.0',
    requestingExtensionId: 'ms-dotnettools.csdevkit',
    mode: 'runtime', // or 'aspnetcore'
    installType: 'global',
    rethrowError: true,
});
```

## Validation

Exact head: `6a9c9d32309ac542099e9867ba77146ffca33778`

- Runtime library compilation passes.
- Runtime extension compilation passes.
- 47 focused library tests pass across acquisition-worker, installer-resolver, modal-event, and Windows installer behavior.
- 3 direct Windows Install Tool functional tests pass for global runtime, global ASP.NET Core, and invalid SDK-mode rejection using the fake global-install boundary; both runtime products verify no PATH mutation.
- The convergence commit removes unrelated macOS test-runner/cache-test changes and restores the corresponding baseline files byte-for-byte.
- No real MSI/system installation was run on the development host.

Existing Windows UI screenshots are pinned to immutable evidence commit `c50e82c8`. They demonstrate sample-command invocation and existing-install detection only; they are neither exact-head evidence nor proof of a fresh MSI installation. Exact-head fresh-install evidence on a disposable Windows machine remains required before this PR is ready for final review.


vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -361,6 +362,67 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
...
+
+            commandContext.version = fullyResolvedVersion;
+            telemetryObserver?.setAcquisitionContext(workerContext, commandContext);
+
+            outputChannelObserver.showOutput();
+            const dotnetPath = mode === 'aspnetcore'
+                ? await worker.acquireGlobalASPNET(workerContext, globalInstallerResolver)
+                : await worker.acquireGlobalRuntime(workerContext, globalInstallerResolver);
+
+            new CommandExecutor(workerContext, utilContext).setPathEnvVar(dotnetPath.dotnetPath, moreInfoUrl, displayWorker, vsCodeExtensionContext, true);
```
nagilson: setPathEnvVar should usually be handled by the admin installer itself

nagilson (commented):
Let me know if you want a review on this! But we'd need to talk because I dont think people want this to be the default

vscode-dotnet-runtime-extension/src/extension.ts
JakeRadMSFT: 🤖 Removed this call. Global runtime and ASP.NET acquisition now leave PATH management to the system installer. Addressed in `bd91c1ac`.

*Cowritten with Copilot*

JakeRadMSFT:
🤖 Review welcome. This remains an explicit programmatic API: it does not change `dotnet.acquire`, startup acquisition, JSON acquisition, or any existing default. I updated the title and description to make that scope clear.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/GenericDistroSDKProvider.ts
```diff
@@ -163,10 +163,10 @@ export class GenericDistroSDKProvider extends IDistroDotnetSDKProvider
 
     public async getDotnetVersionSupportStatus(fullySpecifiedVersion: string, installType: DotnetInstallMode): Promise<DotnetDistroSupportStatus>
     {
-        if (versionUtils.getFeatureBandFromVersion(fullySpecifiedVersion, this.context.eventStream, this.context) !== '1' ||
+        if ((installType === 'sdk' && versionUtils.getFeatureBandFromVersion(fullySpecifiedVersion, this.context.eventStream, this.context) !== '1') ||
             Number(versionUtils.getMajor(fullySpecifiedVersion, this.context.eventStream, this.context)) < 6)
         {
-            this.context.eventStream.post(new DistroSupport(`Distro: Dotnet Version ${fullySpecifiedVersion} is not supported by this extension. It has a non 1 level band or < 6.0.`));
+            this.context.eventStream.post(new DistroSupport(`Distro: Dotnet ${installType} version ${fullySpecifiedVersion} is not supported by this extension.`));
```
nagilson: I like the inclusion of the installType but this message is still only scoped to SDKs so I don't know why the extra context was removed, I would re add it.

vscode-dotnet-runtime-library/src/test/unit/DotnetCoreAcquisitionWorker.test.ts
```diff
@@ -259,6 +259,42 @@ suite('DotnetCoreAcquisitionWorker Unit Tests', function ()
         await acquireWithVersion('1.0', 'aspnetcore');
     }).timeout(expectedTimeoutTime);
 
+    test('Global runtime detection uses the runtime listing command', () =>
```
nagilson: I would also look into `runGlobalSdkInstallTest` and adding e2e unit tests for global runtime and aspnet runtime installation.

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -356,11 +368,32 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
         return null;
     }
 
-    private async sdkIsFound(context: IAcquisitionWorkerContext, version: string): Promise<boolean>
+    private getInstalledVersionsCommand(context: IAcquisitionWorkerContext)
```
nagilson: Instead of this, I would recommend using `getDotnetInstalls` from the `DotnetResolver`.

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -138,12 +142,17 @@ This report should be made at https://github.com/dotnet/vscode-dotnet-runtime/is
...
+    }
+
+    public override async installGlobal(installation: DotnetInstall): Promise<string>
     {
         return executeWithLock(this.acquisitionContext.eventStream, false, GLOBAL_INSTALL_STATE_MODIFIER_LOCK(this.acquisitionContext.installDirectoryProvider, installation), GLOBAL_LOCK_PING_DURATION_MS, this.acquisitionContext.timeoutSeconds * 1000,
             async (install: DotnetInstall) =>
             {
-                // Check for conflicting windows installs
-                if (os.platform() === 'win32')
+                // Check for conflicting windows installs (SDK-specific)
```
nagilson: I'm not sure if this is true, I think the installer will also block a runtime downgrade or conflicting version. I would verify this and test it e2e. `GlobalWindowsInstallWithConflictingVersionAlreadyExists` could probably instead conditionalize the feature band check.

vscode-dotnet-runtime-library/src/Acquisition/IGlobalInstaller.ts
```diff
@@ -26,9 +26,37 @@ export abstract class IGlobalInstaller {
 
     public abstract getExpectedGlobalSDKPath(specificSDKVersionInstalled : string, installedArch : string, macPathShouldExist? : boolean) : Promise<string>
 
+    /**
+     * Install a .NET product globally (SDK, runtime, or ASP.NET Core runtime).
+     * Default implementation delegates to installSDK for backward compatibility.
+     */
+    public async installGlobal(install : DotnetInstall) : Promise<string>
```
nagilson: There was never such a method for 'installGlobal' so I don't believe this is needed for backwards compatibility. I would instead suggest InstallGlobal abstract and simply conditionalize on the mode of the install in  its implementation, while removing the installSdk/uninstallSdk terminology. I would do the same for 'uninstall global'.

vscode-dotnet-runtime-library/src/Acquisition/LinuxVersionResolver.ts
```diff
@@ -248,11 +248,16 @@ If you experience issues, please reach out on https://github.com/dotnet/vscode-d
      * A string is returned in case we want to make this return more info about the update.
      * @remarks it is expected you are holding the global modifier lock when calling this function.
      */
-    private async UpdateOrRejectIfVersionRequestDoesNotRequireInstall(fullySpecifiedDotnetVersion: string, existingInstall: string | null): Promise<string>
+    private async UpdateOrRejectIfVersionRequestDoesNotRequireInstall(fullySpecifiedDotnetVersion: string, existingInstall: string | null, mode: DotnetInstallMode): Promise<string>
     {
         await this.Initialize();
 
         this.workerContext.eventStream.post(new DotnetInstallLinuxChecks(`Checking to see if we should install, update, or cancel...`));
+        if (mode !== 'sdk')
```
nagilson: I don't think we can ignore this for runtime installs, but if you want to defer it new issue + another PR and own updating it I'm ok to move forward on this.

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -432,19 +465,18 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
...
-            new WinMacGlobalInstaller(context, this.utilityContext, installingVersion, await globalInstallerResolver.getInstallerUrl(), await globalInstallerResolver.getInstallerHash());
+            new LinuxGlobalInstaller(context, this.utilityContext, installingVersion, install.installMode) :
+            new WinMacGlobalInstaller(context, this.utilityContext, installingVersion, await globalInstallerResolver.getInstallerUrl(), await globalInstallerResolver.getInstallerHash(),
+                null, null, install.installMode);
 
         // See if we should return a fake path instead of running the install
         if (process.env.VSCODE_DOTNET_GLOBAL_INSTALL_FAKE_PATH && process.env.VSCODE_DOTNET_GLOBAL_INSTALL_FAKE_PATH === 'true')
         {
             context.eventStream.post(new DotnetFakeSDKEnvironmentVariableTriggered(`VSCODE_DOTNET_GLOBAL_INSTALL_FAKE_PATH has been set.`));
-            // Return a realistic executable file path (directory + host executable) so callers that derive the
```
nagilson, line 442: nit: why  was this removed?

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -376,6 +377,66 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
         return pathResult;
     });
 
+    const dotnetAcquireGlobalRuntimeRegistration = vscode.commands.registerCommand(`${commandPrefix}.${commandKeys.acquireGlobalRuntime}`, async (commandContext: IDotnetAcquireContext): Promise<IDotnetAcquireResult | undefined> =>
```
nagilson, line 392: The sample extension must also be updated and have a test added that tests this command so that vendors can validate its functionality.

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -376,6 +377,66 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
         return pathResult;
     });
 
+    const dotnetAcquireGlobalRuntimeRegistration = vscode.commands.registerCommand(`${commandPrefix}.${commandKeys.acquireGlobalRuntime}`, async (commandContext: IDotnetAcquireContext): Promise<IDotnetAcquireResult | undefined> =>
+    {
+        const mode: DotnetInstallMode = commandContext.mode === 'aspnetcore' ? 'aspnetcore' : 'runtime';
+
+        if (commandContext.requestingExtensionId === undefined)
```
nagilson: The code from this point forward is relatively the same as `dotnetAcquireGlobalSDKRegistration` so I would instead do this:

commit 1. copy and paste the dotnetAcquireGlobalSDKRegistration implementation outside of the mode selection into its own inner function without changing anything 
commit 2 . delete the code in  dotnetAcquireGlobalSDKRegistration and call the helper function 
commit 3. modify the helper function to conditionalize the `worker.acquireGlobalSDK` etc part and any other needed changes  
commit 4. make this method also use the helper function.

Documentation/commands.md
```diff
@@ -34,6 +34,14 @@ The `version` accepts multiple formats:
 
 **Offline behavior:** If the machine is offline, the extension will attempt to find a compatible existing installation that it previously managed. In practice this rarely helps for global SDK installs, since externally installed SDKs are not tracked by the extension. If no compatible managed install is found while offline, a warning is posted and the install will time out.
 
+### dotnet.acquireGlobalRuntime
+
+This command installs a .NET runtime in a system-level location. It accepts an [IDotnetAcquireContext](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/IDotnetAcquireContext.ts) object and returns an [IDotnetAcquireResult](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/IDotnetAcquireResult.ts).
```
nagilson:
```suggestion
This *preview* command installs a .NET runtime in a system-level location. It accepts an [IDotnetAcquireContext](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/IDotnetAcquireContext.ts) object and returns an [IDotnetAcquireResult](https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/IDotnetAcquireResult.ts).

Note that limitations that apply to the global SDK command, such as the minimally supported Distros (Ubuntu/RHEL), also apply to the global runtime install command.
```

nagilson (commented):
Thank you for taking this on while I was gone.

I have left some feedback. For the validation, it would be very helpful to have requested e2e test as well as a screenshot showing the sample extension is able to correctly invoke the runtime command and then successfully install the runtime, on at least:
Windows, Mac, Ubuntu. 

That can make the review process easier/faster 😉 

We wouldn't have a validateable build until the PR is merged, and vendor testing often takes between 1-3 days, so to meet the release timeframe you mentioned we'd need you to advocate for the prioritization of testing this.

I tried to review it being a bit loose so it can move forward, but also still have some amount of rigor, hence the *preview* emphasis. I want to acknowledge that the local runtime installation is more reliable and reduces friction over some of the challenges with system level modification so I wouldn't recommend this pattern outside of what C#DK is aiming to do but that's adjacent to the PR review.

JakeRadMSFT:
## Windows sample-extension UI validation  Validated commit `c50e82c8` in a Stable VS Code Extension Development Host loading both the PR Install Tool extension and the updated sample extension.  ### Commands exposed by the sample  Sample global runtime commands  ### Version prompt  Global runtime version prompt  ### Runtime and ASP.NET Core results  Global runtime acquisition results  Observed results:  - `sample.dotnet.acquireGlobalRuntime` requested `.NET Runtime 7.0.20` and resolved `C:\Program Files\dotnet\dotnet.exe`. - `sample.dotnet.acquireGlobalASPNETRuntime` requested `ASP.NET Core Runtime 10.0.3` and resolved the same global host. - 50 focused library tests passed. - Both sample extension-host E2E tests passed using the fake global-install boundary.  These Windows screenshots validate the real VS Code command/API path and existing-install detection. Both products were already installed, so this is not evidence of a fresh MSI installation. Fresh-install screenshots for Windows, macOS, and Ubuntu remain outstanding.  *Cowritten with Copilot*

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -376,6 +377,66 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
...
+            architecture: commandContext.architecture ?? DotnetCoreAcquisitionWorker.defaultArchitecture()
+        } as DotnetInstall;
+
+        if (pathResult !== undefined && pathResult?.dotnetPath)
+        {
+            globalEventStream.post(new DotnetAcquisitionTotalSuccessEvent(commandContext.version, install, commandContext.requestingExtensionId ?? '', pathResult.dotnetPath));
+        }
+
+        loggingObserver.dispose();
+        return pathResult;
```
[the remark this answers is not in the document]
JakeRadMSFT: Fixed in `132bd585`. The shared `acquireGlobal` helper now awaits `loggingObserver.flush()` and never disposes the shared observer, so later commands retain logging.

*Cowritten with Copilot*

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -376,6 +377,66 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
...
+        const mode: DotnetInstallMode = commandContext.mode === 'aspnetcore' ? 'aspnetcore' : 'runtime';
+
+        if (commandContext.requestingExtensionId === undefined)
+        {
+            return Promise.reject(new Error('No requesting extension id was provided.'));
+        }
+
+        let fullyResolvedVersion = '';
+        const workerContext = getAcquisitionWorkerContext(mode, commandContext);
+        const worker = getAcquisitionWorker();
```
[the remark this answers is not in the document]
JakeRadMSFT: Fixed in `132bd585`. `acquireGlobal` assigns `commandContext.mode = mode` before creating the worker context, so all downstream resolver/detection paths receive the normalized mode.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/GenericDistroSDKProvider.ts
JakeRadMSFT: Restored the install-specific context in `132bd585`: SDK failures again explain the non-1 feature-band or < 6 condition, while runtime products report only the applicable < 6 condition.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/test/unit/DotnetCoreAcquisitionWorker.test.ts
JakeRadMSFT: Added vendor-facing end-to-end coverage in `132bd585`. The sample extension exposes runtime and ASP.NET Core global-acquisition commands, and the extension-host tests invoke both through the fake global-install boundary.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
JakeRadMSFT: Updated in `132bd585`. Installed global products are now represented by structured `DotnetResolver.getDotnetInstalls(...)` results and matched by mode/version instead of parsing separate lists.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
JakeRadMSFT: Verified and changed in `132bd585`. Runtime and ASP.NET Core installs now reject same-major/minor downgrades and exact conflicts, while the feature-band condition is applied only to SDKs. `WinMacGlobalInstaller.test.ts` covers runtime and ASP.NET downgrade, upgrade, and side-by-side major/minor cases.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/IGlobalInstaller.ts
JakeRadMSFT: Refactored in `132bd585` as suggested. `IGlobalInstaller` now exposes neutral `installGlobal`, `uninstallGlobal`, and `getExpectedGlobalDotnetPath` methods; implementations conditionalize on the install mode.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/LinuxVersionResolver.ts
JakeRadMSFT: Addressed in `132bd585` and `264c8a0e`. Existing runtime and ASP.NET Core packages now use the package-manager update/repair path, with focused tests for both modes. Nonzero install exit codes are also propagated rather than converted to success.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
JakeRadMSFT: This was only a wording cleanup, not a behavior removal. The current comment still documents that the fake path includes a realistic host executable so callers can derive its directory as they would for a real install.

*Cowritten with Copilot*

vscode-dotnet-runtime-extension/src/extension.ts
JakeRadMSFT: Added in `132bd585`. The sample now has `sample.dotnet.acquireGlobalRuntime` and `sample.dotnet.acquireGlobalASPNETRuntime`, and two extension-host E2E tests invoke those commands. Windows UI screenshots are linked in the PR conversation.

*Cowritten with Copilot*

vscode-dotnet-runtime-extension/src/extension.ts
JakeRadMSFT: Refactored in `132bd585`. Both registrations now delegate to one `acquireGlobal` helper; only mode selection remains in the public command registrations. The helper owns resolution, acquisition, PATH handling for SDK only, telemetry, and log flushing.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/LinuxVersionResolver.ts
```diff
@@ -299,29 +304,35 @@ If you experience issues, please reach out on https://github.com/dotnet/vscode-d
...
         await this.VerifyNoCustomInstallExists(supportStatus, fullySpecifiedDotnetVersion, existingInstall);
 
         // Check if we need to install or not, if we can install (if the version conflicts with an existing one), or if we can just update the existing install.
-        const updateOrRejectState = await this.UpdateOrRejectIfVersionRequestDoesNotRequireInstall(fullySpecifiedDotnetVersion, existingInstall);
+        const updateOrRejectState = await this.UpdateOrRejectIfVersionRequestDoesNotRequireInstall(fullySpecifiedDotnetVersion, existingInstall, mode);
         if (updateOrRejectState === '0')
         {
-            return await this.distroSDKProvider!.installDotnet(fullySpecifiedDotnetVersion, 'sdk') ? '0' : '1';
+            return await this.distroSDKProvider!.installDotnet(fullySpecifiedDotnetVersion, mode) ? '0' : '1';
         }
```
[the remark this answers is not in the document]
JakeRadMSFT: Fixed in `264c8a0e`. The code now compares the string result explicitly with `'0'`. A platform-independent regression test supplies `'1'` and verifies it is returned as failure; the focused acquisition suite is green with 69 passing tests.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -356,24 +370,28 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
...
+            return install.mode === mode && (install.version === version ||
+                (mode === 'sdk' && allowSdkMajorMinorMatch && install.version.startsWith(`${version}.`)));
+        });
+    }
+
+    private async dotnetInstallIsFound(context: IAcquisitionWorkerContext, version: string): Promise<boolean>
+    {
+        const mode = context.acquisitionContext.mode ?? 'runtime';
+        const installedDotnets = await new DotnetResolver(context, this.utilityContext).getDotnetInstalls(
+            this.dotnetExecutable, mode, context.acquisitionContext.architecture);
```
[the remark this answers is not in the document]
JakeRadMSFT: Fixed in `f22481b5`. `getValidExistingInstallPath` now passes the tracked global `dotnetPath` into `dotnetInstallIsFound`, and `DotnetResolver.getDotnetInstalls(...)` enumerates that exact host instead of whichever `dotnet` is on PATH. The regression test first failed with `dotnet.exe` vs `global-dotnet\dotnet.exe`; after the fix, 21 worker tests and the 70-test focused acquisition suite pass.`n`n*Cowritten with Copilot*

JakeRadMSFT:
@nagilson The requested refactors, sample commands/E2E coverage, runtime downgrade handling, Linux update/repair behavior, and structured install detection are now on the branch. I also addressed the two follow-up review findings in `264c8a0e` and `f22481b5` with regression tests.

Validation on the final product diff: library compile succeeds; 70 focused acquisition tests pass; the runtime/sample extension E2E cases pass; and the Windows UI evidence is in the earlier validation comment with its existing-install limitation called out. All review threads are resolved. Ready for another look.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -160,9 +174,9 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
                 // Requested version has already been installed.
                 const dotnetExePath = install.dotnetInstall.isGlobal ?
                     os.platform() === 'linux' ?
-                        await new LinuxGlobalInstaller(context, this.utilityContext, install.dotnetInstall.version).getExpectedGlobalSDKPath(
+                        await new LinuxGlobalInstaller(context, this.utilityContext, install.dotnetInstall.version).getExpectedGlobalDotnetPath(
```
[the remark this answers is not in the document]
JakeRadMSFT: Fixed in `e3f16d07`. `getSimilarExistingInstall` now passes the tracked install's `installMode` into both `LinuxGlobalInstaller` and `WinMacGlobalInstaller`. A cross-platform regression test seeds a tracked global runtime record and verifies the constructed installer remains in `runtime` mode; it failed with `sdk` before the fix.`n`n*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -77,12 +80,14 @@ export class WinMacGlobalInstaller extends IGlobalInstaller
 We cannot verify our .NET file host at this time. Please try again later or install the SDK manually.`;
```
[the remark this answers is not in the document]
JakeRadMSFT: Fixed in `e3f16d07`. The integrity failure now says "install .NET manually" so it applies to SDK, runtime, and ASP.NET Core products. The installer test asserts the guidance is product-neutral.`n`n*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
```diff
@@ -566,28 +594,35 @@ Permissions: ${JSON.stringify(await this.commandRunner.execute(CommandExecutor.m
...
     {
-        // Note that we could be more intelligent here and consider only if the SDKs conflict within an architecture, but for now we won't do this.
-        const sdks: Array<string> = await this.registry.getGlobalSdkVersionsInstalledOnMachine();
-        for (const sdk of sdks)
+        const installedVersions = this.mode === 'sdk'
+            ? await this.registry.getGlobalSdkVersionsInstalledOnMachine()
+            : (await new DotnetResolver(this.acquisitionContext, this.utilityContext, this.commandRunner).getDotnetInstalls(
+                await this.getExpectedGlobalDotnetPath(requestedVersion, this.acquisitionContext.acquisitionContext.architecture ?? getDefaultArchitecture(), false),
+                this.mode,
+                this.acquisitionContext.acquisitionContext.architecture)).map(install => install.version);
```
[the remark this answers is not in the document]
nagilson: Agreed this is an issue.

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
nagilson: I don't think this is a wording clean up, this only removed the comment.

The comment was left by a human to outline a nuanced assumption and expectation of the fake path content that other code has a dependency upon to specifically prevent a future developer from creating a bug if they didn't know the built-in assumptions.

I would add back this comment without any changes to it.

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -372,8 +379,19 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
             globalEventStream.post(new DotnetAcquisitionTotalSuccessEvent(commandContext.version, install, commandContext.requestingExtensionId ?? '', pathResult.dotnetPath));
         }
 
-        void loggingObserver.flush();
+        await loggingObserver.flush();
```
nagilson: Moving to await would be a performance regression, this should not be awaited.

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -311,17 +312,16 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
         return dotnetPath;
     }
 
-    const dotnetAcquireGlobalSDKRegistration = vscode.commands.registerCommand(`${commandPrefix}.${commandKeys.acquireGlobalSDK}`, async (commandContext: IDotnetAcquireContext): Promise<IDotnetAcquireResult | undefined> =>
+    const acquireGlobal = async (commandContext: IDotnetAcquireContext, mode: DotnetInstallMode, commandKey: string): Promise<IDotnetAcquireResult | undefined> =>
```
nagilson, line 315: Nice change here for the helper function :+1:

vscode-dotnet-runtime-library/src/Acquisition/WinMacGlobalInstaller.ts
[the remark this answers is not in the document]
JakeRadMSFT: Fixed in `9650cb01`. Runtime conflict candidates are now filtered to `install.mode === this.mode` before version comparison. The regression supplies an older `.NET Runtime` plus a newer `ASP.NET Core Runtime` and verifies the ASP.NET patch does not block the runtime request.

The same commit fixes the cross-platform CI failures: Linux success mocks now return the explicit package-manager success code, and the cache-expiry test advances the cache clock instead of sleeping at an external server's TTL boundary. Full library validation is green with 285 passing and 7 platform-skipped tests.

*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -356,24 +370,28 @@ export class DotnetCoreAcquisitionWorker implements IDotnetCoreAcquisitionWorker
...
-        const listSDKsCommand = CommandExecutor.makeCommand('dotnet', ['--list-sdks', '--arch']);
-        const result = await executor.execute(listSDKsCommand, { dotnetInstallToolCacheTtlMs: DOTNET_INFORMATION_CACHE_DURATION_MS }, false);
-
-        if (result.status !== '0')
+        return installs.some(install =>
         {
-            return false;
-        }
+            return install.mode === mode && (install.version === version ||
+                (mode === 'sdk' && allowSdkMajorMinorMatch && install.version.startsWith(`${version}.`)));
```
nagilson: This is a regression for versions that include a minor, I would try to include the check for MajorMinor rather than using startsWith.

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
```diff
@@ -60,11 +60,13 @@ import
     IsEquivalentInstallation
 } from './DotnetInstall';
 import { DotnetInstallMode } from './DotnetInstallMode';
+import { DotnetResolver } from './DotnetResolver';
```
nagilson, line 63: This PR is missing changes to the ModalEventRepublisher, where it has baked in assumptions that an installType of global indicates an SDK event. There also do not currently exist global events for Global Non SDK Events.

vscode-dotnet-runtime-extension/src/extension.ts
JakeRadMSFT: Changed in `1c52976d`. Global acquisition now uses `void loggingObserver.flush()` so log I/O does not delay command completion, matching the local acquisition path.`n`n*Cowritten with Copilot*

vscode-dotnet-runtime-extension/src/extension.ts
JakeRadMSFT: Thanks. The shared helper remains unchanged apart from restoring non-blocking log flushing in `1c52976d` as requested.`n`n*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
JakeRadMSFT: Fixed in `1c52976d`. SDK fallback matching now compares `getMajorMinor` for both the discovered and requested versions instead of relying on `startsWith`. The regression covers both `10.0` and fully specified `10.0.101` requests against installed `10.0.100` on the Linux compatibility path.`n`n*Cowritten with Copilot*

vscode-dotnet-runtime-library/src/Acquisition/DotnetCoreAcquisitionWorker.ts
JakeRadMSFT: Fixed in `69827982`. `ModalEventRepublisher` now distinguishes global SDK, runtime, and ASP.NET Core acquisitions for requested, started, total-success, and final-error events. The commit adds eight global runtime/ASP.NET modal child event types and a dedicated test covering all 12 global mode/lifecycle mappings.

*Cowritten with Copilot*

vscode-dotnet-runtime-extension/src/test/functional/runTest.ts
```diff
@@ -14,7 +14,7 @@ async function main()
     {
         // The folder containing the Extension Manifest package.json
         // Passed to `--extensionDevelopmentPath`
-        const extensionDevelopmentPath = path.resolve(__dirname, '../../../');
+        const extensionDevelopmentPath = [path.resolve(__dirname, '../../../'), path.resolve(__dirname, '../../../../sample')];
```
nagilson: This seems to be an AI hallucination. The sample extension should not be validated inside of the runtime extension directory and these concerns should not be coupled.

vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
```diff
@@ -582,6 +582,24 @@ suite('DotnetCoreAcquisitionExtension End to End', function ()
         await runGlobalSdkInstallTest('10.0.1xx');
     }).timeout(standardTimeoutTime * 1000);
 
+    for (const sampleCommand of ['sample.dotnet.acquireGlobalRuntime', 'sample.dotnet.acquireGlobalASPNETRuntime'])
```
nagilson: This is not an accurate representation of `runGlobalSdkInstallTest` but mirrored for the runtime installs. What was requested earlier was an E2E test for the install tool / runtime extension itself like the global SDK e2e tests. The vendor testing is committed separately on the sample extension and a vendor testing plan.

vscode-dotnet-runtime-extension/src/test/functional/runTest.ts
JakeRadMSFT: Fixed in `b63101d1`. The runtime functional harness is back to a single `extensionDevelopmentPath` and no longer loads or validates the sample extension.`n`n*Cowritten with Copilot*

vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts
JakeRadMSFT: Replaced in `b63101d1` with direct install-tool E2E coverage modeled on `runGlobalSdkInstallTest`. The tests call `dotnet.acquireGlobalRuntime` for `runtime` and `aspnetcore`, assert the fake global host result, and verify these non-SDK modes do not update PATH. Both targeted tests pass against VS Code 1.130.0.`n`n*Cowritten with Copilot*

vscode-dotnet-runtime-extension/src/extension.ts
```diff
@@ -311,17 +312,16 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
         return dotnetPath;
     }
 
-    const dotnetAcquireGlobalSDKRegistration = vscode.commands.registerCommand(`${commandPrefix}.${commandKeys.acquireGlobalSDK}`, async (commandContext: IDotnetAcquireContext): Promise<IDotnetAcquireResult | undefined> =>
+    const acquireGlobal = async (commandContext: IDotnetAcquireContext, mode: DotnetInstallMode, commandKey: string): Promise<IDotnetAcquireResult | undefined> =>
     {
-        commandContext.mode = commandContext.mode ?? 'sdk' as DotnetInstallMode;
+        commandContext.mode = mode;
```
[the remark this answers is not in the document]
JakeRadMSFT: Fixed in `ec4012bf`. The shared global-acquisition helper now normalizes `commandContext.installType` to `global` before constructing the worker context. The existing Windows runtime/ASP.NET E2E cases cover both an omitted value and an accidental `local` value, and verify both are normalized to `global`; both tests pass.

*Cowritten with Copilot*

PR #2582, opened 2026-02-23
Fix Linux Image

nagilson:
We need to migrate to the new Linux image in CI so we can run tests on Linux again. https://github.com/dotnet/vscode-dotnet-runtime/pull/2557 This PR initially had the changes but it should really be an isolated fix - was intending to put it there to improve PR merge velocity.

The new linux image doesn't support some linux versions and is missing some vscode dependencies, which is also resolved in this PR.


nagilson:
Thank you 🥳

PR #2619, opened 2026-03-31
Ensure Dependency Pinning

nagilson:
To be cautious, this PR pins axios related dependencies as well to ensure they don't pull in updated versions of axios. 
Commits:
1. Manual pinning by hand
2. Using npm to update shas 
3. Cleanup of the build
4. Updates to other dependencies while we're at it that aren't around axios (chore) + adding .mjs over .js for linting rules (newer lint version)


vscode-dotnet-runtime-library/src/test/unit/TestUtility.ts
```diff
@@ -151,7 +151,7 @@ export async function getLinuxSupportedDotnetSDKVersion(context: IAcquisitionWor
         }
         if (distroInfo.version < '24.04')
         {
-            return '8.0.100';
+            return '9.0.100';
```
joeloff, line 154: Could this be on 10.0.100 since 10 is LTS?

nagilson: Good question, but the ci image seems to not have access to 10 yet so this is good for now. This isn't my favorite pattern.

Issue #2594, opened 2026-03-02
Question about isOnline() DNS check behavior when using HTTP proxy


tkyggit:
Hi,

First of all, thank you for maintaining this extension.

I have a question about the implementation of `isOnline()` in
WebRequestWorkerSingleton.

As far as I understand, `isOnline()` performs a DNS lookup against
`www.microsoft.com` to determine whether the machine is online.

I was wondering how this is expected to behave in environments where
an HTTP proxy is configured.

In many enterprise networks:

- The client machine does not resolve external DNS directly
- DNS resolution is performed by the proxy server
- HTTP/HTTPS traffic works correctly via the proxy
- But direct DNS lookups from the client may fail

In such a setup, DNS resolution failure does not necessarily mean
that the machine is offline.

So I was curious:

- Is the DNS lookup intentionally used as a simplified connectivity check?
- Is proxy-based DNS resolution considered out of scope?
- Would it make sense to instead attempt a lightweight HTTP request
  using the configured proxy settings?

I may be misunderstanding the design intention, so I would appreciate
any clarification on this.

Thank you!

PR #2635, opened 2026-04-09
Fix isOnline() to fall back to HTTP HEAD when DNS fails behind proxy

nagilson:
In enterprise/proxy environments, the client machine may not resolve external DNS directly, and instead DNS resolution is handled by the proxy server. The existing DNS-only check in isOnline() would incorrectly report the machine as offline, causing unnecessary fallback to cached/offline installs.

When DNS resolution fails and the axios client is available, isOnline() now attempts a lightweight HEAD request to www.microsoft.com through the auto-detected (or manually configured) proxy. This ensures connectivity is correctly detected in proxy environments while keeping DNS as the fast-path for non-proxy setups.

#2594


vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -290,7 +290,61 @@ export class WebRequestWorkerSingleton
...
+        // even though HTTP connectivity works fine. Fall back to a lightweight HEAD request through the proxy-configured client.
+        if (this.client)
+        {
+            const httpFallbackTimeoutMs = Math.max(timeoutSec * 1000, 2000);
+
+            // Auto-detect proxy for this fallback request since we don't have a full acquisition context here.
+            let proxyAgent: HttpsProxyAgent<string> | null = null;
+            try
+            {
+                const autoDetectProxies = await getProxySettings();
```
nagilson: we shouldn't be duplicating this code, need to think about  the context anti pattern

vscode-dotnet-runtime-library/src/test/unit/WebRequestWorker.test.ts
```diff
@@ -137,3 +141,271 @@ suite('WebRequestWorker Unit Tests', function ()
...
+    });
+
+    server.on('connect', (req: http.IncomingMessage, clientSocket: net.Socket, head: Buffer) =>
+    {
+        const [host, portStr] = (req.url ?? '').split(':');
+        const port = parseInt(portStr, 10) || 443;
+
+        const targetSocket = net.connect(port, host, () =>
+        {
+            clientSocket.write('HTTP/1.1 200 Connection Established\r\n\r\n');
```
nagilson, line 331: need to review this still

nagilson (commented):
seems good, but want to take a 2nd look at the tests

MiYanni (approved):

vscode-dotnet-runtime-library/src/Utils/WebRequestWorkerSingleton.ts
```diff
@@ -290,7 +290,47 @@ export class WebRequestWorkerSingleton
...
+        {
+            return true;
+        }
+
+        // DNS failed — but in proxy environments, the proxy handles DNS resolution, so direct DNS lookups may fail
+        // even though HTTP connectivity works fine. Fall back to a lightweight HEAD request through the proxy-configured client.
+        if (this.client)
+        {
+            const httpFallbackTimeoutMs = Math.max(timeoutSec * 1000, 2000);
+            const proxyAgent = await this.getProxyAgent(proxyUrl, eventStream);
```
nagilson, line 303: should validate this doesn't throw and break the offline flow

Issue #2665, opened 2026-04-28
Improve rimraf usage


nagilson:
https://github.com/dotnet/vscode-dotnet-runtime/pull/2648/changes demonstrates mixed rimraf usage - I think we use v3 and v6. We should try to consolidate down to the promisified version of rimraf if that properly exists now. There used to be issues using the newer library due to breaking changes but I dont recall what they were, they seem to be fixed now.

We should also see when its used as a dev dependency and when its used as an actual dependency and make sure that we properly author that in each package.json file in the repo. 

It's also possible rimraf could be replaced if the implementation isn't doing anything deep, its simply used to wipe directories. We should audit the source code to see what it's doing and if we still want to use it as they have consistently broken/changed their API. 

AI agents tend to hallucinate or have issues managing rimraf so something interesting is going on with these versions.

Green00101:
Hi, I’d like to work on this issue. Is that okay?

nagilson:
Go ahead, thanks for offering :)

Green00101:
Hi, I have a question about code style.

When I save files, my IDE automatically changes the formatting. I tried to check the code style guidance in `contributing.md`, but the code style link seems to be unavailable.

<img width="883" height="173" alt="Image" src="https://github.com/user-attachments/assets/1bae2ed6-d1f3-4927-b561-a5da786d8378" />

Is there a recommended formatter setting or command I should use to make sure my changes follow the project’s expected style?

nagilson:
Thanks, good callout there. Sorry for the out of date documentation.

`vscode-dotnet-runtime\eslint.config.mjs` this is the source of all info with respect to formatting, if you run `npm run lint` or any other  linter it should enforce the proper style. 

You may want to install the 'EsLint' Extension from Microsoft, or instead, Disable any aggressive format-on-save (Prettier, "Format Document With…") for .ts files in this workspace, since there's no Prettier/EditorConfig config and it will fight ESLint, in settings.json.

PR #2672, opened 2026-05-04
3.0.2 Branding

nagilson:
Prepare to ship a 3.0.2 release, which will be the 3.0.1 prerelease version but not as a prerelease. Any additional fix PRs for the prerelease issues caught will add onto the changelog.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -180,22 +180,22 @@
 			}
 		},
 		"node_modules/@azure/msal-browser": {
-			"version": "5.8.0",
-			"resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@azure/msal-browser/-/msal-browser-5.8.0.tgz",
-			"integrity": "sha1-+P2CvMKinz9IHugQd7IlGJ2XYoU=",
+			"version": "5.9.0",
+			"resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@azure/msal-browser/-/msal-browser-5.9.0.tgz",
+			"integrity": "sha1-IrCFPnDybmOGKtfvaG/6S07r5KM=",
 			"dev": true,
 			"license": "MIT",
 			"dependencies": {
-				"@azure/msal-common": "16.5.1"
+				"@azure/msal-common": "16.5.2"
 			},
 			"engines": {
 				"node": ">=0.8.0"
 			}
 		},
 		"node_modules/@azure/msal-common": {
-			"version": "16.5.1",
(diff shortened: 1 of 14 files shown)


nagilson:
So far 3.0.1 shows no issues, doubt there will be any needed fixes caught to go to release

vscode-dotnet-sdk-extension/yarn.lock
```diff
@@ -1863,7 +1863,7 @@ util-deprecate@~1.0.1:
   integrity sha1-RQ1Nyfpw3nMnYvvS1KKJgUGaDM8=
 
 "vscode-dotnet-runtime-library@file:../vscode-dotnet-runtime-library":
-  version "1.0.0"
+  version "3.0.2"
```
NikolaMilosavljevic, line 1866: This seems like a big jump - intentional?

package-lock.json
```diff
@@ -180,38 +180,37 @@
...
+			"resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@azure/msal-node/-/msal-node-5.1.5.tgz",
+			"integrity": "sha1-lvjFgBhXHOUK3BC83x/BgZCosA4=",
 			"dev": true,
 			"license": "MIT",
 			"dependencies": {
-				"@azure/msal-common": "16.5.1",
-				"jsonwebtoken": "^9.0.0",
-				"uuid": "^8.3.0"
+				"@azure/msal-common": "16.5.2",
+				"jsonwebtoken": "^9.0.0"
```
NikolaMilosavljevic, line 213: I believe `jsonwebtoken` dependency isn't needed anymore.

sample/package-lock.json
```diff
@@ -94,7 +96,7 @@
 				"mocha": "^11.7.0",
 				"node-cache": "^5.1.2",
 				"open": "^8.4.0",
-				"rimraf": "^6.1.3",
```
NikolaMilosavljevic, line 97: Is this a regression?

vscode-dotnet-sdk-extension/yarn.lock
nagilson: Yes - thank you for the thoughtful question. The library should have been updated, it didn't really matter in the past, but we're planning to ship the library types at some point so it's better to keep it in sync now.

sample/package-lock.json
nagilson: Great question - It is not, this happens - rimraf is used as a dev dep and an internal code dep, but there are breaking changes we need to account for - https://github.com/dotnet/vscode-dotnet-runtime/issues/2665 tracks this work

package-lock.json
nagilson: I double checked but it looks like it is still needed, trying to rebuild and this is still included as a dep in the lock.

NikolaMilosavljevic (approved):

Issue #2642, opened 2026-04-13
[NETE2ESDK] Installed .NET runtimes should be arranged in order on the Uninstall list.


vwenchen:
### Describe the bug.

The Installed .NET runtimes are listed disorderly.

### Please attach log file(s) here if you have any.

[log.zip](https://github.com/user-attachments/files/26668561/log.zip)

### Reproduction Steps

1. On windows, run the sample command 'Sample: Concurrently acquire .NET runtimes(8.0, 9.0, 10.0)' 
2. Open the Command '.NET Install Tool: Uninstall .NET' 
3. Observe the listed runtimes

### Expected Behavior

The Installed .NET runtimes should be arranged in order

### Exceptions & Errors (if any)

The Installed .NET runtimes are listed disorderly.

<img width="452" height="137" alt="Image" src="https://github.com/user-attachments/assets/18268a24-24a7-4a50-94dc-c70f94c74164" />

### .NET Install Tool Version

3.0.1

### Your operating system? Please provide more detail, such as the version and or distro above.

N/A

### VS Code Version

1.115.0

vwenchen:
The installed .NET SDK also arranged disorderly.

NicoleWang001:
Here is the output if installed .NET 11.0 Preview as well.

<img width="431" height="232" alt="Image" src="https://github.com/user-attachments/assets/d678cc9d-cad5-422e-8eb0-4810257167f6" />

nagilson:
This can definitely be improved, though not related to the new changes I dont believe. not a blocker.

PR #2740, opened 2026-06-28
improve sorting of installed versions when uninstalling dotnet version

nitheeshgovind:
When the command "Uninstall .NET" is selected, the displayed versions are not sorted based on the version numbers.
#2642 

## Root Cause

`localeCompare` was used which did not sort the numbers correctly.

## Changes

Added a new `compareVersion` function which splits the version into integers and compares each index.

```ts
const compareVersions = (x: InstallRecord, y: InstallRecord): number =>
        {
            const xParts = x.dotnetInstall.version.split('.').map(part => Number(part.split('-')[0]));
            const yParts = y.dotnetInstall.version.split('.').map(part => Number(part.split('-')[0]));
            const length = Math.max(xParts.length, yParts.length);
            for (let index = 0; index < length; index++)
            {
                const difference = (xParts[index] ?? 0) - (yParts[index] ?? 0);
                if (difference !== 0)
                {
                    return difference;
                }
            }
            return 0;
        };
```

I also thought about creating a new function in `VersionUtilities` under the runtime-library, but thought to check first.
Let me know if it is ok.

### DIFF
--- a/vscode-dotnet-runtime-extension/src/extension.ts
+++ b/vscode-dotnet-runtime-extension/src/extension.ts
@@ -575,27 +575,35 @@ export function activate(vsCodeContext: vscode.ExtensionContext, extensionContex
 
     const dotnetUninstallPublicRegistration = vscode.commands.registerCommand(`${commandPrefix}.${commandKeys.uninstallPublic}`, async () =>
     {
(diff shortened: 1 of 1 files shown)


nitheeshgovind:
@dotnet-policy-service agree

PR #2760, opened 2026-07-21
Support `preview` .NET versions in vscode acquisition 

### LINKED ISSUE #2759: `Preview` versions are unsupported, which causes issues for pinned `global.json`
`preview` versions have not been supported for acquisition scenarios; this was intentional and part of why the 'getting started' or 'install a .NET SDK' page in C# DevKit does not have a button for .NET 11 yet since it has not ga'd. 

The main reason for this is the version parsing logic does not handle `-preview` or other types of versions. There are also functions in the code which rely on `install id`s, which, in order to support legacy code, have varying formats where the version must be extracted from a string that may or may not include other characters such as `-` (found in preview versions.) We also didn't view it as necessary as DevKit did not intend to build off these newer .NET versions.

`global.json` pins may make the workspace requirements try to install .NET 11 which is unsupported, this for example is present in the `runtime` repo today (reported by @333fred.) This increases the urgency of that fix.

```
Cancelled Installation of .NET 11.0.100-preview.6.26352.110-global~x64.
The requested version was not in the correct format. Allowable formats are
        * <MAJOR> (for example '6')
        * <MAJOR>.<minor> (for example '6.0')
        * <MAJOR>.<minor>.<feature band without patch version> (for example '6.0.4xx')
        * <MAJOR>.<minor>.<patch> (for example '6.0.402')
        * Your version was resolved to:  11.0.100-preview.6.26352.110

Error : (DotnetAcquisitionFinalError)
```

nagilson:
Resolves https://github.com/dotnet/vscode-dotnet-runtime/issues/2759

I also built the VSIX to confirm manually that it works now, and will have vendors also try our testing plan with preview versions.

### DIFF
--- a/vscode-dotnet-runtime-library/src/Acquisition/VersionUtilities.ts
+++ b/vscode-dotnet-runtime-library/src/Acquisition/VersionUtilities.ts
@@ -19,6 +19,8 @@ import { IAcquisitionWorkerContext } from './IAcquisitionWorkerContext';
 import { BAD_VERSION } from './StringConstants';
 
 const invalidFeatureBandErrorString = `A feature band couldn't be determined for the requested version: `;
+// Historical evidence for each accepted suffix family is documented in Documentation/version-suffixes.md.
+const historicalDotnetPreReleaseSuffixPattern = /^(?:preview|rc)(?:(?:\.\d+)+|\d+(?:\.\d+)*(?:-(?:\d+(?:-\d+)*|final))?|-\d+(?:-\d+)*)?$/i;
 
 /**
  *
@@ -144,6 +146,10 @@ export function getSDKPatchVersionString(fullySpecifiedVersion: string, eventStr
 }
 
 
+/**
+ *
+ * FullySpecifiedVersion is expected to be a fully specified version, but may be a preview version.
+ */
 export function getSDKFeatureBandOrPatchFromFullySpecifiedVersion(fullySpecifiedVersion: string): string
 {
     const patch: string | undefined = fullySpecifiedVersion.split('.')?.[2]?.substring(1)?.split('-')?.[0];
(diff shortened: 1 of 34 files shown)


vscode-dotnet-runtime-library/src/Acquisition/VersionUtilities.ts
```diff
@@ -18,6 +19,8 @@ import { IAcquisitionWorkerContext } from './IAcquisitionWorkerContext';
 import { BAD_VERSION } from './StringConstants';
 
 const invalidFeatureBandErrorString = `A feature band couldn't be determined for the requested version: `;
+// Historical evidence for each accepted suffix family is documented in Documentation/version-suffixes.md.
+const historicalDotnetPreReleaseSuffixPattern = /^(?:preview|rc)(?:(?:\.\d+)+|\d+(?:\.\d+)*(?:-(?:\d+(?:-\d+)*|final))?|-\d+(?:-\d+)*)?$/i;
```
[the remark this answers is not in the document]
nagilson: I'm ok with this.

vscode-dotnet-runtime-library/src/Acquisition/VersionUtilities.ts
```diff
@@ -220,12 +227,12 @@ export function isValidLongFormVersionFormat(fullySpecifiedVersion: string, even
 
 /**
  *
- * @param fullySpecifiedVersion the requested version to analyze.
- * @returns true IFF version is of an rc, preview, internal build, etc.
+ * @param version the version string to analyze.
+ * @returns true IFF the version has a non-empty pre-release suffix. This function does not validate the version.
  */
-export function isPreviewVersion(fullySpecifiedVersion: string, eventStream: IEventStream, context: IAcquisitionWorkerContext): boolean
+export function hasPreReleaseSuffix(version: string): boolean
```
mthalman, line 233: This is a breaking change since `VersionUtilities` is exported for this library: https://github.com/dotnet/vscode-dotnet-runtime/blob/7fb823ba0e1e012be68804ee3efb3bbd22d759de/vscode-dotnet-runtime-library/src/index.ts#L37

nagilson: That's a great point, though in this case we aren't concerned with breaking changes to the interfaces. This library has never been exported publicly and there are no prod dependencies on it though outside of the VS code commands and type interfaces into those APIs themselves - for those, we have to be extremely cautious as they would essentially break all C# / C#DK functionality immediately when shipped. We break the library APIs outside of that quite frequently, however.

Thank you for taking a look!

mthalman (approved):

